PR #39, opened 2017-05-08
Create .travis.yml

stephengroat:
Adds travis builds for `left` and `right`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: generic
+os: osx
+cache:
+  directories:
+    - $HOME/Library/Caches/Homebrew/Cask
+install:
+  - brew cask install eclipse-cpp gcc-arm-embedded
+  - /Applications/Eclipse.app/Contents/MacOS/eclipse -noSplash 
+    -application org.eclipse.equinox.p2.director 
+    -repository http://gnuarmeclipse.sourceforge.net/updates 
+    -installIUs ilg.gnuarmeclipse.managedbuild.cross.feature.group 
+    -profileProperties org.eclipse.update.install.features=true
+script:
+  - /Applications/Eclipse.app/Contents/MacOS/eclipse -noSplash 
+    -application org.eclipse.cdt.managedbuilder.core.headlessbuild 
+    -import $TRAVIS_BUILD_DIR/left/build/kds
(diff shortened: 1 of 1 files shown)


mondalaci:
On one hand, I'm very happy that you managed to automate the build, but I'm afraid that your solution won't work on the long term due to a major sweeping change. Kinetis Development Studio has been deprecated, and I'll be porting the firmware to [MCUXpresso IDE](https://mcuoneclipse.com/2017/03/28/mcuxpresso-ide-unified-eclipse-ide-for-nxps-arm-cortex-m-microcontrollers/), its replacement.

The major question to ask is whether your solution relies on the GCC ARM Eclipse plugins. If so, your approach won't work. The reason is that MCUXpresso IDE doesn't use the GCC ARM Eclipse plugins by default and [installing them is a pain in the ass](https://mcuoneclipse.com/2017/04/02/mcuxpresso-ide-importing-kinetis-design-studio-projects/), so we'll realy the native build system of MCUXpresso IDE.

stephengroat:
@mondalaci I just looked and MCUXpresso doesn't use the GCC ARM Eclipse plugin, but it still relies on Eclipse CDT. I can't be 100% sure, but I think the projects should still import and build. Lines 8-12 would be removed from the travis script, and something else _might_ have to be added, but the basic concept should still work

mondalaci:
@stephengroat I'll give your script a try tomorrow. I'm delighted that the build can be executed from the command line this way. Seems like this capability is implemented in the core of Eclipse and not dependent on the GCC ARM Eclipse plugins. If it properly returns with success or error then it should work for us.

I'd be a little more comfortable with a Linux travis script because the majority of us (developers and backers included) use Linux, so we could maintain it easier on the long run, but it's a minor issue, and porting the script should be trivial.

I'm also excited to ditch the ARM GCC Makefile based command line build scripts because they're a mainteance burden.

stephengroat:
@mondalaci the reason i chose `osx` was because of installing `eclipse` and the `gnu-arm-embedded` toolchain. the versions in `brew cask` are very up to date, the versions in `apt` are old (in the case of `eclipse`) and non-existent (in the case of `gnu-arm-embedded` toolchain).

mondalaci:
@stephengroat Works like a charm! Thank you so much!

PR #43, opened 2017-06-19
enabled parallel build, fixed refresh policy and ignoring generated Eclipse settings

ErichStyger:
minor modifications. Fixed refresh policy setting and enabled parallel build for faster building time, plus ignoring generated eclipse settings files.


mondalaci:
The build got so much faster! Thank you very much, @ErichStyger!

PR #45, opened 2017-07-07
Improve the configuration parser

### LINKED ISSUE #42: Parse keymaps
Currently, the configuration parser is only able to parse layers. As the next step, it should parse keymaps (which are composed of 4 layers). Please see `right/src/config/parse_keymap.c` for the configuration parser and also the `shared/src/config-serializer` directory of Agent for the serialization format.

When using Agent, ctrl+u sends the current layer to the UHK. Similarly, ctrl+i sends the current keymap and ctrl+o sends the whole configuration. Press ctrl+shift+i in Agent to invoke developer tools, then click on the Console tab to see the USB packages that get sent to the UHK.

Upon finishing the implementation, the `ParseLayer` function call of `usb_protocol_handler.c` should be replaced with the newly created `ParseKeymap` function call.

eltang:
This allows the configuration parser to parse keymap configurations and resolves #42. In addition, it introduces a bit of error checking to so that unexpected data is less likely to cause problems.

### DIFF
--- a/right/src/config/parse_keymap.c
+++ b/right/src/config/parse_keymap.c
@@ -169,7 +169,7 @@ static void clearModule(uint8_t layer, uint8_t moduleId) {
     memset(&CurrentKeymap[layer][moduleId], 0, MAX_KEY_COUNT_PER_MODULE * sizeof(key_action_t));
 }
 
-void parseLayer(serialized_buffer_t *buffer, uint8_t layer) {
+static void parseLayer(serialized_buffer_t *buffer, uint8_t layer) {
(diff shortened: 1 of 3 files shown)


right/src/config/parse_keymap.c
```diff
@@ -4,6 +4,16 @@
 
 #define longCompactLengthPrefix 0xff
 
+enum {
```
mondalaci: Please typedef this enum as `parser_error_t`, move it to `parse_keymap.h`, and use `parser_error_t` instead of `uint8_t` wherever it makes sense.

PR #57, opened 2017-08-08
Implement keymap switching

### LINKED ISSUE #52: Switch between keymaps
In order to make the firmware handle keymap switch actions, a number of changes must be made.

First and foremost, a staging area should be created for newly transferred configurations. This would involve a `config_buffer_t StagingBuffer` variable. Agent will transfer the new configuration to this variable, then the parser will validate the configuration of the staging area. Agent will then apply the configuration, at which point the firmware copies `StagingBuffer` to `ConfigBuffer`.

Admittedly, `StagingBuffer` takes a lot of RAM, but we can afford it, and this is the only way for the firmware to properly validate the new configuration without overwriting it. `ConfigBuffer` will be actively used during the operation of the UHK to switch keymaps, play back macros, and who knows for what else, so we cannot just overwrite it with a possibly invalid configuration.

Make sure to validate the keymap ids of switch keymap actions.

The active keymap must be remembered when a new configuration gets applied. The 3 letter abbreviation must be remembered because the index of the keymap can change in the new configuration. If the active keymap no longer exists in the new configuration then the default keymap must be activated.

eltang:
This resolves #52.


right/src/config_parser/parse_keymap.c
```diff
@@ -175,7 +179,8 @@ static parser_error_t parseLayer(config_buffer_t *buffer, uint8_t layer) {
     return ParserError_Success;
 }
 
-parser_error_t ParseKeymap(config_buffer_t *buffer) {;
+parser_error_t ParseKeymap(config_buffer_t *buffer, uint8_t keymapIdx, uint8_t keymapCount) {;
```
mondalaci: Extra semicolon here and curly should be in the next line.

right/src/usb_protocol_handler.c
```diff
@@ -86,6 +87,8 @@ void readMergeSensor(void)
 void applyConfig(void)
 {
     uint8_t *temp;
+    char oldKeymapAbbreviation[3];
+    char oldKeymapAbbreviationLen;
```
mondalaci: Shouldn't this be a `unit8_t`?

eltang: Yes, let me fix that.

PR #58, opened 2017-09-27
Improve OSX dependency management through brew bundle

stephengroat:
a new and improved osx dependency management system, takes the dependencies out of `.travis.yml` and moves them into a more track-able file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache:
   directories:
     - $HOME/Library/Caches/Homebrew/Cask
 install:
-  - brew cask install eclipse-cpp gcc-arm-embedded
+  - brew tap Homebrew/bundle && brew bundle
   - /Applications/Eclipse.app/Contents/MacOS/eclipse -noSplash 
     -application org.eclipse.equinox.p2.director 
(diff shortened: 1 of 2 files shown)


mondalaci:
Thanks for your help! Unfortunately, just like the master branch, this PR fails in Travis. Can you possibly resolve this issue?

stephengroat:
@mondalaci i looked (even tried a bump) and it seems like there's a timeout during a download.

could you try restarting the last travis job that succeeded on master to double check?

mondalaci:
I've just restarted the last Travis job of the master branch. In any case, by looking at the commit history of the repo it's easy to notice that sometimes the build succeeds and most of the times it doesn't regardless of the state of the repo. It'd be great if the build worked consistently.

mondalaci:
The build that was successful now errored when rebuilt, proving the inconsistency of the current build process.

stephengroat:
`travis` is having problems with their `osx` infrastructure.

i'll look into simplifying a script for `linux`

mondalaci:
I'm merging this in the hope that it'll make the build issues less prevalent.

PR #59, opened 2017-10-09
brew updated, travis is behind

stephengroat:


mondalaci:
This also seems to fail, unfortunately:

```
$ /Applications/Eclipse.app/Contents/MacOS/eclipse -noSplash -application org.eclipse.equinox.p2.director -repository http://gnuarmeclipse.sourceforge.net/updates -installIUs ilg.gnuarmeclipse.managedbuild.cross.feature.group -profileProperties org.eclipse.update.install.features=true
/Users/travis/.travis/job_stages: line 57: /Applications/Eclipse.app/Contents/MacOS/eclipse: No such file or directory
The command "/Applications/Eclipse.app/Contents/MacOS/eclipse -noSplash -application org.eclipse.equinox.p2.director -repository http://gnuarmeclipse.sourceforge.net/updates -installIUs ilg.gnuarmeclipse.managedbuild.cross.feature.group -profileProperties org.eclipse.update.install.features=true" failed and exited with 127 during .
Your build has been stopped.
/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found
```

Would you please explain why this is an improvement over the exisitng build script?

stephengroat:
sure, `bundle` is definitely `brew`'s solution for dependency management. currently, `brew install [PKG]` fails if the package is already installed. `bundle` and `Brewfile`s are a way around that particular _feature_, allow packages to be listed and upgraded individually. it also controls the dependency outside of travis, allow a user to grab the repo and get all dependencies and tools with a single file.

[linuxbrew](http://linuxbrew.sh/) is trying to do the same thing for linux

mondalaci:
Thanks for explaining! Why would be a package already installed that is to be installed by brew? Is it sometimes cached by Travis?

stephengroat:
@mondalaci kinda, `travis` (in their infinite wisdom) installs **some** packages (poorly documented) and updates their `osx` images kinda randomly (maybe they tell their on-prem customers, but i'm an enterprise customer and don't hear about it)

since the `osx` images are randomly updated, those pre-installed packages get randomly out of date. when dependencies exist, i've found that listing them in `Brewfile` allows them to be installed and updated with crazy scripting

mondalaci:
Ah, now I see! How about migrating to a Linux build eventually as a long term solution? I assume that you're more familiar with OSX, right?

stephengroat:
@mondalaci i'm probably more familiar with linux, just maintaining the `eclipse` and `arm-gcc` install scripts can be a real pain (java moves around, etc). i'm hoping to see if i can get the `nix` package manager or `linuxbrew` in eventually

mondalaci:
Ok, Thanks! Well, seems like you're making progress. Please keep it up! It'd be great to see CI work.

stephengroat:
@mondalaci running again, there's a linker error, but that's not me :-)

mondalaci:
Thank you, great work!

I cannot reproduce this linker error on my machine. Let me ask our Kinetis guru. :)

mondalaci:
As it turns out, even Erich, our Kinetis expert hasn't ever encountered with this linker issue:

> The error is not apparent. I have never seen this VMA (Virtual Memory Area?) message.
> To me it seems that the linker tries to do something with this, and this very much could be because of that different GCC/linker version you are using

I'm merging your PR because it's clearly a step forward. Going forward, it'd be great to set up a build environment with matching GCC and linker versions.

The current toolchain versions that ship with KDS 3.2.0 (Kinetis Design Studio, an Eclipse fork for Kinetis microcontrollers):
* gcc version 4.8.4 20140725 (release) [ARM/embedded-4_8-branch revision 213147] (GNU Tools for ARM Embedded Processors) 
* GNU ld (GNU Tools for ARM Embedded Processors) 2.23.2.20140731

PR #60, opened 2017-10-10
Fix gcc-arm-embedded version in cask

stephengroat:
#59 pin version of `gcc-arm-embedded` by manually including the `cask` formula

`gcc-arm-embedded.rb` can be moved around to get it out of the root directory, just make sure to update the `Brewfile` with the path


stephengroat:
success

not exactly speedy, but it works. super excited to get my keyboard soon!

mondalaci:
You're so awesome! So happy you've made it work!

We're working on your keyboard, and eager to deliver. :)

PR #61, opened 2017-10-10
Build linux in CI

stephengroat:
Haven't gotten the perfect `gcc-arm-embedded` package from nix and the travis file is a little bit messy, but it includes a ~3 minute linux build on travis using `nix` as a package manager

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: generic
 sudo: required
+os: linux
 cache:
   directories:
     - $HOME/Library/Caches/Homebrew/Cask
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ matrix:
         - nix-env -i gcc-arm-embedded-4.8-2014q1-20140314
         - nix-env -i eclipse-cpp
     - os: osx
-      env: PATH=$PATH:/Applications/Eclipse\ CPP.app/Contents/MacOS
+      env: PATH=$PATH:"/Applications/Eclipse CPP.app/Contents/MacOS"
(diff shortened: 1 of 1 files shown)


mondalaci:
Am I right that now the Linux build succeeded and OSX failed? Do we need two builds? If not, which one would you prefer and why?

stephengroat:
slight issue with the `linux` build: the version of `gcc-arm-embedded` (the toolchain) doesn't exactly match what you guys are using. let me see if i can get both working, the only problem i'm currently having is some whitespace issues that should be solve-able

stephengroat:
fixed, you now get both `osx` and `linux`

i need to work on how to use `nix` to get the proper version of `gcc-arm-embedded` toolchain installed, but this is ready to merge

mondalaci:
Thanks, you're seriously awesome!

What's the advantage of building for two OSes? I'd think that it can break easier (in case either build breaks) which is not an advantage, although if either breaks I can just comment it out temporarily and the other will keep working.

stephengroat:
Sorry, i thought the comment sent (i'm currently suffering from some sketchy internet).

the `linux` build currently has the wrong version of the `gcc-arm-embedded` toolchain. I haven't figure out `nix` well enough to get the right version yet.

in short: `linux` is fast but has the wrong `gcc-arm-embedded` toolchain version, `osx` is correct but super slow

mondalaci:
Makes perfect sense, thank you for explaining! Let me merge this one, and then you can send another one if you'll have time to figure out Linux even better.

Thanks so much!

PR #92, opened 2018-02-11
Logical maximum is wrongly assumed to be an unsigned integer

Lauszus:
Thus 0xFF is interpreted as -1 and not 255.

Simply setting the number of bits to 16 solves this.


mondalaci:
Great catch!

Issue #85, opened 2018-01-01
Windows: UHK prevents sleep


laxu:
With the UHK connected, my computer running Windows 10 64-bit won't go to sleep at all. Display won't sleep either so I assume that UHK keeps doing something over the USB as disconnecting it fixes the issue.

mondalaci:
I can reproduce this issue. I think chances are good it's related to #84.

laxu:
DIsabling the mouse in Device Manager doesn't seem to fix the issue for me.

Hurricaaane:
Same here, I can also reproduce this issue, but unlike #84, disabling the mouse the does not fix it.

Hurricaaane:
I noticed today that my computer went to sleep while my UHK was plugged in. Wondering why this was the case, I then noticed that the left half of the keyboard was not responding ( see #74 ). Unplugging the bridge cable and replugging it woke up my computer...

mondalaci:
@Hurricaaane Thanks for the info! #74 apparently affects the USB interrupt, too. Fixing #74 will make this always an active issue, so this will have to be separately fixed regardless.

laxu:
Tested this but alas no luck on my machine. I set Windows display sleep to 1 min and computer sleep to 2 min but my UHK has to be disconnected for either to happen. I tried disconnecting left side, disconnecting cable entirely and connecting the halves together without cable. So I think for this issue you can at least partially rule out the left side causing this.

I also can't wake the computer by pressing a key on the keyboard.

mondalaci:
Disconnecting the left half doesn't eliminate this issue. It gets only temporarily eliminated if the communication between the halves halt due to #74. I think I don't have any more information for this to resolve it.

Jopie01:
To me this seems like there is some unsolicited communication from the UHK to windows. Once I had an Arduino Leonardo, which also could act as mouse. To cheat a bit, I created a program to move the mouse a little bit every now and then, to show my colleagues that I'm still behind my desk and hard at work 😆. My system didn't lock itself or did go to sleep.

I'm also curious if this happens on other OS'es too. IMO it should behave the same. I will check on my system with Fedora 27

Jopie01:
I tested Fedora 27 and Windows 8 on my dual boot system.

- Fedora 27: No problems just works. Even sleep and blanking the screen. Dmesg after plugging in the keyboard shows me that there are 5?? input devices
```
[...]

Issue #84, opened 2017-12-31
Whenever the UHK is plugged in a host machine, Chrome Remote Desktop is blocked from remotely interacting with the machine


Hurricaaane:
Whenever the keyboard is plugged in a host machine, Chrome Remote Desktop is blocked from remotely interacting with the machine.

Let's say I have two machines:
- A desktop with UHK plugged in
- A Chromebook

If I use my Chromebook to connect to my desktop machine remotely using Chrome Remote Desktop,
- The Chromebook will not be able to take control of the mouse and keyboard if the UHK is plugged in
- When the UHK is unplugged while the Remote Desktop session is running, it will allow the Chromebook to take control of the mouse
- When the UHK is plugged again, it will prevent the Chromebook from taking control of the mouse again

Usually, Chrome Remote Desktop prevents the Chromebook from sending inputs to the desktop if the host's keyboard has recently been typed on or when the host's mouse has recently been moved or clicked. Presumably the UHK sends some kind of events that triggers some condition?

Anyways it means that the UHK is not behaving like a traditional mouse or keyboard. When the UHK is at rest, I expect it to behave like a traditional keyboard and mouse at rest. This is a huge problem for me as I don't want to unplug the UHK whenever I leave home.

mondalaci:
This is rather surprising, because the UHK doesn't do anything super-fancy on the USB level that gaming keyboards / mice don't do.

My assumption is that your desktop OS is not in peace with the descriptor of some of the USB interfaces of the UHK, so the descriptor must be further tweaked on the firmware level.

May I know the OS of your desktop machine?

Hurricaaane:
The desktop machine is Windows 10.

To add more observations related to the issue, I've noted when I plug my keyboard, a peripheral appears in my list of peripherals under the category "Mouse and other pointing devices" (translated from French), HID Mouse. If I keep my UHK plugged in and disable this peripheral (using right click > Disable this device), it no longer blocks the remote desktop inputs (but obviously I lose the ability to use the UHK mouse functionnality).

If you need more details don't hesitate to ask me.

G.Y.

mondalaci:
Thanks for the useful info! Will let you know if I need more help.

mondalaci:
I'd like to add that according to my testing this only happens if the device being shared runs Windows. I guess the other end of the connection doesn't matter, but I used Linux.

This should be resolved in the near future now that we have a USB expert on our team.

Hurricaaane:
[...]

PR #96, opened 2018-03-11
Various fixes and improvements

Lauszus:
See commit messages


mondalaci:
This makes the speed of the mouse pointer vary, and doesn't fix neither #84 nor #85.

PR #123, opened 2018-06-08
Fix issue with keyboard freezing

Lauszus:
* Fixes #122 
* Fix PIT timer period and added functions for getting the current time in microseconds - see #127

PR #127, opened 2018-06-22
Fix PIT timer period and added functions for getting the current time…

Lauszus:
… in microseconds

Before:
before

After:
after

### DIFF
--- a/right/src/timer.c
+++ b/right/src/timer.c
@@ -37,7 +37,8 @@ uint32_t Timer_GetCurrentTimeMicros() {
     ms = CurrentTime; // Read the overflow counter
     EnableGlobalIRQ(primask); // Enable interrupts again if they where enabled before - this should make it interrupt safe
 
-    uint32_t us = COUNT_TO_USEC(count, timerClockFrequency);
+    // Calculate the counter value in microseconds - note that the PIT timer is counting downward, so we need to subtract the count from the period value
+    uint32_t us = 1000U * TIMER_INTERVAL_MSEC - COUNT_TO_USEC(count, timerClockFrequency);
     return ms * 1000U * TIMER_INTERVAL_MSEC + us;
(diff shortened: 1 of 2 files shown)


Lauszus:
Note that this fix will make the mouse move twice as fast.

Lauszus:
Closing, as this PR is part of #123 as well.

mondalaci:
@Lauszus Thank you for reopening this PR.

@eltang I know you'd like to get this PR merged for the macro engine but I cannot merge it yet because it makes the movement of the mouse pointer choppy. Sorry, but it will only be merged with #123 which addresses the mouse issue and contains this code.

Lauszus:
@mondalaci sorry but this does not effect the mouse choppiness. It will make the report actually be sent out at 4 ms and not 8 ms, which is the case with the current code in master. So the real fix should be to update the code that uses the timer, as this make the milliseconds timer actually count correctly.

Btw there is a bug in this line of the USB report updater code: https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/usb_report_updater.c#L399-L404, as ```hasActiveMouseState``` should only be set true if the wheel or the mouse is moved.

Lauszus:
I.e. this will effect every piece of code that is relying on the timer. For instance the ```KEY_DEBOUNCER_TIMEOUT_MSEC``` value is currently set to 100, but in reality it is 200 ms, because of this bug in the timer implementation.

mondalaci:
@Lauszus Thanks for the explanation! I understand that this PR shouldn't affect mouse chopiness but according to my testing it clearly does. I'd much rather wait for #123 than delving into this. I understand that the timer affects everything.

@Lauszus @eltang You're welcome to send a PR regarding the `hasActiveMouseState` related issue. I have a lot on my plate right now.

PR #144, opened 2018-07-22
Make some improvements to the sleep/wake code

eltang:
This was originally an attempt to fix some issues with waking up a certain computer. It turned out that the problem was on the computer side, but this is still an improvement which increases compliance with the USB spec.


Lauszus:
Looks good to me :)

Issue #125, opened 2018-06-18
Secondary-role layer switch causes weird alt-tab behaviour


mhantsch:
I assigned the left space to send Space scancode and secondary role to switch to the Mouse layer.
On the Mouse layer, I configured Tab to send LAlt-Tab.
When I press Space-Tab, the windows switch immediately (to the previously active window). If I keep holding down the Space key, I cannot keep tabbing through all the open windows.

When instead I use a primary role switch to the Mouse layer, everything works as intended, and I can tab through the open windows while keeping the layer switch key depressed. However, doing the same with a secondary role switch, UHK only seems to send a single LAlt-Tab and immediately releases LAlt although I keep that secondary role key depressed. I think it should keep the LAlt modifier depressed until I let go of the secondary role key.

mhantsch:
This might be related to #69

mondalaci:
Thanks for reporting! I can clearly reproduce this issue.

TorC8:
That opens up the question of right behavior.  Certainly I think it should hold for repeated presses of the key, but what about the case of, while still holding $MOUSE, pressing a different key that doesn't have $ALT configured as part of its keycode?

My inclination is to think that the right thing is to drop the modifier at that point before sending the new keycode, but I can see arguments for both options.

mondalaci:
I agree that dropping Alt is the right thing to do in the above case. I can't see why any other option would make sense.

mhantsch:
Unless it's Shift; then you need to keep Alt held down because the user is trying to type Alt-Shift-Tab to cycle the windows backwards. So, I would argue: drop Alt only if the next keypress produces an actual character, macro, layer switch, or keymap switch; don't drop it on modifiers-only (Ctrl/Shift/Alt).

mondalaci:
@mhantsch Yep, we're on the same page.

mhantsch:
I can confirm that this is resolved with firmware 8.4.2.

PR #168, opened 2018-09-02
Send primary role modifiers consistently

### LINKED ISSUE #165: Issue with key combos as primary with a secondary roles assigned
Firmware after 8.4.1 (i.e. 8.4.1 doesn't exhibit this behaviour) seems to be having an issue if the primary role of a key is a combination and the secondary role a modifier.

For example: if I have the right alt key (that is to say the 3 key in on the bottom row on the right side) set to send super + space (that is to trigger Spotlight search on a Mac) and the secondary role to send right alt --- the super + space (the primary role) doesn't seem to be sent consistently. 

Occasionally it sends the combination but other times only seems to be sending space. Almost as if the combination is being sent too fast to register to the operation system?

If I revert to 8.4.1, I don't have the issue anymore.

eltang:
This closes #165.

### DIFF
--- a/right/src/usb_report_updater.c
+++ b/right/src/usb_report_updater.c
@@ -411,6 +411,7 @@ static void updateActiveUsbReports(void)
                 if (keyState->previous && secondaryRoleSlotId == slotId && secondaryRoleKeyId == keyId) {
                     // Trigger primary role.
                     if (secondaryRoleState == SecondaryRoleState_Pressed) {
+                        keyState->previous = false;
                         applyKeyAction(keyState, action);
                     }
                     secondaryRoleState = SecondaryRoleState_Released;


mondalaci:
This doesn't work well. Let's say that on the base layer I pick a key and set Alt+Tab as the primary role, and Alt as the secondary role. Then I tap the key. Alt+Tab triggers, but Alt stays held which shouldn't happen.

Please test your fixes in the future. Not testing them causes extra communication overhead and delays.

eltang:
I did test this fix. The behavior you described is the one that was chosen for fixing #125. This fix doesn't change it.

mondalaci:
I understand your reasoning, but the current behavior is clearly wrong from a usability standpoint. To make a simpler example, let's say I bind Alt+Tab to a key on the base layer. (No secondary roles are involved.) Then I tap the key, and Alt gets stuck.

I think the modifiers of a shortcut key should stick only as long as as the current layer switcher key is being held. On the base layer, no layer switcher key is held, so the modifiers of shortcuts should not stick.

eltang:
The purpose of this PR is to ensure that a key's primary role functions the same way regardless of whether that key has a secondary role. I get that you have concerns about the behavior of sticky modifiers, but this PR doesn't touch it at all. That can be dealt with separately after this PR is merged.

Issue #187, opened 2018-10-08
Alphabetic keys do not play well with secondary roles


p4elkin:
I have tried to replicate the 'vim mode' workflow that was really convenient with Karabiner back in days: while holding 's' the hjkl turn into arrows. In UHK terms - I assigned 'switch to MOD layer' as secondary role to 's' key. 

However, now when typing words fast, often the secondary role kicks in for 's'. From the event monitor I see that secondary role activates when some other key is pressed and it looks like **if I type 'st' (even if really quickly) - it is treated as if I pressed 't' while holding 's' (and => MOD).** 

I tried to find an existing issue for this - looks like there're some quirks in a similar spirit, but not exactly the same.

This issue of course can be observed not only for the alphabetic keys, but I think they are the most prominent example. 

Do you think this might be mitigated by maybe fiddling with debounce algorithm or some other settings?

firmware version: 8.5.2 
agent version: 1.2.11
os: macos sierra

luteijn:
Using Space with shift as secondary function and will occasionally get something run togeter likeThis when typing quickly. Perhaps it would be better if there was some sort of setting to delay going to 2nd role for x ms. Of course, ideal settings would differ for different people.

p4elkin:
Another aspect of this is the following: in my IDE I use alt + Enter shortcut extensively. On my UHK alt is facilitated via secondary-role, at the same time Enter key has a secondary role assigned (mouse layer). **So, when I hold ALT, tapping ENTER triggers its secondary role. (Should this be extracted into a separate issue?).**

@luteijn I agree that the simplest solution to this (in ideal world) would be a threshold delay before the secondary role kicks in.

Maybe the maintainers could give some initial pointer at the code base part to look at for investigation and experiments?

mondalaci:
I rather wouldn't introduce a (configurable) delay. Accidental activation is the nature of this feature. A workaround I'd consider is to use a least frequently used key.

https://github.com/UltimateHackingKeyboard/agent/issues/792 may help you a bit in the future.

As for the firmware codebase, feel free to search for "secondaryRole", perferably in a case insensitive fashion. Please note that we can't provide help with custom implementations. We must focus on pressing issues.

p4elkin:
> A workaround I'd consider is to use a least frequently used key.

[...]

PR #193, opened 2018-10-21
re-implement secondary role support with timeout-based algorithm

p4elkin:
*Disclaimer*: the following is an attempt to experiment with the firmware and improve the behaviour of the secondary key role functionallity. This is a follow-up of the discussion started in #187. 

The changeset below re-works the mechanics of the secondary role function to the timeout-based solution. I am aware that the maintainers are quite sceptical of such an approach, but I'd still like to offer it as a food for thought because it could have a couple of advantages over the existing solution.

1 - it does not interfere with typing (at least at lot less). E.g. when 's' has a second role, quickly typing 'st' will yield 'st': typing single-role key 't' breaks the timeout of 's' and forces its first role to be triggered before 't' is typed. Current firmware's experience on the other hand is quite unpredictable sometimes (yields just 't' sometimes or secondary role of 's' activates.
2 - secondary role keys are handled independently from each other. Here's an example: 
I have a secondary role for 'a' key (mouse layer), and CMD modifier assigned as a secondary role for one of the keys on the frame. When I hold the frame key (CMD on) and merely tap 'a' I expect CMD+A emitted (select all). Current firmware however completely ignores 'a'-tap since the secondary role is already flagged by the algorithm and 'a' is discarded. 
The proposed solution tracks both keys sepately and 'a' will be sent - timeout didn't elapse.
3 -  LED indicators light up more or less instantly (currently - they won't light up while the layer key is held, you need to 'activate' it with another key tap).

Of course timeout has its own drawback: one has to hold the modifier a little bit for it to kick in, but I think this is quite essential to the user because when you trigger a shortcut you are intuitively ready to "slow-down" a bit compared to normal typing flow.

The PR is of course not ready to be merged and is very *WIP*. I am submitting it mostly as a food for thought and would love to discuss it if it is of any interest. In the worst case - feel free to decline it, I would still be the happy user of the patch since it makes my personal UHK experience finally quite perfect =) 

*commit details:*
- introduce structure for currently pressed key info
- merely collect the currently pressed keys during matrix traversal,
push single-role keys in a local array, keep track of the dual-role keys
in a global array (they last over time)
- activate secondary role only after timeout (hardcoded value of 150m)
- prevent activation of a secondary role on early key release or
single-role key tap before timeout (typing detection). Conditionally
emit primary role in some cases

Sticky modifier logic might be affected by this (though things seem to work fine so far), also generally the code is quite drafty and should be refactored more for better resource-efficiency and read-ability.


luteijn:
Nice, at least one other person is interested in this, me. I've got most of my modifiers as secondary keys and especially shift (on space) tends to get hit when typing on the fast side, causing a spaceToBeEatenAnd things to be writtenInUnintendedCamelCase...

kareltucek:
I am also interested in this. I find the idea of using home row keys as modifiers quite interesting since it solves multiple ergonomy problems I've been struggling with. However, I am skeptical about predictability of behaviour of such delay mechanism too...

Could you upload a firmware release package for testing please? (Run /firmware/scripts/make-release.js package via node, then upload the uhk-firmware...tar.bz2 from the same directory.)

kareltucek:
I am also interested in this. I find the idea of using home row keys as modifiers quite interesting since it solves multiple ergonomy problems I've been struggling with. However, I am skeptical about predictability of behaviour of such delay mechanism too...

Could you upload a firmware release package for testing please? (Run /firmware/scripts/make-release.js package via node, then upload the uhk-firmware...tar.bz2 from the same directory.)

kareltucek:
I am also interested in this. I find the idea of using home row keys as modifiers quite interesting since it solves multiple ergonomy problems I've been struggling with. However, I am skeptical about predictability of behaviour of such delay mechanism too...

Could you upload a firmware release package for testing please? (Run /firmware/scripts/make-release.js package via node, then upload the uhk-firmware...tar.bz2 from the same directory.)

p4elkin:
The firmware is attached (I've had to wrap the .bz2 archive into a zip, cause of gh's attachement mime-type filters).

Feel free to try it out! Currently the sec role timeout is hardcoded to 150ms, which seems to go nicely with my typing habbits. When i have time/ if needed I can upload the variations with 100/200ms timeouts.

All the feedback will be much appreciated.

p.s. also in order to run the make-release script, I've had to install typescript and ts-node modules globally by hand (dunno if it is okay with the npm infra or maybe should be handed by the module.json).

p4elkin:
The firmware is attached (I've had to wrap the .bz2 archive into a zip, cause of gh's attachement mime-type filters).

Feel free to try it out! Currently the sec role timeout is hardcoded to 150ms, which seems to go nicely with my typing habbits. When i have time/ if needed I can upload the variations with 100/200ms timeouts.

[uhk-firmware-8.5.3.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2500516/uhk-firmware-8.5.3.tar.bz2.zip)

p4elkin:
The firmware is attached (I've had to wrap the .bz2 archive into a zip, cause of gh's attachement mime-type filters).

Feel free to try it out! Currently the sec role timeout is hardcoded to 150ms, which seems to go nicely with my typing habbits. When i have time/ if needed I can upload the variations with 100/200ms timeouts.

[uhk-firmware-8.5.3.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2500516/uhk-firmware-8.5.3.tar.bz2.zip)

p4elkin:
*UPD*: the attachment has been re-uploaded (the first one was an accidental snapshot of master).
 [uhk-firmware-8.5.3-issue-187.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2500705/uhk-firmware-8.5.3-issue-187.tar.bz2.zip)

kareltucek:
Ok, I've got first bug for you.

Consider the following UserConfig file 
[UserConfiguration.json.txt](https://github.com/UltimateHackingKeyboard/firmware/files/2500858/UserConfiguration.json.txt)

Press and hold 's' on the (default) "MA" map ;-). I expect it to send 's', instead, cursor moves.

p4elkin:
could you pls re-upload the config file, seems like smth went wrong with the current link.

p4elkin:
ok, github seems to be back to life and I can see the config - indeed, there's a bug, the 's' shouldn't execute the mouse-layer functions while held as a switch (that should be fairly easy to fix though I am afraid there might be other glitches like this - the state management in C language doesn't seem like an easy thing to do =)

p4elkin:
I quickly patched the algorithm, testing it now, the glitch seems to have gone (trying to see if any side-effect is introduced, oh wonderful world w/out unit tests =))

[uhk-firmware-8.5.3-issue-187-2.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2501350/uhk-firmware-8.5.3-issue-187-2.tar.bz2.zip)

p4elkin:
Here's another iteration - I played with timeouts a bit more: reduced the "typing" timeout to just 120ms, but increased the press/release sec role timeout to 300ms, so that if only the sec-role key was held for some time without any interference - it'll emit the first role within 300ms.

[uhk-firmware-8.5.3-issue-187-3.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2501413/uhk-firmware-8.5.3-issue-187-3.tar.bz2.zip)

kareltucek:
Github has been having problems since yesterday (see https://twitter.com/githubstatus )... I had to wait for ~ an hour before being able to download your firmware uploads.

Anyway, I've been playing with the uhk-firmware-8.5.3-issue-187.tar.bz2.zip version and I've noticed that now the primary action often kicks in when not desired/anticipated - exactly in those case where I don't wait for the delay. I have built muscle memory for many keys like arrow keys, backspace/del, enter, esc via case buttons. I don't want to have to slow down in these cases.

My conclusion is that implementing this via setting simple timeout on activation is not very fortunate.

Thinking a bit more about this, I think a solution might be to resolve this depending on release of the secondary role button. Pseudo code would look like this:

    time secondaryRoleActivated;
    bool secondaryRoleInTimeoutPhase;
    bool secondaryRoleActive;

    onKeyPress(key) {
      if(key has secondary role)
      {
        secondaryRoleInTimeoutPhase = true;
        secondaryRoleActivated = now();

        sleep for timeout or until released;
        if (still active)
        {
          secondaryRoleInTimeoutPhase = false;
          secondaryRoleActive = true;
        }
        else
        {
          produce primary action
        }
      }
      else{
        if (secondaryRoleInTimeoutPhase)
        {
          sleep for secondaryRoleActivated + timeout - now();
        }
        activate the action according to currently active secondary roles, etc.
      }
    }

    onKeyRelease() {
      deactivate secondary role
    }

Another way - more reliable, but also more intrusive way would be to state that secondary role release has to happen after action key release in order to activate secondary role.

p4elkin:
Can you try the uhk-firmware-8.5.3-issue-187-3.tar.bz2? cause I've been experimenting with the timeouts a little bit there - might improve your experience, since I reduced the sec-role activation timeout significantly.

As for your suggestion - the algo I implemented kinda works in a similar way (need to think a bit more for the differences). Also, I'll maybe provide the schematic outline of my logic (not get confused myself =)).

p4elkin:
Here's btw another build with  a very short timeout that surprisingly doesn't still interfere with my typing that much (my speed is fairly moderate 45-50 wpm), and triggers the sec role super fast - just in 100ms.

[uhk-firmware-8.5.3-issue-187-100ms-timeout.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2501661/uhk-firmware-8.5.3-issue-187-100ms-timeout.tar.bz2.zip)

kareltucek:
At version 3: Yes, the problem still persists. I think that I actually make longer delays between writting key presses than between mod+action key presses. Since your algorithm seems to rely on delay between keyDowns of the two keys, I think that simply lowering timeouts won't make the issue go away.

I.e., I expect this feature to work (i.e., activate the secondary role+action) in the following scenario:
- almost simultaneous press on secondary-role-key and then action key 
- immediate release of the action key
- possibly long pause
- release of the role key

I cannot download the newest firmware atm, will get back to it later...

p4elkin:
yeah, my pattern is a bit different - I always press the switcher slightly before the action key is down, giving it some time (got used since Karabiner and later Hammerspoon times, both of which required such behavior). 

What I can try to do is to handle the case when the keys are pressed really together - that is a pattern different from typing (where there's usually a detectable delay between key presses).

> Since your algorithm seems to rely on delay between keyDowns of the two keys

true, it works that way, but I believe it has to in order to send the char in case I am typing.

One possible way out is to try to decide what to do with the action key once it is released: if the sec role key is also released by then - emit chars, if sec-role key is still held - emit the modded action

kareltucek:
> What I can try to do is to handle the case when the keys are pressed really together - that is a pattern different from typing (where there's usually a detectable delay between key presses).

I think that you cannot rely on any pattern which consists purely of keyDowns. I think that in order to introduce the functionality in a way which will not interfere with other user's expectations (or at least mine), you need to take into account the releases too somehow.

p4elkin:
Yeah, added similar thought to my previous comment

> One possible way out is to try to decide what to do with the action key once it is released: if the sec role key is also released by then - emit chars, if sec-role key is still held - emit the modded action

kareltucek:
> but I believe it has to in order to send the char in case I am typing.

I don't think so. In case of typing, you can still afford to postpone keypresses as long as you keep them in correct order. Of course, postponing actions may ba annoying for gamers.

p4elkin:
yeah, since I anyway introduced the buffers to track the states of currently active keys, I'll just have to turn the action key buffer into a global one (like the one for the sec-role keys) and then sprinkle in a little bit more fuzzy logic in the blend. 

Thanks for the suggestions, Karel, pls stay tuned for the updates.

kareltucek:
Sure, I will!

p4elkin:
I’ll take a look as soon as github comes back to life. At the moment I can’t comment, view or download anything ¯\_༼ᴼل͜ᴼ༽_/¯

p4elkin:
@kareltucek here's the first draft of the re-implemented algorithm (indeed I like it better than my previous version):
- the secondary role keys are tracked and triggered if any primary role key is released while they're held.
- primary role keys are either triggered immediately (no pending secondary roles are there currently) or blocked until release, in which case if the sec role is still held - it is applied, or the first role is emmited.
- finally, if one holds the sec role for a little bit without no action - it triggers the activation automatically.

This way the secondary role can indeed kick in pretty much immediately.

The algorithm looks much cleaner than before (imo), but I didn't have much time to test it (typing with it being on tho and it looks okay most of the time, though I see some room for improvement still). 

[uhk-firmware-8.5.3-fuzzy-issue-187-4.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2515456/uhk-firmware-8.5.3-fuzzy-issue-187-4.tar.bz2.zip)

kareltucek:
Hm... interferes a lot with my writing now. 

Writing "test" repeatedly with secondary role on 's' mapping to a layer with capital letters bound (appx. 50-70 wpm writer here, probably faster on these simple words): 

    teest tetest teest teEt teest teest teest teETt
    teETt teET teETT test teET teETTE teest tetest teETt teset teest tetEt teETt teest teest teEt 
    teest teETt teET tetest teEt teETt teEt teest teET tetEt

Also, I get quite a lot of key chatter (unwanted double characters) even on keys without aaaany secondary role.

[UserConfiguration.json.txt](https://github.com/UltimateHackingKeyboard/firmware/files/2515613/UserConfiguration.json.txt) (cycle to M2T via Mouse + '+', then write... Also, while cycling via the shortcut,  you will notice a bug regarding keymap switching and role activation.)

Also, I have noticed that now activation of secondary roles suffers a short lag... I didn't realize this consequence in my proposal...

p4elkin:
yeah, this unwanted chatter seems to be caused by the debouncing logic interfering with the algorithm (probably), I'll try to eliminate that when I have a chance.

mondalaci:
@p4elkin Thank you for your contribution, and sorry for the late follow-up!

https://github.com/UltimateHackingKeyboard/agent/issues/792 is planned to be implemented both in Agent and in the firmware, but #187 is currently not planned. Also, the amount of change in your commit seems to be much more than necessary which is probably due to the algorithm you implemented, I assume.

Given the above, I'll probably end up closing this PR eventually. I know that you deeply care about this feature, and I don't want to be seen as ungrateful, but I wanted to make this clear.

p4elkin:
@mondalaci thanks for clarification - no worries whatsoever! For me the ability to be able to sneak peek and dabble with the source is one of the biggest perks of UHK actually, the code base is super understandable and a great start for those who hardly worked with embedded stuff before.

I can carry on working on this in my fork (cause indeed the algorithm got quite forked anyway) and maybe post the incremental improvements in the #187/#792 discussion. If @kareltucek and @luteijn are still interested - they can follow it there.

 btw, @kareltucek I'm attaching the latest state of the firmware, which is supposedly a lot less chatty (found some reasons for that in my impl). Still seeing some chattiness occasionally though, but a lot less than before. I've split the implementation into three parts (corresponding to three possible states of the keyboard) and that helped quite a bit to debug and to understand what is going on.

Also it handles the deferring of the key events and ordering better - in theory it should never mix the chars when sending them. The work is stilll ongoing though: the biggest current problem i have is that sometimes (quite rarely for me, but still) the 'hold modifier and wait' state falls into 'just type the all the stuff' state, causing e.g. 'ssssssss' to be typed. I am yet to figure out that.

[uhk-firmware-8.5.3-issue-187-5-not-so-chatty.tar.bz2.zip](https://github.com/UltimateHackingKeyboard/firmware/files/2524337/uhk-firmware-8.5.3-issue-187-5-not-so-chatty.tar.bz2.zip)

UPD: after testing this for a day at work - I see quite some chattiness at random moments and I probably know how to fix that (try to stay away from trying to extend the 'pressed' state programmatically, since it backfires occasionally).

that being said - feel free to close this onef

mondalaci:
@p4elkin Sure, feel free to comment to the relevant issues from the point I close this PR.

kareltucek:
@p4elkin yep, I won't miss any followups at either location. I am still interested in this if you are confident that you can make it work well (i.e., without having users relearn their customs in non-intuitive manner). On the other hand, I am not going to go too far out of my way in order to have this feature (e.g., if it turns out that it is not possible to do in a non-buggy fashion, if your branch is not merged into official master and if it can't keep up with official upstream...).

It seems I can't unzip the last version 5, but if it doesn't work well for you, I will rather wait for the next followup.

p4elkin:
@kareltucek fair enough, thanks again for helping and for feedback. In the meantime - I've found some time to fiddle with the algorithm this week and have created a tentative pre-release version (rebased on current canonical master). The link's here: https://github.com/UltimateHackingKeyboard/firmware/issues/187 . I suggest to move the duscussion to the issue thread.

PR #200, opened 2018-11-18
Feature firmware forks in readme

kareltucek:
So, here we go. Is it ok this way? :-)


mondalaci:
Looks great! Thanks! :)

kareltucek:
Thanks goes to you! ;-)

PR #210, opened 2019-03-06
add https://github.com/p4elkin/firmware to the list of featured forks

p4elkin:
this changeset proposes to add https://github.com/p4elkin/firmware to the list of the custom firmware forks.


mondalaci:
Thank you!

PR #215, opened 2019-03-31
fix wrong type

kareltucek:
...


mondalaci:
Great catches! Thanks!

kareltucek:
Not so great after all. The two enums are not mapped the same way, which means that this PR breaks the whole thing :-(.

kareltucek:
Sorry, ignore the deleted commend, I just need to get some sleep :-).

PR #217, opened 2019-04-01
Fix play of empty macro

### LINKED ISSUE #216: Empty macro playback bug
Just created an empty macro accidentally, and when I tried to play it back, it switched to another virtual terminal. I'm not sure what's happened exactly, but empty macros don't seem to be handled properly.

kareltucek:
fixes #216


mondalaci:
Thanks so much!

PR #247, opened 2019-11-19
Fix debouncer related data races

### LINKED ISSUE #245: Key states of the left half aren't synchronised enough with debouncer logic
The usb report updater logic is totally in sync with the right half state: the loop in`main.c` will first scan the right key matrix state and the updater will read it before running yet another iteration of report update. 

However, the left half state is synchronised in `UhkModuleSlaveDriver_Update` function via async I2C communication (`I2C_MasterTransferNonBlocking` API), which [to my limited knowledge, received mostly from reading the API docs and tutorials like [this](https://www.allaboutcircuits.com/technical-articles/using-interrupts-on-arduino/)] may do the following:

> ...an Interrupt  interrupts whatever the processor is doing, and executes some code designed to react to whatever ... is being fed to the Arduino. Once that code has wrapped up, the processor goes back to whatever it was originally doing as if nothing happened!

In case of debouncing this could e.g. mean that the debouncer ensures the key state is set to non-active, but the rx message may update it to active (bounce!) and override bebouncer's efforts for the rest of the USB report updaters iteration logic. 

The current state of the firmware does not seem to obviously suffer from the issue due to the way the algorithm designed (the key states are rarely checked, each key is synched with report right after its bebouncing is handled).

However, if we synthetically make the algo a bit more complex (though still legit) by increasing the debounce timeout and by first handling the bouncing state for all the keys before propagating them to the reports (see this branch: https://github.com/p4elkin/firmware/commits/simulate-sync-issues), we may observe that while the right half will stably produce one key stroke per 5 seconds, the left may allow a series of identical chars over the same period.

One workaround would be to not write the RX update directly into the `KeyStates` matrix, but rather to some intermediate structure and then update the left half's `KeyStates` portion in the USB report updater iteration (pretty much the same way the right half is updated). This will ensure that during the report update phase the key state matrix is completely intact regardless of the algo complexity and the amount of times the current state is tested. (see this commit for reference https://github.com/p4elkin/firmware/commit/573a2ef1dcbd0a9b60fa8e9531005c6a1988fd72).

I used this workaroung for my fork and stopped experiencing chattiness _completely_!

Kudos to @kareltucek for coining in the insights for the investigation (see the discussion in another ticket: https://github.com/UltimateHackingKeyboard/firmware/issues/207#issuecomment-546341421)

kareltucek:
Fixes https://github.com/UltimateHackingKeyboard/firmware/issues/245

Please, give it some testing - there is some chance that I have overlooked some interference with secondary roles and with debouncing. I've done basic test of secondary roles and basic debouncer test with 500ms timeout.

### DIFF
--- a/right/src/key_states.h
+++ b/right/src/key_states.h
@@ -9,10 +9,24 @@
 
 // Typedefs:
 
+    /*
+     * Next is used as an accumulator of the state - asynchronous state updates
+     * are stored there. The next always contains the most up-to-date
+     * information about hardware state of the switch.
+     *
+     * `Previous` and `current` are computed from `next` by "debouncing"
+     * algorithm.  Especially values (0, 1) signify that key has been pressed
+     * right now and an action (e.g., start of a macro) should take place.
+     *
+     * Debouncing flag & timestamp are used by debouncer to prevent the value
+     * of current from changing for next 50 ms whenever the key state changes.
+     * */
+
     typedef struct {
         uint8_t timestamp;
         bool previous : 1;
         bool current : 1;
+        volatile bool next : 1;
         bool debouncing : 1;
     } key_state_t;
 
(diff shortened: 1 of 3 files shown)


p4elkin:
lgtm 👍

mondalaci:
Thank you so much!

Issue #246, opened 2019-11-18
Secondary roles - possibility of debouncer skips


kareltucek:
If I read the code for current implementation of secondary roles correctly (never did that before), then the second key of a shortcut (i.e., first key is a key with secondary layer role, second key is an action key on the referenced layer) disables the key for current debouncer timeframe on transition from `SecondaryRoleState_Pressed` to `SecondaryRoleState_Triggered` - i.e., lets the debouncer lock on `(previous, current) == (0, 0)`. The key is then reactivated again after debouncer runs out.

That means that if the second key is pressed for less than 50 ms, or if the two keys have less than 50ms overlap, then the second keystroke will be ignored.  Also, the key activation is delayed by 50ms, which gamers could be pretty angry about.

As far as I my statistics are correct, the press time of a keystroke can usually fit within 90-250ms, but still - can we rule out that a professional writer or a user with a special type of switches cannot occasionally fit under 50 ms? Could anyone mind an unwanted 50ms delay? And even if we don't mind - Is the requirement of 50ms time overlap a bug or a feature?

Btw., this could simply be fixed by disabling the `KeyState->debouncing` flag at the transition...

kareltucek:
Also, assume that a secondary role key is pressed and then after some delay two different action keys are pressed in the same update cycle. When second one is processed, the state is already `Triggered` so the action is activated immediately, even though the layer has not been activated yet. I.e., it will produce action from wrong layer.

This second realization could be fixed by adding another state between Pressed and Triggered... (Or by rewriting it from scratch.)

mondalaci:
Can you reproduce this issue with a higher than 50 ms timeout value? If so, this should be fixed.

The 50 ms debouncing interval is overkill. We've received reports about chatter, and this is how I tried to combat the issue in a bruteforce manner. As it turned out, the root cause of such reports was hardware related which could be fixed by spraying contact cleaner spray into the misbehaving key switches. Cherry recommends 5 ms timeout, although I'm worried that it might be too low.

This should not cause a delay however. Key presses should immediately register. It's only subsequent key presses of the pressed keys which are delayed.

kareltucek:
Yes, if I increase the release debouncer timeout to 200 ms I can clearly reproduce the first issue.

kareltucek:
Anyway, there is a quick hackfix of both problems.

mondalaci:
[...]

PR #248, opened 2019-11-19
- Fix two bugs of secondary roles

kareltucek:
Fixes both bugs mentioned in https://github.com/UltimateHackingKeyboard/firmware/issues/246


mondalaci:
[Simple fix](https://github.com/UltimateHackingKeyboard/firmware/issues/246#issuecomment-560168636) implemented in b6511bb8947dd0bfa35eaaf1b0ed47b0cf24df5b

Issue #209, opened 2019-02-01
Start scroll timer on keydown


robobenklein:
Currently it's possible to quickly tap a scroll key and not have any scroll event registered, which I think is caused by the repeating scroll timer running independently of the keypress. (Or the trigger just not happening fast enough.)

Could it be changed so that pressing the key down will start the scroll repeat timer (and thus immediately relay a scroll event) instead of having to hold down the scroll key long enough for a scroll timer tick to occur?

Hopefully that's a clear enough explanation, but I can draw a diagram or something to help explain what it feels like is happening.

FW 8.5.3

![image](https://user-images.githubusercontent.com/5423266/52143223-e429c200-2628-11e9-9842-4f17c7de18f5.png)
For example, adding an option here to ensure that the scroll event happens at least once immediately on keydown.

mondalaci:
@robobenklein I agree that what you're suggesting is the correct behavior, and we'll fix the firmware to behave this way. Good catch!

robobenklein:
Just a bump that this is still the case in 8.6.0

Is this something that's relatively easy to fix or should I potentially dig into the code myself during my next break?

mondalaci:
The relevant code has not been changed in recent firmware releases.

See https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/usb_report_updater.c#L60

There's a lot on our plate, and I can't help much. I'll be able to fix this pretty quickly once I get there.

mondalaci:
Fixed by #249.

PR #249, opened 2019-12-23
Fix first tick of scroll timer

kareltucek:
Should solve #209.

Conditions at 130 && 150 apparently handle the first update cycle tick of scroll event. Clauses at 130 & 151 defaulted to 0, due to zeroing at 108. (Distance is always zero in the first tick since elapsed time is also zero.) I believe that `copysignf(1.0, kineticState->ySum)` was unreachable.


mondalaci:
@kareltucek Excellent job! Please use a [close keyword](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) in upcoming PRs, so that the relevant issues will be automatically closed upon merging them.

PR #250, opened 2019-12-23
Refactor keyState queries to be more descriptive

kareltucek:
Should increase readability of the code, since this way it should always be clear *when* the condition is actually met.

### DIFF
--- a/right/src/key_states.h
+++ b/right/src/key_states.h
@@ -14,8 +14,6 @@
     #define KEYSTATE_INACTIVE(KEY) (!((KEY)->current))
     #define KEYSTATE_ACTIVATED_NOW(KEY) (!(KEY)->previous && (KEY)->current)
     #define KEYSTATE_DEACTIVATED_NOW(KEY) ((KEY)->previous && !(KEY)->current)
-    #define KEYSTATE_ACTIVATED_EARLIER(KEY) ((KEY)->previous && (KEY)->current)
-    #define KEYSTATE_DEACTIVATED_EARLIER(KEY) (!(KEY)->previous && !(KEY)->current)
 
 // Typedefs:
 
--- a/right/src/key_states.h
+++ b/right/src/key_states.h
@@ -9,12 +9,13 @@
 
 // Macros:
 
-    #define ACTIVE(KEY) (((KEY)->current))
-    #define INACTIVE(KEY) (!((KEY)->current))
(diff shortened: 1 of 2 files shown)


mondalaci:
I like your macros, and I think they're an improvement when it comes to readability. They should be prefixed with `KEYSTATE_` however, because general names like `ACTIVE` can easily lead to name collisions in the future when including other files.

I've been actually considering to change the coding style of macros as all caps like `KEYSTATE_ACTIVE()` is rather hard to read and `KeyState_Active()` would be easier on the eyes. Unless you have objections, let's use the new style from this point forward. We'd change the rest of the macros later according to the new style.

I'm slightly worried that the extra conditions introduced by these macros will make the code slightly slower. Chances are the GCC optimizations currently used for the release version will make this a non-issue, but it'd give me a peace of mind if you could check this.

kareltucek:
> They should be prefixed with KEYSTATE_

Sure.

> I've been actually considering to change the coding style of macros as all caps [...]

Well, there are many good reasons to want to know that a macro is a macro when you see it, and therefore I would rather not violate the convention... 

One more option (maybe cleaner one) is to use inline functions in this case rather than macros? (Here I am slightly worried if the inlining will work just as well as macros regarding performace, but theoretically it should.)

> I'm slightly worried that the extra conditions introduced by these macros will make the code slightly slower.

I don't think this is a problem - even if there are some additional instructions, they are cheap instructions inlined with other code in easy-to-optimize manner.

The only condition which is at top call-level and is not same as before is the one at 441. Admittedly, the generated asm is different, even though I believe that compiler could legaly optimize it out. However, the line is secondary-role related, which means that it will probably be removed soon anyway.

kareltucek:
Well, I see I've made a mistake on 324, which means that the line also contains one additional check. 

Still this is marginal. If you feel performance of these parts of those code to be an issue, please create a new ticket, and I will look at how to optimize the control flow once the open problems are solved. I am sure much more can be saved there, but no point in optimizing it before needed changes are finished.

mondalaci:
I've been thinking about using inline functions, too. By doing so, looks like we could avoid [various issues](https://www.programmerinterview.com/c-cplusplus/inline-vs-macro/) down the road. Please check whether inline functions actually get inlined for the release build. If so, let's use them instead of macros.

kareltucek:
They seem to be inlined allright. 

So I think that this is prepared for merge. I've made a lot of mess here with commits, so I recommend to merge it as a "squash"...

PR #251, opened 2019-12-26
Refactor of sticky keys.

kareltucek:
Did this accidentally while digging into sticky mod related tickets. Feel free to discard.

It:
- adds a few explanatory comments
- moves the logic behind stickies to one reserved place so that it is not entangled with other logic more than necessary
- adds machinery to easily match selected shortcuts

I could further add a "strategy" settings, which would allow to globaly specify `StickinessStrategy_Never, StickinessStrategy_Always, StickinessStrategy_Smart` if desired.


right/src/usb_report_updater.c
```diff
@@ -323,7 +364,7 @@ static void applyKeyAction(key_state_t *keyState, key_action_t *action, uint8_t
             case KeyActionType_Keystroke:
                 if (keyState->previous) {
                     if (slotId == stickySlotId && keyId == stickyKeyId) {
-                        if (!IsLayerHeld() && !(secondaryRoleState == SecondaryRoleState_Triggered && IS_SECONDARY_ROLE_LAYER_SWITCHER(secondaryRole))) {
+                        if(!stickyModifierShouldStick) {
```
mondalaci: Please merge these three `if` statements as one.

right/src/usb_report_updater.c
```diff
@@ -252,12 +252,54 @@ static void handleSwitchLayerAction(key_state_t *keyState, key_action_t *action)
 static uint8_t basicScancodeIndex = 0;
 static uint8_t mediaScancodeIndex = 0;
 static uint8_t systemScancodeIndex = 0;
-static uint8_t stickyModifiers, stickySlotId, stickyKeyId;
+
```
mondalaci, line 255: Please use `//` comments even for multiline comments because it usually saves a line or two.

right/src/usb_report_updater.c
```diff
@@ -252,12 +252,54 @@ static void handleSwitchLayerAction(key_state_t *keyState, key_action_t *action)
...
+static bool    stickyModifierShouldStick;
+
 static uint8_t secondaryRoleState = SecondaryRoleState_Released;
 static uint8_t secondaryRoleSlotId;
 static uint8_t secondaryRoleKeyId;
 static secondary_role_t secondaryRole;
 
+
+static bool shouldStick_actionIsSelected(key_action_t * action) {
+    if(action->keystroke.modifiers == 0 || action->type != KeyActionType_Keystroke || action->keystroke.keystrokeType != KeystrokeType_Basic) {
```
mondalaci: Add space after `if`.

right/src/usb_report_updater.c
```diff
@@ -252,12 +252,54 @@ static void handleSwitchLayerAction(key_state_t *keyState, key_action_t *action)
...
+        return false;
+    }
+
+    #define TRY_MATCH(MOD, SC) if(action->keystroke.modifiers == MOD && action->keystroke.scancode == SC) { return true; }
```
mondalaci: I appreciate the beauty of your approach, but let's not use macros unless really needed. Make this a simple hardcoded condition for the time being, and let's use an array of modifers and scancodes later if needed.

right/src/usb_report_updater.c
```diff
@@ -252,12 +252,54 @@ static void handleSwitchLayerAction(key_state_t *keyState, key_action_t *action)
...
+
+    TRY_MATCH ( HID_KEYBOARD_MODIFIER_LEFTALT, HID_KEYBOARD_SC_TAB );
+
+    return false;
+}
+
+static bool shouldStick(key_action_t * action) {
+    bool currentLayerIsHeld = IsLayerHeld() || (secondaryRoleState == SecondaryRoleState_Triggered && IS_SECONDARY_ROLE_LAYER_SWITCHER(secondaryRole));
+    bool allowAll = true;
```
mondalaci: Please don't introduce `allowAll` yet. Let's develop this further if needed, and make the code as simple as possible in the meantime.

right/src/usb_report_updater.c
```diff
@@ -266,10 +308,9 @@ static void applyKeyAction(key_state_t *keyState, key_action_t *action, uint8_t
         switch (action->type) {
             case KeyActionType_Keystroke:
                 if (action->keystroke.scancode) {
+                    /* On keydown, reset old sticky modifiers and set new ones */
```
mondalaci: Use `//`.

right/src/usb_report_updater.c
```diff
@@ -266,10 +308,9 @@ static void applyKeyAction(key_state_t *keyState, key_action_t *action, uint8_t
...
                     if (!keyState->previous) {
```
mondalaci: Let's use the new `KeyState_*` functions wherever possible.

kareltucek:
> Please make sure to follow the coding standards in light of the recently spotted inconsistencies.

@mondalaci  Just fixed all codestyle problems I could find.

If there are any more problems, please point them out. (In case you already did so, note that I apparently do not see it)

mondalaci (changes requested):

mondalaci:
@kareltucek Ah, sorry. I thought my comments were visible but they were in pending state and I had to explicitly publish them. I'll know better next time. Can you see them now?

kareltucek:
Ah, explains a lot :-). Yes, I can see them now.

right/src/usb_report_updater.c
kareltucek: AllowAll is the current behaviour (i.e., all modifiers stick). I view the shortcut-matching part as being the "new" code (which maybe should not be introduced yet).   

Or should this PR introduce the proposed "alt-tab-matching-only" behaviour?

right/src/usb_report_updater.c
kareltucek: May I ask why? 

I think this is better from readability standpoint, since it follows the hierarchy of the logic behind. I am afraid that merging them will make the condition resemble a magic formula.

mondalaci: I think it's best to keep the control flow as simple as possible. If multiple, complex conditions are involved, I'd prefer extracting the individual conditions into easy to read variable names, then and'ing these variable names in a single if condition.

kareltucek: Alright, let's try to simply merge it... I would rather not extract the condition, because it would mean inverting reading order of the entire thing, and placing the "less important" information into the higher-level scope.

right/src/usb_report_updater.c
mondalaci: Yes, please let's only introduce the Alt+Tab matching behavior. To be more precise, let's introduce Alt+Tab and Super+Tab matching, the latter being relevant on OSX. Alt = LAlt || RAlt. Please use boolean logic instead of the proposed array solution for the sake of simplicity.

kareltucek: Done, in a manner that any superset of those modifiers will match too (which I think is desirable (?)).

mondalaci: Yes, well done!

PR #252, opened 2019-12-26
Extra usb reports

### LINKED ISSUE #221: Include extra USB reports for shortcuts for maximum compatibility
Currently, UHK shortcuts are sent via USB using the smallest number of USB reports possible. For example, in the case of Ctrl+C, both the Ctrl modifier and the scancode of C gets sent in the same report at once. This is correct, but it messes up some misbehaving applications like Karabiner and Hyper-V. By holding and releasing modifiers in extra reports, this might be worked around at the firmware level:

* Report 1: Hold Ctrl
* Report 2: Hold Ctrl + send the scancode of C
* Report 3: Hold Ctrl

Hopefully, this will work out fine.

kareltucek:
Closes #221. 

Depends on the state refactor PR and therefore will need some minor fixes yet...


kareltucek:
Github refuses to update the diff-against branch/commit... Will open a new PR.

PR #253, opened 2019-12-26
Extra usb reports

### LINKED ISSUE #221: Include extra USB reports for shortcuts for maximum compatibility

kareltucek:
Closes #221


right/src/usb_report_updater.c
```diff
@@ -274,25 +274,27 @@ static void applyKeyAction(key_state_t *keyState, key_action_t *action, uint8_t
                 } else {
                     ActiveUsbBasicKeyboardReport->modifiers |= action->keystroke.modifiers;
                 }
-                switch (action->keystroke.keystrokeType) {
-                    case KeystrokeType_Basic:
-                        if (basicScancodeIndex >= USB_BASIC_KEYBOARD_MAX_KEYS || action->keystroke.scancode == 0) {
+                if(action->keystroke.modifiers == 0 || KeyState_ActivatedEarlier(keyState)) {
```
kareltucek: Just this line actually changes in this block.

kareltucek:
@mondalaci I suspect that there might be some unpublished comments too?

right/src/macros.c
```diff
@@ -348,35 +367,74 @@ bool processScrollMouseAction(void)
...
+
 bool processTextAction(void)
 {
     static uint16_t textIndex;
     static uint8_t reportIndex = USB_BASIC_KEYBOARD_MAX_KEYS;
     char character;
     uint8_t scancode;
+    uint8_t mods;
 
+    /* When all characters have been sent, finish. Yet, one last report should still be sent
```
mondalaci: Please change every comment to:

`// My comment ...`

style.

right/src/usb_report_updater.c
```diff
@@ -311,34 +311,40 @@ static void applyKeyAction(key_state_t *keyState, key_action_t *action, uint8_t
 
         switch (action->type) {
             case KeyActionType_Keystroke:
+            {
+                bool stickyModifiersUnChanged = true;
```
mondalaci: It'd be easy to comprehend a non-negated variable like `stickyModifiersChanged`.

kareltucek: Sure, however result is having to negate its every use further down the road so I am not sure if it is much of an improvement...

Issue #256, opened 2019-12-31
Diagonal mouse movement speed


greneholt:
This is more of a feature request than a bug, but one of the things I am struggling with when using mouse mode is that the cursor moves faster diagonally than straight. I believe this is because the UHK is moving the cursor in both the X and Y axes at N pixels/second, which is truly a speed of `sqrt(2)*N` pixels/second, or 41% faster than when it is moving only in the X or Y axis at N pixels/second. This causes me to miss targets far more often when moving the cursor diagonally, as my brain is confused by the sudden acceleration.

I believe the usability of mouse mode would be greatly enhanced by decreasing the X and Y velocity to `N/sqrt(2)` when moving diagonally, so the cursor has the same absolute speed regardless of direction. This would better match the way humans typically use a real mouse, since we do not think about separate velocities in the X and Y axis, but only about the speed with which we are moving our hand.

kareltucek:
I've got no idea what Laci's opinion on this is going to be, but nevertheless, feel free to try this firmware _removed_.

(You will need to unzip the file first - github does not allow uploading tar.bz2 archives.)

kareltucek:
(It is vanilla 8.7.0. with just the speed normalized - compiled from https://github.com/kareltucek/firmware/tree/diagonal_slow_movement )

greneholt:
Thank you for the quick response! I won't be able to try it until later, but I'm confused by the code change. It seems like the distance variable is modified after it has already been used to modify xSum and ySum, so I don't see how the change would have any effect...

kareltucek:
True, that happens when one writes faster than thinks...

kareltucek:
Should be better now :-). [uhk-firmware-8.7.0.diagonal_movement_slow.zip](https://github.com/UltimateHackingKeyboard/firmware/files/4013591/uhk-firmware-8.7.0.diagonal_movement_slow.zip)

greneholt:
Thank you! That works perfectly, and feels so much better in my opinion.

pcooke9:
I'm not a coder, so I don't know if this is related or not. It's just something that caught my attention a while back. I noticed that when holding the mouse left, and then hitting right while still holding left, the cursor doesn't change direction. Doing the opposite (right to left) does change cursor direction. The same goes for up & down. It occasionally effects my accuracy if I don't lift my fingers enough before changing direction.

I have Agent v1.3.0 and firmware v8.7.0 installed.

greneholt:
[...]

PR #257, opened 2020-01-01
Fix symmetry of mouse movement behaviour

kareltucek:
Solves symmetry problem mentioned in #256 .

I am afraid that this adds a bit of complexity into the code, unfortunately I don't see any way around.

Basically, I am extracting "sign" related switches from 115 and 136 into their own functions, and adding them into the kinetic state.

Please test that up&down scrolling works as expected on Mac.

### DIFF
--- a/right/src/usb_report_updater.c
+++ b/right/src/usb_report_updater.c
@@ -35,6 +35,8 @@ mouse_kinetic_state_t MouseMoveState = {
     .downState = SerializedMouseAction_MoveDown,
     .leftState = SerializedMouseAction_MoveLeft,
     .rightState = SerializedMouseAction_MoveRight,
+    .verticalStateSign = 0,
+    .horizontalStateSign = 0,
     .intMultiplier = 25,
     .initialSpeed = 5,
     .acceleration = 35,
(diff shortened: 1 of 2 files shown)


right/src/usb_report_updater.c
```diff
@@ -57,6 +61,52 @@ mouse_kinetic_state_t MouseScrollState = {
...
+static void updateDirectionSigns(mouse_kinetic_state_t *kineticState) {
+    updateOneDirectionSign(&kineticState->horizontalStateSign, -1, kineticState->leftState, kineticState->rightState);
+    updateOneDirectionSign(&kineticState->horizontalStateSign,  1, kineticState->rightState, kineticState->leftState);
+    updateOneDirectionSign(&kineticState->verticalStateSign, -1, kineticState->upState, kineticState->downState);
+    updateOneDirectionSign(&kineticState->verticalStateSign,  1, kineticState->downState, kineticState->upState);
+}
+
+// Called on keydown of mouse action. Direction signs ensure that the last pressed action always takes precedence, and therefore
+// have to be updated statefully.
+static void activateDirectionSigns(uint8_t state) {
```
mondalaci, line 81: This function could be much simpler by not breaking for every case, but gathering 1 and -1 cases.

kareltucek: I don't see it. How?

mondalaci: It cannot be done, indeed. Sorry, my bad!

kareltucek: :-(

Issue #264, opened 2020-01-17
Key identification


kareltucek:
`usb_report_updater` commonly identifies keys by a pair of `slotId, keyId`. This is quite clumsy if functions get broken down, because one has to pass them as separate parameters, and save them separately too. 

What is your opinion on using the corresponding `key_state_t* keyState` pointer instead since it is always available and carries same information? Would it be desired, or would it be considered as "conceptually unclean" solution?

kareltucek:
Closing since this proposal got merged with #265.

Issue #260, opened 2020-01-04
Secondary role refactor.


kareltucek:
Finally, here is the proposal on replacing the secondary role implementation. Originally, I wanted to do this together with layer handling refactor, but I think it makes sense to at least discuss the two issues separately (this text is long enough as it is). Due to integration reasons, I will need to look into layers first though. (At the moment, I have a clear idea on what to do with secondary roles, yet I am still unclear on layer handling, so I will continue there with research in a few days time.)

So long story short - I would like to extend the keystate preprocessing by one more phase using a new module called Postponer. Its purpose would be to allow any keyboard action initiate "postponing mode", in which key activations are queued rather than activated straight away. The postponer would then be used to implement secondary roles via a simple "secondary role driver". 

The integration would be pretty minimalistic - just a few hooks in usb_report_updater, and the rest of the logic would be placed in a separate files, hopefully not adding much overall complexity and slightly simplifying the usb_report_updater.

In more detail, there are the following new modules (i.e., new files):

Postponer
---------
General purpose: to allow postponing new key activations until (e.g.) some condition is met. 

Specific usecase: this will allow secondary roles to intercept keydown events of other keys. If desired, this can allow implementation of more advanced secondary role resolution strategies, and of other things too. 

Principle: 
- The new module would contain an event queue (circular buffer), where postponed events would be stored. Postponer would be always either in an active or an inactive state. When inactive, `preprocessKeyState` would function just as now - when `keyState->next` changes, it is written into `keyState->current`. When active, `preprocessKeyState` would pass the state changes to the postponer instead. The postponer would save them into the queue, and later replay them into `keyState->current`. 
- Postponer would be active whenever there is an unprocessed event in its queue, or if someone asked it to be active in previous update cycle. 

Needed changes:
- Keystate_t would have to be extended by one more state - lets rename `next` to `hardwareSwitchState` and lets call the new state `debouncedSwitchState` - since `previous` and `current` will become detached from the hardware state of the switch. (Postponer will be inserted in the transition between `keyState->debouncedSwitchState` and `keyState->current`.)
- A few hooks would be placed at various places in report updater.

Pros & Cons:
- (+) It is a general solution which allows (simple) implementation of more advanced features, such as the resolution strategy for alphabetic keys.
[...]

Issue #254, opened 2019-12-29
Extra USB reports are omitted for keys with enabled secondary roles.


kareltucek:
Primary action of secondary role key currently lasts one tick only. This should be dealt with - either as a part of bigger refactor, or separately.

For the time being, extra USB of #221 reports are omitted for this case.

    if(stickyModifiersUnChanged || KeyState_ActivatedEarlier(keyState) || action->keystroke.secondaryRole) 
                                                                          ^^

kareltucek:
Resolved by #265. Now, the keypress takes two cycles, which allows standard way of extra modifier activation.

Issue #207, opened 2019-01-29
Secondary role issue


cbucknell:
- Firmware: 8.5.4
- Agent: 1.2.12
- OS: Windows 10

To replicate this issue, assign the "Mod" layer as a secondary role to the right spacebar, using Space as scancode (default). This enables the use of the arrow cluster with one hand, while still being able to use the right spacebar as Space when tapping.

Issue: Whenever space and another key are quickly pressed alternately, most spaces are not registered (e.g., alternating "a" and space: _aa a a aa a a aaaaaa_). Issue can be also replicated when typing very fast since some spaces are not registered between words. This issue is solved when the secondary role is deactivated (e.g. _a a a a a a a a a_).

Additionally, when enabling the secondary role through a third party software (TouchCursor: https://martin-stone.github.io/touchcursor/) and disabling the secondary role in the UHK, the issue disappears. This suggests that the issue is firmware related.

mondalaci:
@kareltucek @eltang Can you guys reproduce this, or have an idea about the root cause? This hasn't ever been reported yet.

kareltucek:
If I understand the issue correctly, it's the same old good one which has been discussed between me and @p4elkin ( https://github.com/UltimateHackingKeyboard/firmware/pull/193 and  https://github.com/UltimateHackingKeyboard/firmware/issues/187 ). 

Basically, when writing fast, the keypresses overlap in time. I.e., when writing " a", the a is pressed prior release of the spacebar which therefore acts as a layer switch instead.

So yes and yes. Unfortunately, I do not have a good idea about a proper solution... 

/E: removed the no-longer-relevant paragraph

@cbucknell, would you mind investigating and describing here how the touchcursor solves the resolution? I didn't manage to get it running in a virtual machine...

/E: sorry for the edit mess... I hope that email notifications are not fired at every update?

cbucknell:
@kareltucek managed to find the source code of TouchCursor: https://github.com/martin-stone/touchcursor -- is it useful?

kareltucek:
Ok, once again, seems like I've let my expectations mislead me into a terribly bad analysis.

I went through the source code and tried the program on an old windows box. The touchcursor solution does not contain any explicit delays at all - it simply resolves the situation depending on the first key up event (if the "layer switcher" is released first, it is taken for primary action, otherwise for secondary). Activation of the "action key" seems to be postponed until its release. If both keys are held without release, it seems that windows repetition hook kicks in on its own.

[...]

PR #265, opened 2020-01-17
Rewrite of secondary roles

### LINKED ISSUE #228: Can't use two secondary roles at the same time
I configured my UHK with secondary roles on F (LCtrl) and D (Shift).
I'm not able to select words by clicking Shift + LCtrl + RightArrow.
UHK Agent 1.2.12 on Windows

kareltucek:
So, rough list of changes:
- added postponer
- renamed keystate->next to keystate->hardwareSwitchState and added keystate->debouncedSwitchState
- integrated postponer on transition between the keystate->debouncedSwitchState and keystate->current (search for "PostponerCore" in usb_report_updater.c)
- added secondary role driver
- moved keystroke code from applyAction to applyKeystroke and added one indirection to decide between primary and secondary "keystroke" scenario. (Integrated secondaryRoleDriver here.)
- removed old secondary role hooks (updateActiveUsbReports simplified)
- "inverted" keystate handling of applyKeyAction. The switch is now carried out whenever the state is non-zero rather than if it is active. This allows key deactivation handling at the same place as the corresponding activation, rather than in a separate switch. 
- applied proposal from #264. Can revert any time. 

Unfortunately, the last three changes have made the diffs pretty unreadable, which is why I offered postponer for separate revision yesterday.

Closes #228, #192, #254, #260.

#207 and UltimateHackingKeyboard/agent#792 will be taken care of separately.


kareltucek:
Btw, I gathered some performance measurements which might be of interest - or not. Feel totally free to ignore.

TL;DR Performance of this PR is good. Besides this PR, I have noticed increase in time on the keystate refactor merge (originally, I have been trying to investigate the freezing problem).

One update cycle takes on average:
- on pure v8.7.0, 0.4ms/cycle
- on 8.7.0 with keystate refactor via inlined functions 0.7ms/cycle (!)
- on 8.7.0 with keystate refactor via macros (exactly same expressions) 0.4ms/cycle
- on this PR with keystates via inlined functions 0.3ms/cycle
- on his PR  with keystates via inlined functions and non-volatile keystate 0.2ms/cycle
- on his PR  with keystates via macros 0.5ms/cycle

Removing `volatile` from `keystate->next/hardwareSwitchState` does not seem to have any significant effect (under 0.1ms) - sometimes time decreases but sometimes increases. 

As seen above, using inlined functions vs macros yields drastically different results even if the code is "same", however it differs whether macros or inlined functions are faster. I assume that language semantics might slightly differ (e.g., I might expect an inlined function to act as a memory barrier), still I am quite surprised by macro version being significantly slower in some cases.

Furthermore, sending a set of USB reports (measured on usb testing mode, so sending mouse report + basic keyboard report every cycle) takes around 4ms/cycle.

If desired, I can integrate simple stopwatch into firmware (it is a short snippet, just as the one which controls the "simulate keypress" option). Then LED display shows time, updated once per 1000 cycles.

right/src/postponer.c
```diff
@@ -0,0 +1,91 @@
...
+
+//#######################
+//### Query Functions ###
+//#######################
+
+
+uint8_t PostponerQuery_PendingKeypressCount()
+{
+    uint8_t cnt = 0;
+    for ( int i = 0; i < buffer_size; i++ ) {
```
mondalaci: Is this supposed to be `uint8_t` instead of `int`?

right/src/postponer.c
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return cnt;
+}
+
+bool PostponerQuery_IsKeyReleased(key_state_t* key)
+{
+    if (key == NULL) {
+        return false;
+    }
+    for ( int i = 0; i < buffer_size; i++ ) {
```
mondalaci: Is this supposed to be `uint8_t` instead of `int`?

right/src/postponer.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+// Includes:
+
+    #include "key_states.h"
+
+// Macros:
+
+    #define POSTPONER_BUFFER_SIZE 32
+    #define POSTPONER_BUFFER_MAX_FILL (POSTPONER_BUFFER_SIZE-5)
```
mondalaci: Why `-5`? Extracting -5 to a properly named macro may be the best way to explain it.

right/src/postponer.c
kareltucek: Yes.

right/src/postponer.h
kareltucek: Quite arbitrary number. Added some explanation...

mondalaci:
@kareltucek I've been thinking about this PR recently, and I'm wondering if postponing is truly necessary. It makes application logic quite a bit more complex, and I don't believe it's the only way. The recursive resolution of secondary roles should be possible without it. Would you please elaborate about this?

kareltucek:
Yes and no. 

If we talk on abstract level, then the secondary role is an action whose behaviour depends on something that is going to happen "in future". So in this sense yes it is necessary, and doing so by a general mechanism is a clean way to implement it. 

The old secondary role mechanism basically does the same thing, but instead of delegating the logic and complexity into separate units, it achieves similar result by placing special branchings all over the place.

Having full-fledged postponer is not necessary for the simple resolution strategy because there only the "secondary role" needs to be postponed. Advanced strategies however need postponing of the primary role too so they are not possible without the general mechanism.

However, under more scrutiny, the even the simple case runs into (more and more) problems with the "simple" implementation - the old implementation is prone to race conditions, and is pretty much incompatible with the extra usb reports.

Basically:
- I can make the secondary roles work recursively even without the postponer. But thats pretty much all I can do:
- I cannot implement more advanced resolution strategies without postponer. 
- I cannot deal with race conditions (without adding a bunch of new states and additional complexity (right into logic of usb_report_updater)).
- I cannot make the extra usb reports compatible with secondary roles (without adding more states and complexity).
- I cannot simplify the code because the hooks are just placed where they are placed and cannot be moved. 
- The old mechanism makes the control flow hard to optimize and eats up time near the top of the calltree where the performance impact is multiplied by number of keys. The proposed mechanism will take a few more instruction in the leaves of the calltree, but only once per update cycle and only in case that the secondary role is actually bound on the activated key. 

Regarding structure of the code -  I see that the odds of red and green lines are strongly against accepting the PR, because the call tree is deeper and the code is longer, but does it really seem more complex when you read the new version in a top-down manner without any diffs shown?

mondalaci:
What makes it difficult for me to accept the postponer is not only the added code, but the nature of the algorithm. In specific, `cycles_until_activation` gets incremented every now and then, gets decremented in every update cycle, and when it reaches zero then the events get processed, one at a time. It seems rather arbitrary, and I find it hard to wrap my head around it.

I'd much prefer a more straightforward and predictable algorithm that doesn't try to postpone processing into the future, but rather deals with it in the now based on the current and past state. You're welcome to add state whenever needed, and it's preferred to make it specific rather than general.

kareltucek:
> It seems rather arbitrary, and I find it hard to wrap my head around it.

It is corresponds to a three state automaton:
- INACTIVE (queue is empty, everything is executed in time)
- WAITING_FOR_INPUT (There is a secondary role which needs further input. In the simple case it is either its own release, or press of another key.)
- REPLAYING_EVENTS (There is a periodic timer - `cycles_per_activation` which executes one event from the queue every second cycle, which is necessary for events to properly complete - it allows a key to go through all four states of its lifecycle.)

(In the simple strategy the queue always contains at most one keystroke (two events), unless new event manages to come earlier than the old keystroke is finished - and in that case, this strategy inherently prevents race conditions.)

> one at a time

That's the whole point - doing things otherwise leads to long hours spent on hunting for race conditions.

> I'd much prefer a more straightforward and predictable algorithm 

That is the thing - I do not have anything which is more straightforward and more predictable at the same time.

kareltucek:
I will see what I can do with the old algorithm, but I am not very optimistic in terms of long term maintainability.

mondalaci:
Maybe your algorithm is the best one, and my understanding is limited. I'll try to dig into this more, but it requires deep thinking, and I'm quite overloaded nowadays. Your patience is much appreciated.

kareltucek:
On the other hand I am clearly biased and not seeing its drawbacks (especially w.r.t. readability of the code, which is obviously not as good as it should be). 

> quite overloaded nowadays

I see, btw congratulations on progress with modules.

greneholt:
@kareltucek do you have a build available of this branch for testing? I've tried the latest build of your version of the firmware, but it does not fix the issues I'm seeing with activating the mod layer via a secondary role while one secondary role is already active. I'd like to see if this branch does.

kareltucek:
> do you have a build available of this branch for testing?

Sure. [uhk-fw-8.8.1-origin_secondary_roles-v2.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/4382465/uhk-fw-8.8.1-origin_secondary_roles-v2.tar.gz)

Thanks for help with testing! If you encounter any inconsistencies, please do not hesitate to report them :-).

kareltucek:
@mondalaci I have tried to clear up things in the Refactor commit above. If anything is unclear, unreadable, I will gladly supplement more comments or refactors. I definitely think that code should be readable without headaches.

After a few days of thinking I still prefer this variant. Brief motivations:
- I feel uneasy about trying to maintain the current solution, because it is complex, has bad locality in the code, is hard to extend and gives no general guarantees on event interfecence.
- I do not feel strongly about writing a new minimalistic solution, because I do not have clear idea about how it should look like and how it should be different from the current one in order to actually improve the state of affairs and yet remain acceptably minimalistic for you.
- This PR is a bit of an overkill, but gives strong guarantees regarding interference and chronology of events. Also, it allows me to maintain, test and improve the same code for both branches, which in the end should be a benefit for you too, because this PR simply has a built-in developer who is going to keep using the implementation on daily basis.

greneholt:
Thank you @kareltucek! That build works perfectly, and fixes all the issues I've had with multiple secondary roles. You have saved me from many future undo operations when I press modifiers in the wrong order :)

mondalaci:
@kareltucek Sorry for delaying this PR so long! I've been unsure whether this is the best solution possible, and I'm still not sure to be honest, but I can't offer a better solution.

Most importantly, your solution works, you're much more experienced with this part of the firmware than I am, and you deeply care. Keep up the great work!

mondalaci:
@kareltucek Please close the issues which are resolved by this PR, and reference this PR in their comments.

kareltucek:
Happy to hear that!

Sure!

Issue #44, opened 2017-06-24
Fix BusPal


mondalaci:
Currently, BusPal partially works. For example, the `get-property` command works as expected:

```
$ blhost --usb 0x1d50,0x6121 --buspal i2c -d get-property 1
Inject pre-config command 'i2c'
[01 00 0c 00 c1 00 00 02 10 00 00 00 64 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
<03 00 0c 00 a0 00 00 02 00 00 00 00 c1 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>
Successful generic response to command 'i2c'
  - took 0.000 seconds
Response status = 0 (0x0) Success.
Inject command 'get-property'
[01 00 0c 00 07 00 00 02 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
<03 00 0c 00 a7 00 00 02 00 00 00 00 00 00 01 4b 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>
Successful response to command 'get-property(current-version)'
  - took 0.000 seconds
Response status = 0 (0x0) Success.
Response word 1 = 1258356736 (0x4b010000)
Current Version = K1.0.0
```

Unfortunately, the `flash-image` command aborts:

```
$ blhost --usb 0x1d50,0x6121 --buspal i2c -d flash-image uhk-left.srec 
Inject pre-config command 'i2c'
[01 00 0c 00 c1 00 00 02 10 00 00 00 64 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
<03 00 0c 00 a0 00 00 02 00 00 00 00 c1 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>
Successful generic response to command 'i2c'
  - took 0.000 seconds
Response status = 0 (0x0) Success.
Inject command 'flash-image'
Wrote 192 bytes to address 0
[01 00 0c 00 04 01 00 02 00 00 00 00 c0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
[...]

PR #277, opened 2020-01-30
Fix typo in buspal.

### LINKED ISSUE #276: Possible typo in buspal
https://github.com/UltimateHackingKeyboard/firmware/blob/14f1fd2ad478e5e05641570e69f436078b1e15a1/right/src/buspal/command.c#L179

@mondalaci would you take a look at this?

kareltucek:
Closes #276

### DIFF
--- a/right/src/buspal/command.c
+++ b/right/src/buspal/command.c
@@ -176,7 +176,7 @@ static void finalize_data_phase(status_t status)
           status = kStatus_Fail;
       }
     }
-    if (kStatus_Success!=kStatus_Success)
+    if (status!=kStatus_Success)
 #else
         // receiving framing packet header
         if (peripheral_read(rxBuf, sizeof(framing_data_packet_t)) != kStatus_Success)


mondalaci:
The firmware update is unchanged based on my tests, but what makes you think that this change doesn't affect BusPal negatively?

kareltucek:
Not sure if you want to actually know.

In the first place, it is quite clear that the author most likely didn't mean `if (kStatus_Success!=kStatus_Success)`. If he did, he would have written it in more direct fashion. Like `if (true)`, or something along those lines. He most likely meant `status!=kStatus_Success`.

Now, if you look at the preprocessor directive, you can notice that the first branch is just a different version of the second branch. Second branch is original, the first is "fixed". The first branch therefore follows API signature of the second branch.

So the logic should quite clearly be:

        if (peripheral_read(rxBuf, sizeof(framing_data_packet_t)) != kStatus_Success)
        {
            status = kStatus_Fail;
        }

The "fixed" branch is just a more chatty way of saying the ` if (peripheral_read(rxBuf, sizeof(framing_data_packet_t)) != kStatus_Success) `

Now this is actually written at multiple places - 182+185, then at 174+176. It is the line 174+176 which makes me quite confident that the author meant the same by 179+185. This is consistent with the original code. Also, I simply don't believe that the author meant "If reading bus failed, then skip failed branch and perform success branch instead, but at the same time leave failed status". 

Second reason is that it looks much safer to me to simply set "failed" status and leave in case of failure (even at the risk is that it does affect BusPal negatively) than to leave there the variant which carries out "success" branch of computation in "failed" case.

Third reason is actually reading what it does. The only change in the flow is the line 190 (whether or not it is performed in the failed scenario). Question whether or not it has some side effects is surely relevant, but it does not seem to make sense to be obliged to "consume" data from the input stream if the communication has simply failed. Under more scrutiny there are two arguments:
- `peripheral_read` at 42 does not take care of returning data into the buffer if it did not find what it was searching for. If the side effect was important, it would have to do its job with much more care. 
- amount of data read at 190 depends on length extracted from the framing packet. But if read of the frame packet failed, then the value is junk. 

Disclaimer and a funny fact: I still have no idea what the code does. If it bothers you, feel totally free to close this.

kareltucek:
Either way, I am very sorry for wasting your time with this entire buspal problem. It turns out to be a no-problem in either case...

kareltucek:
And one more detail - I looked into git history and found https://github.com/UltimateHackingKeyboard/firmware/issues/44 .

mondalaci:
Thanks for the thorough explanation, and I deeply appreciate that you care about these issues so much!

Issue #261, opened 2020-01-10
8.7.1 periodically ceases to send keystrokes


josephvusich:
Firmware: 8.7.1
Agent: 1.3.0

First incident happened within an hour after upgrading two UHKs from 8.5.3 to 8.7.1. Both keyboards had previously been running 8.5.3 for over a year with no issue.

### Description of event
The UHK abruptly stops sending keystrokes.

Keyboard seems to be responsive (layer switching, keymap switching, etc. still work). Unplugging and replugging the USB cable resolves the issue immediately.

### Occurrences
Total of 4 incidents on two different UHKs, over a period of 48 hours. In all cases it happened while typing, not while sitting idle.

UHK 1:
* Upgraded via Agent 1.3.0 on Windows 10 desktop with November 2019 Update.
* Occurred twice, both times connected to the same Windows 10 desktop, once with Agent running, once without.

UHK 2:
* Upgraded via Agent 1.3.0 on Mac OS X 10.15.2.
* Occurred twice, once while connected to Mac OS X 10.15.2, and once while connected to a laptop running Windows 10 1709. Agent was not running in either case.

josephvusich:
I have downgraded to 8.7.0 for now, to determine if the issue persists.

mondalaci:
Thanks so much for reporting this! I've experienced similar issues with 8.7.1 lately. I thought that my USB hub is the culprit, but given your report, chances are it's actually the firmware. Just followed suit and downgraded to 8.7.0. Let's keep each other updated!

kareltucek:
This might be unrelated, but I have been experiencing similar issue for quite some time on my fork. 

Approximately once every week, some key becomes "stuck" (in pressed state) and UHK does not react to subsequent presses. Not sure if all the details are the same (layer switching LEDs), will check next time it happens.

So far, I have been blaming my poor soldering skills and related possibility of either having damaged something by heat or having shortcirquited something by rosin residue.

pcooke9:
> I thought that my USB hub is the culprit, but given your report, chances are it's actually the firmware. Just followed suit and downgraded to 8.7.0. Let's keep each other updated!

[...]

Issue #172, opened 2018-09-13
Freeze bug


mondalaci:
Firmware versions larger than 8.2.5 sometimes freeze. When it happens, your UHK gets unresponsive all of a sudden. Your UHK may not be affected at all, or it can take hours, days or even weeks for the freeze to hit.

The reason we haven't yet fixed this bug is because it happens so rarely, and only on certain UHKs. We need your feedback, so if you're reading this please update to [firmware 8.4.5](https://github.com/UltimateHackingKeyboard/firmware/releases/tag/v8.4.5) via Agent.

If your UHK freezes, run the following script and paste its output to provide diagnostics information, so that we can hopefully figure out what the hell is going on. (You'll need another keyboard to do this if your UHK froze).

Pull the [Agent repo](https://github.com/UltimateHackingKeyboard/agent) and build it. Then run `packages/usb/get-debug-info.js`. It'll output something like this:

```
I2cWatchdog:919988055 | I2cSlave:306651339 | I2cWatch:950237 | I2cRecovery:0 | KeyMatrix:783778471 | UsbReport:157778114 | Time:101366130 | UsbGeneric:237 | UsbBasic:41632 | UsbMedia:252 | UsbSystem:4 | UsbMouse:110380
I2cWatchdog:919997815 | I2cSlave:306654592 | I2cWatch:950247 | I2cRecovery:0 | KeyMatrix:783783350 | UsbReport:157782993 | Time:101367237 | UsbGeneric:238 | UsbBasic:41632 | UsbMedia:252 | UsbSystem:4 | UsbMouse:110380
I2cWatchdog:920007573 | I2cSlave:306657846 | I2cWatch:950257 | I2cRecovery:0 | KeyMatrix:783788300 | UsbReport:157787816 | Time:101368361 | UsbGeneric:239 | UsbBasic:41640 | UsbMedia:252 | UsbSystem:4 | UsbMouse:110380
I2cWatchdog:920017331 | I2cSlave:306661098 | I2cWatch:950267 | I2cRecovery:0 | KeyMatrix:783793238 | UsbReport:157792640 | Time:101369482 | UsbGeneric:240 | UsbBasic:41644 | UsbMedia:252 | UsbSystem:4 | UsbMouse:110380
[...]

PR #278, opened 2020-02-02
Prevent deadlocks and races in usb semaphores.

kareltucek:
See code comments.

Relates to #172 and #261.

### DIFF
--- a/right/src/ksdk_usb/usb_device_hid.c
+++ b/right/src/ksdk_usb/usb_device_hid.c
@@ -656,10 +656,12 @@ usb_status_t USB_DeviceHidRecv(class_handle_t handle, uint8_t ep, uint8_t *buffe
     {
         return kStatus_USB_Busy;
     }
+    hidHandle->interruptOutPipeBusy = 1U;
     error = USB_DeviceRecvRequest(hidHandle->handle, ep, buffer, length);
-    if (kStatus_USB_Success == error)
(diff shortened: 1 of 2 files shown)


right/src/ksdk_usb/usb_device_hid.c
```diff
@@ -656,10 +658,12 @@ usb_status_t USB_DeviceHidRecv(class_handle_t handle, uint8_t ep, uint8_t *buffe
     {
         return kStatus_USB_Busy;
     }
+    hidHandle->interruptOutPipeBusy = 1U;
```
kareltucek, line 663: This was a typo, I meant to do this on the In pipe rather than Out pipe, but the same problem applies here.

mondalaci:
Very interested whether this will fix the the issue. I assume that uhk-fw-8.7.1-fix_candidate_2.tar.gz contains this PR, and we'll be testing it for a while before I can safely merge it.

kareltucek:
> Very interested whether this will fix the the issue.

Me too...

> I assume that uhk-fw-8.7.1-fix_candidate_2.tar.gz contains this PR

Correct.

> , and we'll be testing it for a while before I can safely merge it.

Testing it for a while is a reasonable course of action, but I think this should be considered mostly independently of the two freeze problems, and based on logic argumentation rather than just on testing. The variables (the InterruptPipeBusy ones) are read and written at very few places, which makes logic argumentation possible even without detailed knowledge of the rest of the usb stack. (And leads me to a belief that the PR should be safe.)

PR #284, opened 2020-03-28
Fix memory corruption in led_display driver

kareltucek:
What happened here (in case it is not obvious):
`LedDisplayIcon_Last` allocated array of 2 states instead of 3. As a result, writing adaptive mode led resulted into rewriting someone else's piece of memory - whatever the linker placed right after the `ledIconStates` array. In this case it blew up macro recorder on my fork. 

Please, lets be careful about memory corruptions (allocations, indices, etc.). These bugs can be pretty nasty, and today I was very lucky to be able to hunt it down so easily.


mondalaci:
Great catch! I'll do my best to not make such bugs in the future.

PR #294, opened 2020-04-24
Layer handling refactor

### LINKED ISSUE #167: Don't suppress keys upon keymap change
Let's say I map fn+\ to toggle between several keymaps. I used to be able to hold fn and tap \ several times and the display updated, but starting from 8.4.5 i have to release fn every time.

Please restore the old behavior regarding keymap switching but keep the new key suppression behavior regarding layer switching.

### LINKED ISSUE #195: Toggle mouse takes precedence over Mod + Mouse, making this combination impossible for CAPS LOCK if toggle is used
I've tried to use Mod + Mouse to switch on CAPS lock. However, after changing Mouse to Toggle mouse this becomes impossible as the toggle seems to take precedence over the Mod + Mouse combination. Possibly combinations could take precedence over any toggle behavior. I'm filing this separate from other ideas with regards to CAPS lock as I think this could be seen as a bug rather than a feature request.

### LINKED ISSUE #199: Layer deactivation
Let's assume following mapping:

- left case button is mapped to "activate mod"
- right case button is mapped to "toggle mod"

Scenario:
- Single tap right button
- Single tap left button

Expected outcome:
- First tap toggles the layer
- Second tap untoggles the layer

Actual outcome:
- The first tap toggles the layer
- Second tap does nothing

Scenario no. 2:
- Doubletap left button
- Tap right button
- Tap right button again

Expected outcome:
- layer toggles
- layer untoggles
- layer toggles again

Actual outcome:
- layer toggles
- layer untoggles
- nothing

What is the reason for checking for `isLayerDoubleTapToggled` on line 214 in usb_report_updater.c? I see no good reason for that and I assume that it is the culprit.

    if (!keyState->previous && isLayerDoubleTapToggled && ToggledLayer == action->switchLayer.layer) {
        ToggledLayer = LayerId_Base;
        isLayerDoubleTapToggled = false;
    }

### LINKED ISSUE #222: Misbehaving layer toggling via activated layer
1. On the Base layer, let's bind activate mouse layer to key 1 and toggle mod layer to key 2.
2. Hold key 1.
3. Press key 2.
4. At this point, instead of triggering the action of key 2 on the mouse layer, the base layer action gets triggered, hence the mod layer gets toggled.

See https://github.com/UltimateHackingKeyboard/agent/issues/949

### LINKED ISSUE #291: Toggle layer stopped working on 8.9.0
It seems like "Toggle" layer stopped working. (However, "double-tap" lock still works fine.)

Found out accidentally while verifying Piratk's report.

kareltucek:
I have found some time to finally refactor layer switching. Some code has just moved to different places, some code was rewritten but the idea remained the same. But generally I chose to do such changes which would allow me to reason about correctness of the solution, which mostly means that I have deleted and refactored a lot.

The main ideas:
- This time I am not introducing any new abstract constructs, and am keeping it in a minimal form. 
- I have moved the end logic of layer switching into a new file `layer_switcher.c`, so it is no longer scattered across many different places. Decision making logic is generally branched in `usb_report_updater.c` according to action structure, but all calls end up in `layer_switcher.c`. This should allow better overview of the logic. Of course, this again means deeper call tree and a bit more code, but (I believe) better locality of the code.
- I have tried to write down some actual "specification" of the logic so that there is something to refer to when reasoning about issues and feature requests. -> `layer_switcher.c`
- I have strictly separated "hold" effects from "toggle" effects. "Hold" effects are now always decided based on base-layer action, irrespectively of actual active layer. On the other hand, "Toggle" effects are always treated as regular actions (they *always* respect active layer and code branching structure).
  - This is unfortunately not very nice, since double-tap-toggle actions have effects of both types, and therefore in these case the flow goes independently by two different paths.
  - This allowed me to get rid of the antipattern of "toggle" action which (for instance) went through layer hold mechanism and even intercepted non-base-layer actions. 
  - Holds are now evaluated as part of the standard action branching and evaluation. This should simplify the code a bit and also speed up the update loop considerably.
- I even managed to make secondary roles comply with behaviour of regular hold switches. However, in this case, I am not very sure if my sense for detail haven't gotten ahead of myself. At the moment, I leave the regular implementation in place via `SECONDARY_AS_REGULAR_HOLD` directive.
- I have done some small refactors, like moving the UsbStackTest into a separate procedure.

Although I have considered all design decisions carefully and am therefore quite sure about correctness and sanity of the *design*, the thing should definitely be tested for some time before accepted into upstream (esp. due to possible presence of typos and other dumb mistakes). However, anyone interested is welcome to give it a try and report back:

[uhk-fw-8.9.0-layer-switching-refactor.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/4531193/uhk-fw-8.9.0-layer-switching-refactor.tar.gz)

Fixes #291 properly.
Closes #222 .
Closes #199 .
Closes #195 .
Closes #167 . (Even including secondary roles.)

(The ratio of added code is not as bad as it looks like if one considers amount of added comments, and layer.c whose deletion github doesn't render by default.)


mondalaci:
Was #291 not fixed properly?

kareltucek:
It was a quick hotfix which simply restored pre-8.9.0 behaviour. It works as far as we are content with the way it worked before.

But in principle, the fix it is making the environment around the bug behave the same way as before, rather than fixing the nature of the problem, which in my opinion lays in design.

mondalaci:
Very impressed by your work and appreciate your thorough explanation. Keep it up!

PR #301, opened 2020-06-21
Fix for USB3.x issues

### LINKED ISSUE #269: USB 3.x compatibility problem
**Firmware**: 8.7.0, but noticed this on earlier versions as well.

# Context and the problem
I've been using the UHK with Gigabyte Z370P and Intel i5-8400 without any issues. It's also working perfectly fine with ThinkPad W520 and X1C6. Recently I switched my desktop to Ryzen 7 3800x and the keyboard still worked fine via USB hub built in my monitor. However, when I tried connecting it directly to the motherboard (to update the firmware), the keyboard didn't work (the LED screen was working, but the keystrokes were not registered via BIOS). I replugged it, logged into Linux and checked `dmesg` to see "Not enough bandwidth for new device state." after plugging it back directly to motherboard.

# Testing
Since then I've tried the following boards:
- ASUS TUF GAMING X570-PLUS - **not working** (I'm currently using this one) - the manufacturer claims it's working fine when I sent them the board for the warranty claim; I also tested it with "Xbox 360 Wireless Receiver for Windows" (which is also a `full-speed` device) and it seems to be having the same problem,
- Gigabyte X570 AORUS ULTRA - **works fine** - unfortunately cooling settings are incompatible with my chassis' fan hub, so keeping this board was not an option for me, because it was pretty noisy and the CPU temperatures were high on idle,
- MSI MEG X570 ACE - **not working** - UHK worked with the front panel of my chassis (which didn't work for ASUS), but I still cannot update the firmware via it, as it claims to be a USB hub; the board supports "Xbox 360 Wireless Receiver for Windows", though,
- ASRock X570 Phantom Gaming X - **not working** - the same as "MSI MEG X570 ACE".

# Conclusions
I'm suspecting that not all manufacturers care enough about proper implementation of `xhci_hcd` support for `full-speed` devices, but I'm definitely not an expert in this matter, so I'm just judging by my observations and some research. It also doesn't seem to be OS-related, as the keyboard doesn't work in BIOS (nor in Windows) as well.

# Logs
My logs from `dmesg` (along with some comments) are here:
https://privatebin.net/?7d14d583cccf58e4#GKFpvLrDL7wz29dZmW8VFwBbo8MFDAwTNnj3upuhNNmL

I included 4 scenarios there:
- plugging UHK to the rear USB (of the motherboard),
- plugging UHK to the USB hub (in the monitor),
- plugging UHK to the front USB panel (USB 3.0),
- booting with UHK plugged to the front USB panel (USB 2.0).

xxxajk:
Fixes #269.

This eliminates the bandwidth issue. Linux no longer complains.
This should fix USB3.x on Mac too.

I'll have to see if the following are happier:

1. Ancient Mac that I have
2. BIOS on the ASUS machine


kareltucek:
Could you elaborate a bit more on why does this fix the issue please? (Just interested.) 

Also, is there any reason why you did not lower the `USB_BASIC_KEYBOARD_INTERRUPT_IN_INTERVAL` when lowering all other four intervals?

xxxajk:
*SOME* USB implementations have bugs and strange quirks, where they get all stupid and report the wrong bandwidth, or only have a single transaction translator. Since the TT is busy, it can't handle the low speed or full speed device... Well it CAN, but it's fooled it's self, and thinks you will be eating 750% of the bandwidth when all are set to 4. Fun.
By setting it to 1, it says to the controller that you are only going to use one FS micro-frame. it's happy with 125uS. :-)


As far as the other one, You are correct, I missed that one. 
Adding it. Thanks for the sharp eye!

kareltucek:
Thanks for clarification!

xxxajk:
@kareltucek  Yeah, no problem.
Now someone needs to teach Travis how to do `/bin/sh which` or at least fake it.
Who takes care of that?
I like to keep my (several dozen, not kidding) ARM tool chains separate. 
On a normal system, which fixes that for those who don't care and have it either on $PATH or global workspace or project setting. It also allows me to just use the command line, and pretty much ditch eclipse (ok, KDS, whatever) for any IDE I want.

kareltucek:
> Who takes care of that?

@mondalaci ?

Agrendalath:
> BIOS on the ASUS machine

I just tested this within ASUS BIOS and inside Windows and it works like a charm.

xxxajk:
Great! Happy that it's not a problem any more.

PR #318, opened 2020-11-19
Simplify building firmware

dancek:
- add make-release.js option --skip-agent to avoid building the agent
- add instructions for the simplest way of building the firmware

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,18 @@ If you want to use the latest firmware version for your UHK, then instead of goi
 
 If you're one of the brave few who wants to hack the firmware then read on.
 
+## Building
+
+1. Make sure to clone this repo with:
+
+`git clone --recursive git@github.com:UltimateHackingKeyboard/firmware.git`
+
+2. Install `arm-none-eabi-binutils`, `arm-none-eabi-gcc`, `arm-none-eabi-newlib` and Node.js v12.
+
+3. `cd scripts && npm install && ./make-release.js --skip-agent`
+
+## Full development setup
+
 1. Make sure to clone this repo with:
 
 `git clone --recursive git@github.com:UltimateHackingKeyboard/firmware.git`
--- a/scripts/make-release.js
+++ b/scripts/make-release.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 const fs = require('fs');
 const path = require('path');
 require('shelljs/global');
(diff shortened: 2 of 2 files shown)


mondalaci:
Thank you for your PR, but I don't think it fits our use case. The release script is meant to build a fully-fledged release tarball for which Agent is necessary to construct the binary representation of the user configuration. Unless you have a valid objection, I'd rather close this PR.

dancek:
Hi @mondalaci! I'm not completely sure what you mean. I'm currently running a firmware built with this script on my UHK and it works for me.

I wanted to modify the firmware (I'll send another PR your way later) but didn't want to install anything unnecessary. If this approach doesn't fully work, I'll suggest removing the `--skip-agent` part but editing the README anyway. It took me a while to figure out how to build so I think the instructions are valuable for other contributors in the future.

EDIT: just to clarify: I built a tarball with `./make-release.js --skip-agent`, then flashed it with UHK Agent installed from the official AppImage.

dancek:
Since we're talking, I'm planning on adding the HID Usage Page for controlling the led display, Auxiliary Display Page (0x14). I already have a proof of concept for changing the text with a Set_Report request and some libusb code for controlling it from userspace, but there's a lot to be done before it's a good implementation.

I'm thinking of adding the usages for text and 14-segment display as specified in [HID Usage Tables, p. 191-203](https://www.usb.org/sites/default/files/hut1_21.pdf) to `UsbBasicKeyboardReportDescriptor`. If you have any comments about that, I'd be very happy to hear them.

Of course I realize you might not want the change. In that case I'll just have a fork with the feature.

mondalaci:
Thank you for elaborating on your use case and goal.

It's unnecessary to include the user configuration into the tarball for flashing it via Agent, but it should be necessary to reset the user configuration via Agent. Hence, firmware releases that don't contain the user configuration are incomplete.

As for building the firmware as easily as possible, you can simply execute `make` in the top-level directories of this repo where Makefile is present. `make flash` flashes the individual firmware projects to the UHK, and it depends on Agent. I consider building the tarball and uploading it via Agent in each development iteration counter-productive. Feel free to revise the readme accordingly, and I'll take a look.

Regarding making the LED display controllable via USB, a new USB command should be added to the existing ones instead of modifying the descriptors. See `right/src/usb_commands`. We can discuss this further if you want to, but perhaps a separate issue would be better suited.

dancek:
Ok! Thanks for the pointers. I'll consider another stab at the README once I better understand what I'm doing :)

mondalaci:
Cool! I can write a specification for controlling UHK LEDs via USB in a new issue. Let me know if you're serious about this, and I'll do it.

PR #320, opened 2020-11-20
WIP / RFC: commands for setting led segments

dancek:
So, here's some code. Thanks for the hint about using the same command system as for everything else. I was planning to fully model the display as a HID feature with usage pages, reports and all, but that's probably overkill. The only slight downside of this much simpler approach is that it's vendor specific. But there probably are no generic HID segment display drivers anyway.

For this firmware patch there's proof-of-concept userspace code for writing to the display at https://github.com/dancek/uhk-leds/blob/master/src/main.rs .

I arbitrarily chose the command ids 0x21 and 0x22 but expect them to change. Please recommend whatever suits your plans.

Any code review is welcome, as are comments regarding what should be done so this can be included in the official firmware (as opposed to a fork).

### DIFF
--- a/right/src/led_display.c
+++ b/right/src/led_display.c
@@ -116,16 +116,22 @@ static const uint8_t segmentLedIds[maxSegmentChars][ledCountPerChar] = {
     {105, 121, 124, 136, 137, 106, 122, 123, 120, 138, 139, 140, 107, 108},
 };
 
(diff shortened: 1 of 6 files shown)


mondalaci:
Your PR looks good, but I'd much rather prefer getting #322 implemented.

dancek:
Cool, let's work on that!

PR #321, opened 2020-11-20
Documentation for a CLI-based dev setup

dancek:
This is another try at better documenting how this firmware can be developed without installing the full IDE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,17 @@ This repository hosts the firmware of the [Ultimate Hacking Keyboard](https://ul
 
 If you want to use the latest firmware version for your UHK, then instead of going through the pain of building the firmware, simply download the [latest release of Agent](https://github.com/UltimateHackingKeyboard/agent/releases/latest) and update to the latest firmware version within Agent with a click of a button.
 
+## Developing
+
 If you're one of the brave few who wants to hack the firmware then read on.
 
 1. Make sure to clone this repo with:
 
 `git clone --recursive git@github.com:UltimateHackingKeyboard/firmware.git`
+
+Then, depending whether you want a full IDE experience or just minimal tools for building and flashing firmware, read *IDE setup* or *Minimal development setup* (if you prefer a text editor + command line).
+
+### IDE setup
 
 2. Download and install MCUXpresso IDE for [Linux](https://storage.googleapis.com/ugl-static/mcuxpresso-ide/mcuxpressoide-11.2.0_4120.x86_64.deb.bin), [Mac](https://storage.googleapis.com/ugl-static/mcuxpresso-ide/MCUXpressoIDE_11.2.0_4120.pkg), or [Windows](https://storage.googleapis.com/ugl-static/mcuxpresso-ide/MCUXpressoIDE_11.2.0_4120.exe).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,6 +33,20 @@ If you're one of the brave few who wants to hack the firmware then read on.
 
 Going forward, it's easier to flash the firmware of your choice by using the downwards toolbar icon which is located rightwards of the *green play + toolbox icon*.
 
+### Minimal development setup
+
+1. Install the ARM cross-compiler, cross-assembler and stdlib implementation. Eg. on Arch Linux the packages `arm-none-eabi-binutils`, `arm-none-eabi-gcc`, `arm-none-eabi-newlib`.
+
+2. Install Node.js v12. If you have a later version, editing the version requirement in `lib/agent/package.json` *might* work.
+
+3. Build UHK Agent. `cd lib/agent && git pull --autostash && npm ci && npm run build`.
```
mondalaci: `git clone --recursive` should pull Agent.

README.md
```diff
@@ -27,6 +33,20 @@ If you're one of the brave few who wants to hack the firmware then read on.
...
+
+4. Still inside the Agent submodule, compile flashing util scripts. `cd packages/usb && npx tsc`.
```
mondalaci, line 44: `npm run build` should take care of this.

README.md
dancek: Yes, the Agent repo should already be cloned but it might not be the latest commit. This was added because `make-release.js` also pulls the current master and I thought it might be useful.

README.md
dancek: It didn't for me. Maybe some editor plugin that you're using automatically transpiles typescript to javascript?

mondalaci: @ert78gb Can you comment on this?

README.md
mondalaci: Now that I think about it, it's probably not a good idea to pull the master because it might be broken. We should stick to the Agent revision that the firmware repo points to, and change it explicitly if needed. Would you please modify the script accordingly?

dancek: I did. Note that it's now possible to accidentally *not* update the submodule before building a release if submodules are not part of the typical workflow. I'm not sure what the best practices with submodules are.

README.md
ert78gb: the `npm run build` compile the bundles from the ts files. It doesn't compile the ts files in `lib/agent/packages/usb`, but he step 4 do it.

@dancek if you experienced the bundles not compiled for you please check the output of the build command. If you don't find the error inside then send us the output for me and I will help.

The ts files in the `lib/agent/packages/usb` starts with the `#!/usr/bin/env ../../node_modules/.bin/ts-node-script` shebang so it will compile at runtime. If someone use it frequently and use low capacity cpu-s then worth to compile them with the step 4.

README.md
```diff
@@ -27,6 +33,20 @@ If you're one of the brave few who wants to hack the firmware then read on.
 
 Going forward, it's easier to flash the firmware of your choice by using the downwards toolbar icon which is located rightwards of the *green play + toolbox icon*.
 
+### Minimal development setup
+
+1. Install the ARM cross-compiler, cross-assembler and stdlib implementation. Eg. on Arch Linux the packages `arm-none-eabi-binutils`, `arm-none-eabi-gcc`, `arm-none-eabi-newlib`.
+
+2. Install Node.js v12. If you have a later version, editing the version requirement in `lib/agent/package.json` *might* work.
```
ert78gb, line 40: It is a very interest point. The agent built on the electron and every electron has it own node.js runtime. We always use the same version for development that provided by the electron.
The repo contains an `.nvmrc` file that contains the recommended node.js version for development. It is a de facto standard in the node ecosystem.
To manage multiple node.js on 1 machine recommended to use the [nvm](https://github.com/nvm-sh/nvm) or windows equivalent.

There are many technical requirement of an electron major version upgrade. For example we have to wait to all of our 3rd party dependencies support it.
So I don't recommend to use other version for development that we use.
This is the reason, why we implemented the node.js version checking as a prerequisite of the build.

README.md
ert78gb: Agent can be develop without any IDE.

dancek: The flashing scripts in this repo use the compiled js files. Eg. https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/Makefile#L32

```
FLASH_CMD = node ../lib/agent/packages/usb/update-device-firmware.js $(PROJECT_OBJ:.axf=.hex)
```

This is why I added step 4. I do realize there are other ways, but I wanted to document how this works now rather than change it.

README.md
dancek: Yeah, I expected as much. For just running the flashing scripts the hacky approach seems sufficient, but I wouldn't do that to actually build the Agent for use.

README.md
mondalaci: Recently, @ert78gb converted the js files to ts, so the latter should be used, and Makefiles should be updated accordingly.

PR #323, opened 2020-12-26
WIP: Exponential speed driver for trackball/touchpad

kareltucek:
Here is some experimental code, which should make the touchpad and trackball modules behave even nicer.

Constants may need some tweaking  (esp. for touchpad; these are for the trackball; touchpad will need higher minimum speed coeficient and lower mid speed)...

Feel free to discard it :-).


kareltucek:
In order to function properly, the code will need proper float handling. (This is prerequisite for any speed scaling, so it is not directly related to this proposal.)  

It already exists in the keyboard mouse handler, therefore just needs to be refactored so that it is applicable to all module types.

PR #325, opened 2021-01-10
Experimental mouse driver code

kareltucek:


mondalaci:
@kareltucek Why haven't you used the exponential driver for the trackpoint?

right/src/mouse_controller.c
```diff
@@ -336,13 +336,13 @@ void MouseController_ProcessMouseActions()
     if(moveDeltaChanged) {
         float xSumInt;
         float ySumInt;
-        if(ActiveMouseStates[SerializedMouseAction_ScrollMode]) {
```
kareltucek: @mondalaci Are you intentionally trying to make my life more painful, or are you just not realizing that by removing abstractions from code you are diverging your code away from mine (as well as anyone else's who might be (alghough probably is not) trying to make a configurable fork of the code)?

Also, do you realize that by hardcoding scrolling to mouse layer, you are effectively discarding' people's mouse layer binding for mouse buttons? Personally, while using UHK, my brain is either in mouse mode or in non-mouse mode, and expects bindings of mouse and non-mouse be interchangable.

In plain english this means that I tend to keep my mouse modifer pressed even if using modules to move cursor, but at the same time, I am used to press mouse buttons using my mouse-layer bindings. Hardcoding this as you have just now did makes my mouse layer cursor movement not work with module mouse buttons (because those are currently hardcoded exactly to the base layer), as well as (the other way around) it makes the module cursor movement not work with my mouse layer bindings.

The only reason this setup worked fine for me on the experimental branch is that my keymaps have the mod and mouse layers swapped. (And ofc since one usually wants to combine cursor movement with clicking, but does not care for combiging scrolling movement with clicking.)

right/src/mouse_controller.c
```diff
@@ -245,7 +245,7 @@ static void recalculateCurrentSpeed(float x, float y) {
         static uint32_t lastUpdate = 0;
         uint32_t elapsedTime = CurrentTime - lastUpdate;
         float distance = sqrt(x*x + y*y);
-        currentSpeed = distance * 1000.0f / elapsedTime;
```
kareltucek: There is basically nothing wrong with doing this, except that it changes interpretation of the values and deprecates the comments. 

I.e., for sake of other people's reading abilities, it might be a good idea to update the comments on lines 240 and 257 to read `px/ms` instead of `px/s`. (Of course I do not care much. I am not especially bad at figuring out meaning of my own code.)

right/src/mouse_controller.c
mondalaci: @kareltucek According to [this blog post](https://ultimatehackingkeyboard.com/blog/2021/02/01/module-testing-result-and-manufacturing-progress), I want to make speed, acceleration, and navigation mode adjustable on a per-layer and per-module basis. I want to make a dedicated configuration data structure but want to start with sane defaults. No worries, this will be configurable.

kareltucek: On the second thought, I can see that you are having trouble to coactivate layers with mouse modes, since obviously your firmware is not prepared to do such a piece of arcane magic, so I give you that you might be having good/"good" reason to solve the problem so easily at the expense of abstractness, etc..

However, the other argument stands still :-/.

kareltucek: You know what? I guess I actually don't care. 

If you want to hear my opinion, then my solution candidate would be the following, but I guess I do not have the energy to pour into it anymore anyway:

Have one action slot of the module's layer reserved for expressing module mode functionality. I.e., have this behaviour hardcoded in layers of keymap.c in the default keymap, and as a configurable part of the configured keymaps. And then - let's activate the slot action whenever the module is active. 

In order to do this, you would have to have the scroll mode (and other modes) exist as a regular action (doesn't matter whether as a part of SerializedMouseAction or some other action/enum/whatever). Then making this thing configurable via Agent is a piece of cake even without any changes to serialization format. 

This idea can easily be extended to your idea of media mode bindings...

kareltucek: > @kareltucek According to [this blog post](https://ultimatehackingkeyboard.com/blog/2021/02/01/module-testing-result-and-manufacturing-progress), I want to make speed, acceleration, and navigation mode adjustable on a per-layer and per-module basis. I want to make a dedicated configuration data structure but want to start with sane defaults. No worries, this will be configurable.

Sorry! Didn't see this until now... ...the problem with your idea of things is that 1. they usually don't compose well with other functionality. 2. (and this is especially cruel from me to say... ...but true) they take decades to come true :-(.

kareltucek:
> @kareltucek Why haven't you used the exponential driver for the trackpoint?

Because it does not make sense and worsens usability. 

The reason is that touchpad and trackball control directly the position of the cursor, while trackpoint controls the speed of cursor - i.e., the derivative of position (i.e., the speed of change of position). 

Thereforel, by applying the exp driver to the trackball/touchpad, we let the user to somehow affect the derivative of position (i.e., speed) on top of controlling the position directly. On the other hand, trackpoint **already** controls that speed directly, and therefore the only thing to control there is the speed coefficient constant.

Expressed mathematically, applying the exp driver to the touchpad/trackball output is same as applying the derivative of the exp driver to trackpoint output. And  the constant speed coefficient of trackpoint is exactly the derivative of the exp function (for whom the 'exp' name no longer makes any sense, since now it is actually a simple polynomial, not an exponential). 

|        | Touchpad/Trackball .................................. | Trackpoint ..................................
| :---------- | :----------------: | :----------------:  |
| User directly controls | position  <br />`p` | speed of cursor <br />i.e., derivative of position  <br />`d(p)`  | 
| Firmware sees (i.e., our input `inp` equals) | change of position<br />`d(p)` | change of speed <br /> `d(d(p))` | 
| Output `d_out` without any driver (by definition) equals [^1] | `d_out(p) := d(p)` | `d_out(p) := d(d(p))` |
| We want to apply the driver to `d(p)`, therefore we apply it as | `d_out(p) :=` <br />`= d(p)*drv(p)` | `d_out(p) := ` <br /> `= d(d(p)*drv(p))  ` <br /> `= d(d(p))*d(drv(p))` |
| Since  `drv(p) = a*p+b` | `d_out(p) := ` <br /> `= d(p)*(a*p+b) ` | `d_out(p) := ` <br />`= d(d(p))*d(a*p+b)` |
| Substituted with `inp` again [^2] | `d_out(p) :=` <br /> `= inp*(a*d'(inp)+b)` | `d_out(p) :=` <br />`= inp*a` |

</div>

[^1]: note that basically `d_out` (as output delta) and `d` both denote the same thing - a derivative or a difference, just since we are applying some transformations, writing equality between them would be... ...awkward.
[^2]: for `d'` being inversion of derivative.

right/src/mouse_controller.c
mondalaci: 1. I've just committed bc9d28fba4341cf804b48732a1be123dcdd544e4. Do you think that the configuration data structure introduced wouldn't provide you sufficient freedom? If so, please elaborate on your use case.
2. Do you mean exposing related configuration options via Agent? I plan to merge your macro engine and make this configurable via macros, so no additional Agent features would be needed for the time being.

kareltucek: >     1. I've just committed [bc9d28f](https://github.com/UltimateHackingKeyboard/firmware/commit/bc9d28fba4341cf804b48732a1be123dcdd544e4). Do you think that the configuration data structure introduced wouldn't provide you sufficient freedom? If so, please elaborate on your use case.

The trouble is that it is not independent on current layer. 

From abstract point of view, there should exist a simple expression which will be able to answer question "Which navigational mode are we in now?" (Or equally "Are we now in scrolling operational mode?".)

The usecase is to have modifiers which are able to switch navigational mode independently of current layer. 

Motivation behind this:
- I do not know why, but I do not perceive navigational mode as something related to the concept of layers.
- Concrete usecase - I like to be able to navigate by one hand only. Let's say that I am in the middle of a breakfast, or that I am holding a musical instrument in one hand while turning (pdf) pages of sheet music by the other hand. This means that I usually toggle mouse layer, and operate with it one handed. Therefore I would really like to be able to move cursor and scroll at the same time from within the same layer. The submotivations for using the mouse layer even when using mouse modules are:
  - When controlling the ball with my thumb, all my layer switches, as well as module mouse buttons are effectively unreachable, so I prefer to use the freedom of bindings on the mouse layer.
  - Space-wise, allocating anything on the base layer is very expensive.

>     2. Do you mean exposing related configuration options via Agent? 

Well, of course I do. Is not easy yet powerful configurability your main selling point? I mean, why not have a cake and eat it too, when it does not actually have to cost you the second cake?

> I plan to merge your macro engine and make this configurable via macros, so no additional Agent features would be needed for the time being.

I know, but this is still a very problematic thing, because:
- I expect it still is years down the road.
- I am not sure that you fully understands the implications. The truth is that the macro engine has accumulated some historic baggage (some bad design decisions, some insufficiently-self-confident design decisions, etc), and even though it works quite fine, quality-wise it nears its end of life. At the same time, even despite our agreements, I do no longer possess the vitality / energy to actively continue its development. The engine now ideally needs one thing - a complete rewrite worth around 1 man month (maybe a bit less) of my time, which however I am not capable of delivering due to the current lack of energy (not even against a king's ransom of recompense).
- I sort-of don't understand what's actually in it for you, or rather what you actually want to achieve with UHK (both personally and technically), because the macro language is pretty user-unfriendly and does not seem to fit anywhere in your bussiness model. If you want to focus on power-user capable firmware, then you might be much better off to port the GUI-configurable subset of your config format onto QMK (as a custom extension) and continue providing the Agent as a thing exclusive to a QMK-powered UHK, which I have a feeling (although unverified as I don't own any of those keyboard yet) would provide you with something even much more powerful than my firmware. After all, my firmware is a one-man project of some random unsattisfied guy, while QMK is a community effort with quite a strong drift. I have a feeling that choosing my macro engine might make sense only under assumptions which I struggle hard to imagine to come true (like the almost-complete rewrite of the engine on one side, and strong support for the macro language on the agent side - e.g., are you capable to provide Agent with a syntax aware macro editor (featuring autocomplete features, built-in documentation, and on-the-fly verification of the grammar)?). I know that apart from knowing angular and js sufficiently well, this all is within my expertise (and even at reasonable expenses), but I simply do not posses the power to do it now/"now", so making sense of it would be up to you.

kareltucek: Regarding the code, feel free to leave it as it is as it clearly is convenient for you at the moment - it really is not a big deal. I am just a bit overworked so I tend say what's on my mind without all applying all the proper filters of proper etiquette ;-).

kareltucek: Anyway, I am a already quite spent for today, so do not expect me to follow-up in turn...

mondalaci: It'd be possible to make navigation modes independent of layers. Still, I strongly believe it's practical to make them layer-dependent because the more independent modes are available, the harder it is to understand how the UHK works. Layers are quite central conceptually, so it's easy to reason about them.

I understand that your use case calls for more sophisticated configuration features. We have yet to see what features our customers desire, and we may introduce additional features if justified.

Thank you for elaborating on the implications of including your macro engine. I think it has merit because it allows for more sophisticated configuration without recompiling the firmware, and it provides features that people do want. But based on the issues you mentioned, it doesn't seem like a good idea to include it, indeed.

Still, I plan to implement declarative options via specially named macros to allow for advanced configuration scenarios. I also plan to implement module navigation modes and speed settings via special macros in the beginning. Later, we'll move them to the binary configuration and expose them in a more accessible manner via Agent.

right/src/mouse_controller.c
mondalaci: Thanks for the heads-up! Just updated the units.

mondalaci:
Thank you for explaining the difference of the trackpoint in detail! It makes sense, and it also highlights that your math is much stronger than mine. :)

I have yet to figure out exactly what options to expose per module. I'll merge your PR soon, and I appreciate your contributions very much!

right/src/mouse_controller.c
kareltucek: Thanks for clarifications!

> Later, we'll move them to the binary configuration and expose them in a more accessible manner via Agent.

Just a sidenote: I think that performance-wise, this is a decision which should be though through very carefully, since I see the binary format as the main source of rigidness of your development. From some point of view, sure it makes sense, but from other points of view it also may do a lot of harm. 

EDIT: I mean, this is exactly where you are paying the cake twice yet have nothing in the end insteand of paying one, yet eating it and still having one more in the end.

kareltucek: (Also note that my specific usecase is indeed specific, but the point of the usecase is relatively general, and therefore I think I am not going to be the only one to come across this.)

PR #337, opened 2021-03-14
fixes #332 tab backlight

### LINKED ISSUE #332: Tab backlight
This does not look entirelly right...
Tab

steamraven:


mondalaci:
@steamraven Thanks so much! Now only backspace is left to be fixed. The issue is possibly the same.

PR #341, opened 2021-04-04
Bugfix: slave count should be 8

steamraven:


steamraven:
FYI, this bug is probably leading to memory corruption as i2c errors are being logged outside the array bounds.

kareltucek:
Nice catch!

On my fork, this seems to correspond to:

```
 .bss.I2cWatchdog_RecoveryCounter
                0x000000001fff3914        0x4 build_make//opt/firmware/right/src/i2c_watchdog.o
                0x000000001fff3914                I2cWatchdog_RecoveryCounter
 .bss.I2cWatchdog_WatchCounter
                0x000000001fff3918        0x4 build_make//opt/firmware/right/src/i2c_watchdog.o
                0x000000001fff3918                I2cWatchdog_WatchCounter
 .bss.prevWatchdogCounter
                0x000000001fff391c        0x4 build_make//opt/firmware/right/src/i2c_watchdog.o
 .bss.I2cMainBusActualBaudRateBps
                0x000000001fff3920        0x4 build_make//opt/firmware/right/src/init_peripherals.o
                0x000000001fff3920                I2cMainBusActualBaudRateBps
 .bss.IsBusPalOn
                0x000000001fff3924        0x1 build_make//opt/firmware/right/src/init_peripherals.o
                0x000000001fff3924                IsBusPalOn
 .bss.AllKeymapsCount
                0x000000001fff3925        0x1 build_make//opt/firmware/right/src/keymap.o
                0x000000001fff3925                AllKeymapsCount
 .bss.CurrentKeymapIndex
                0x000000001fff3926        0x1 build_make//opt/firmware/right/src/keymap.o
                0x000000001fff3926                CurrentKeymapIndex
 .bss.DefaultKeymapIndex
                0x000000001fff3927        0x1 build_make//opt/firmware/right/src/keymap.o
                0x000000001fff3927                DefaultKeymapIndex
 .bss.KeyStates
                0x000000001fff3928      0x200 build_make//opt/firmware/right/src/key_states.o
                0x000000001fff3928                KeyStates
```

Which would explain some of my problems, including the weird baud rate...

right/src/slave_scheduler.h
```diff
@@ -9,8 +9,8 @@
 // Macros:
 
     #define SLAVE_COUNT (sizeof(Slaves) / sizeof(uhk_slave_t))
-    #define MAX_SLAVE_COUNT 6
-    #define IS_VALID_SLAVE_ID(slaveId) (0 <= slaveId && slaveId <= MAX_SLAVE_COUNT)
```
kareltucek, line 13: Also, I guess it would be better to throw the MAX_SLAVE_COUNT out so that this sort of error (i.e., forgetting to bump the value when actual SLAVE_COUNT changes) does not repeat in future...

kareltucek:
Looked into `origin/master` affected regions - looks the same within local build.

right/src/slave_scheduler.h
steamraven: Pushed a new commit to cleanup.  If the number of slave is greater than count, a warning is generated

right/src/slave_scheduler.h
```diff
@@ -8,9 +8,8 @@
 
 // Macros:
 
-    #define SLAVE_COUNT (sizeof(Slaves) / sizeof(uhk_slave_t))
```
mondalaci, line 11: @steamraven Why not keep the actual implementation of `SLAVE_COUNT`?

steamraven: I tried to just use that, but could not get it to compile.  It seams you can't specify the size of an array as reference to the size of the array. And if you leave the size blank, you get " invalid application of 'sizeof' to incomplete type 'uhk_slave_t[] "

mondalaci: You're completely right. Thanks for the explanation!

mondalaci:
@steamraven Thanks for the fix! Great catch!

PR #343, opened 2021-04-05
Pack memory maps with builds.

kareltucek:
If I am right in assuming that memory order of stuff is not guaranteed by the linker (or is it?), then memory maps are not easily reproducible. 

Packing them with every binary should help in memory corruption hunts...


mondalaci:
I don't know whether the linker guarantees memory order. I assume it's not guaranteed. If you think memory maps are useful for debugging purposes, we can include them into firmware releases, but then the module firmwares should have their own maps, too.

kareltucek:
I sure can add them too, but I don't think modules are that important since their construction is much simpler and therefore less prone to problems of this type. E.g., I do not remember any single memory corruption or race condition which would be caused by a problem in a module. It is the right half where all the strings come together in one complex knot.

mondalaci:
I generally agree, but still, it'd be nice to add memory maps for the modules, too, because we might found module-related bugs eventually.

kareltucek:
Sure, will add them in the evening!

kareltucek:
Done.

PR #344, opened 2021-04-05
Implement caret mode

kareltucek:


mondalaci:
I'm blown away by how sophisticated, and well-made your caret implementation is! Axis locking is especially thoughtful and useful.

- When using the trackball in caret mode, it feels like applying an exponential would make it even more useable.
- The horizontal axis of the touchpad should be inverted.
- As usual, the coding style... :)

This PR is great as it is, so please let me know if you plan to improve it soon. If not, I'll merge it, and we can improve it later.

kareltucek:
> I'm blown away by how sophisticated, and well-made your caret implementation is! Axis locking is especially thoughtful and useful.

:-)

>     * When using the trackball in caret mode, it feels like applying an exponential would make it even more useable.

What exactly do you mean? The standard acceleration driver is already applied to the caret mode. The only trouble might be that the actual speed is bounded by the usb polling rate (remember, one movement, one key tap) to approximately 500 ticks/s.

>     * The horizontal axis of the touchpad should be inverted.

Oh, I see I am applying the `yInversion` irrespectively of the actual axis.

>     * As usual, the coding style... :)

Thanks!

> This PR is great as it is, so please let me know if you plan to improve it soon. If not, I'll merge it, and we can improve it later.

I do not have any improvement plans atm.. I.e., I just plan to resolve the review findings.

mondalaci:
As for my exponential suggestion, it might be useful to use an exponential on top of the existing exponential.

kareltucek:
> As for my exponential suggestion, it might be useful to use an exponential on top of the existing exponential.

I am still quite unclear on the exact idea and motivations behind. 

I guess you are having troubles with activation of the axis locks, which decreases sensitivity of the first tick. I have pushed a tweak which increases first tick sensitivity of right side modules while keeping dampened sensitivity of the mini-trackball.

If you still think that additional acceleration is needed, I am offering this refactor: #352 (I have created new instance of `caret_mode` branch in your repo in order to be able to open a PR. Careful with actual merging.)

kareltucek:
> * As usual, the coding style... :)

Better now? :-)

mondalaci:
Thank you! Caret navigation feels noticeably faster, but now I often switch lines accidentally with the trackball when only trying to move the caret horizontally. If I had to choose, I'd choose the previous version, all things considered.

kareltucek:
Allright, will revert it.

kareltucek:
Done.

kareltucek:
On second though the current strongly nonlinear version is hardly ideal. I guess that leaving it with `1.0f` (halfway between the two versions) is correct. 

Regarding locking sensitivity - If you have the mentioned problems, then I think that indeed the correct solution is to apply an aditional acceleration on top of the thing (-> #352). (Acceleration affects locking properties just as much as the skew parameter, so they are interchangable.)

kareltucek:
But feel free to experiment with the values (in that case, I suggest that you check out the `caret_mode_exp` branch and try to fiddle with the values on lines 312 (acceleration) and 324 & 325 (the "first tick" skew)).

kareltucek:
On third though, I am no longer sure, so please pick the variant which feels the best to you.

PR #345, opened 2021-04-05
Add minimalistic debug library.

kareltucek:
This is just an offer of my private debugging mini library. 

It is quite handy when one needs to watch just a couple of variables. 

Usage:
- Define/uncomment `WATCHES` directive.
- Place some watches (i.e., macro calls) into the code, while specifying unique slot in range 0-6 as second parameter.
- While using the keyboard, tap one of the left half top row keys. 
- Led display will start watching that value every 500ms (i.e., first call of the macro after 500ms since last call is shown onto display and will remain for at least the next 500ms)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,3 +152,7 @@ Header files are composed of sections. The order of sections is fixed. Every hea
 ## Semantics
 
 The build process must not yield any warnings, and the build must pass [on Travis](https://travis-ci.org/UltimateHackingKeyboard/firmware).
+
+## Clang format
+
+There is a `.clang-format` present, which allows you to format your `.c` files simply by `clang-format -i <your source file>`. (It does **not** ensure the required header file structure though!) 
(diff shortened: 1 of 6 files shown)


mondalaci:
Your debug library seems quite handy! I'll merge it, but let me make a couple of suggestions first.

- `ShowNumberExp` might be a better name than `ShowNumberMag`. I'm not sure if it's worth exposing it because it can be used via other relevant functions. It would be more intuitive for me if the alphabetic exponent part were placed at the end, not at the beginning.
- The exponent notation shouldn't be used for numbers below 1,000.
- I think it's overkill to create a dedicated directory for these files for the time being. Let's use `right/src/debug.[ch]` for now.
- Please follow our coding standards.

kareltucek:
All good points, will refactor it.

kareltucek:
Resolved. 

Is the codestyle fine like this? (I am quite blind in this respect 😅.)

mondalaci:
Well done, and thank you! :)

PR #347, opened 2021-04-06
Fix release instructions in README.md

kareltucek:
Should decrease amount of confusion around misbehaving npm... https://github.com/UltimateHackingKeyboard/agent/issues/1475


mondalaci:
Looks like the points below "To build a full firmware tarball:" are not indented and numbered properly.

kareltucek:
I see, will fix it.

kareltucek:
A bit better now, although newline convention of IDE setup 3 is still not respected for some reason (no idea why).

PR #350, opened 2021-04-09
Squal

kareltucek:
Here is some code that shows trackball's surface quality value by default on the led display (slot 0, activated via left half's `~` key).

This is more for reference and testing than a serious PR, process it however you want. 

In any case if you decide to merge it, you will want to throw out the `WATCH_VALUE` call and `WATCHES`  define first...

### DIFF
--- a/right/src/dev_tools/utils_dbg.c
+++ b/right/src/dev_tools/utils_dbg.c
@@ -40,10 +40,10 @@ void WatchValue(int v, uint8_t n)
     }
 }
 
-void WatchString(int v, char const * c)
+void WatchString(char const * v, uint8_t n)
 {
     if(CurrentTime - lastWatch > watchInterval) {
-        LedDisplay_SetText(3,  c);
+        LedDisplay_SetText(3,  v);
         lastWatch = CurrentTime;
     }
 }
--- a/right/src/dev_tools/utils_dbg.c
+++ b/right/src/dev_tools/utils_dbg.c
@@ -56,13 +56,19 @@ void ShowNumberMag(int a) {
     if(num < 0) {
         LedDisplay_SetText(3,  "NEG");
     } else {
-        while(num >= 100) {
-            mag++;
-            num /= 10;
+        if(num < 1000) {
+            b[0] = '0' + num/100;
+            b[1] = '0' + num%100/10;
+            b[2] = '0' + num%10;
+        } else {
+            while(num >= 100) {
+                mag++;
+                num /= 10;
+            }
+            b[0] = mag == 0 ? '0' : ('A' - 1 + mag);
+            b[1] = '0' + num/10;
+            b[2] = '0' + num%10;
         }
-        b[0] = mag == 0 ? '0' : ('A' - 1 + mag);
-        b[1] = '0' + num/10;
-        b[2] = '0' + num%10;
         LedDisplay_SetText(3,  b);
     }
 }
--- a/right/src/dev_tools/utils_dbg.c
+++ b/right/src/dev_tools/utils_dbg.c
@@ -1,0 +1,70 @@
+#include "utils_dbg.h"
+
+#ifdef WATCHES
+
+#include "led_display.h"
+#include "timer.h"
+#include "key_states.h"
+
+uint8_t CurrentWatch = 0;
+
+static uint32_t lastWatch = 0;
+static uint32_t watchInterval = 500;
+
+void TriggerWatch(key_state_t *keyState)
+{
+    int16_t key = (keyState - &KeyStates[SlotId_LeftKeyboardHalf][0]);
+    if(0 <= key && key <= 7)
+    {
+        //set the LED value to RES until next update occurs.
+        LedDisplay_SetText(3, "RES");
+        CurrentWatch = key;
+    }
+}
+
+void WatchTime(uint8_t n)
+{
+    static uint32_t lastUpdate = 0;
+    if(CurrentTime - lastWatch > watchInterval) {
+        ShowNumberMag(CurrentTime - lastUpdate);
+        lastWatch = CurrentTime;
+    }
+    lastUpdate = CurrentTime;
+}
+
+void WatchValue(int v, uint8_t n)
+{
+    if(CurrentTime - lastWatch > watchInterval) {
+        ShowNumberMag(v);
+        lastWatch = CurrentTime;
+    }
+}
+
+void WatchString(int v, char const * c)
+{
+    if(CurrentTime - lastWatch > watchInterval) {
+        LedDisplay_SetText(3,  c);
+        lastWatch = CurrentTime;
+    }
+}
+
+
+void ShowNumberMag(int a) {
+    char b[3];
+    int mag = 0;
+    int num = a;
+    if(num < 0) {
+        LedDisplay_SetText(3,  "NEG");
+    } else {
+        while(num >= 100) {
+            mag++;
+            num /= 10;
+        }
+        b[0] = mag == 0 ? '0' : ('A' - 1 + mag);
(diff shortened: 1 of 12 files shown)


mondalaci:
Great job at noticing the SQUAL value (of which I didn't know), and excellent job at exposing the value via the LED display. I've made some tests with the 3 balls we currently provide with the trackball module.

- The red plastic ball is my favorite, and its tracking is flawless. Its SQUAL value is 30.
- The black resin ball generally has good tracking, but the sensor doesn't detect swift motions from time to time. Its SQUAL value is 40.
- The steel ball has flawless tracking, and its SQUAL value is 60.

Given the above findings, the SQUAL value doesn't precisely correlate with tracking accuracy, and it's rather a development feature, so I rather wouldn't merge this PR.

Nevertheless, seeing the SQUAL value is interesting and educational, so anyone interested is welcome to give a shot: 
[uhk-firmware-squal-8.10.8.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/6289779/uhk-firmware-squal-8.10.8.tar.gz)

Even though I don't merge this PR, I highly appreciate your work, and please keep it up!

kareltucek:
(Just for the record, I would love to be able to do testing too, have prepared some resin balls with metallic pigment of my own, but without a functional trackball module, I simply cannot :-).)

> Given the above findings, the SQUAL value doesn't precisely correlate with tracking accuracy, and it's rather a development feature, so I rather wouldn't merge this PR. 

What about merging it into a separate development branch? If that's fine with you, I will implement reading of the "camera" output, do some refactors and then it can live in your repository... would prefer that over having to juggle with a dozen of branches in my repos.

> The black resin ball generally has good tracking, but the sensor doesn't detect swift motions from time to time. Its SQUAL value is 40.

Just out of curiosity what value will you get if you polish the ball gently with a bit of resin polish paste (in order to simulate what happens to the ball after some months of use)?

kareltucek:
>     * The black resin ball generally has good tracking, but the sensor doesn't detect swift motions from time to time. Its SQUAL value is 40.

But that's not a problem of the SQUAL command. If you read the datasheet, you will notice that is uses dynamically adjusted shutter interval. 

In order to pick up black surface, it obviously will need to use much longer shutter interval in order to get the same amount of light. Results are two:

- the image is going to be over-exposed. Yielding something that can detect many details on the ball, yet overall, the image is probably pretty bad quality. 

- fast movement is going to be blurred because of the longer shutter interval (and therefore miss out at high speeds). (*Edit: It is not about blur - much more about framerate, because that has to drop significantly in order to pick up light.*)

kareltucek:
Will expose that, and probably also some other values too...

mondalaci:
Thanks for the explanation, but then, what the SQUAL value is supposed to indicate in practical terms?

How about forking this repo as a firmware-trackball-dev repo under your account? I think the suggested features are amusing and educational, but I'd rather keep this repo as clean as possible and not even have any extra branches.

I haven't forgotten about you, and we'll send you one of each production module and plenty of balls to play with. I'm waiting for a final ball; then we'll send your package. I wouldn't like to spend time polishing balls, but I'll make sure to send you extra balls for this purpose.

kareltucek:
> Thanks for the explanation, but then, what the SQUAL value is supposed to indicate in practical terms?

From the data sheet: 

> SQUAL (Surface Quality) is a measure of the number of valid features visible by the sensor in the current frame. The maximum SQUAL register value is 167. Since small changes in the current frame can result in changes in SQUAL, variations in SQUAL when looking at a surface are expected. 

It just tells you data which are rock solid. Making out conclusions whose reliability is hard to quantify is up to you. 

>  How about forking this repo as a firmware-trackball-dev repo under your account? I think the suggested features are amusing and educational, but I'd rather keep this repo as clean as possible and not even have any extra branches.

It seems like I will have to :D. Fair enough.

kareltucek:
Allright, the new module has arrived, an therefore - here goes some data:

IMG_20210420_182304


 |  | Min squal | Shutter | Observed rating | Computed rating | Note |
 |  :----- | :-------------:  | :-------------:  |  :----------:  |  :-----------:  | -----: |
 | Production balls (matte) |  |  |  |  | |
 | Steel ball | 40 | 100 | 500 | 400 |  |
 | Red Plastic ball | 13 (avg 30) | 200 | 100 | 65 |  |
 | Black resin ball | 45 | 400 | 170 | 128 |  |
 | Modified prod balls |  |  |  |  |  |
 | Polished black resin | 20 | 1000 (600-2000) | 40 | 20 |  |
 | Custom resin balls (all shiny) |  |  |  |  |  |
 | Blackish | 60 | 400 | 170 | 150 |  |
 | Bronze (J&L) | 90 | 70 | 1100 | 1285 |  |
 | Yellow (J&L) | 70 | 60 | 1400 | 1166 |  |
 | Blue (J&L) | 60 | 600 (400-2000) | 100 | 100 |  |
 | Red | 110 | 200 (100-500) | 500 | 550 |  |
 | Custom resin balls (all shiny) gen 2 |  |  |  |  |  |
 | Wine Red (mica) | 80 | 100 (60-300) | 600-1500 | 800 |  |
 | Bronze (mica) | 100 | 60 (30-150) | 1300 | 1666 |  |

Column meaning:

- Squal is taken as minimum encountered value (the minimum is computed by firmware across 500ms intervals, which are then shown on the display).
- Shutter is a rough average.
- Observed rating and Computed rating are taken by the formule below. Observed one is computed on board from current values, while the computed one is computed from the first two columns. (The computed one uses the minimum squal while observed one uses current value in the firmware, therefore the differences.)

My conclusions (or rather draft hypotheses) from this are:

- All tested balls have a fair surface quality. Therefore the most important value is probably the shutter interval (which I believe stands for sharper image and better framerate of the sensor). The shorter the better. The ratings in the table are computed by `squal * 1000 / shutter`. 

- Therefore the most important column is probably shutter interval. Obviously lighter colours yield shorter values and should theoretically behave better with fast movement. (This correlates with your liking of the red and steel balls. The rating metric is obviously not perfect.)

- While the matte black balls have relatively low shutter (~400), a polished black ball has still decent squal, but significantly worse shutter (800-2000). This (probably) means that specular highlight is (correctly) reflected away from the sensor. Consequently, diffuse/subsurface texture is picked up by the sensor. This is good news. Probably means that all your problems are in the balls and not in the housing.

- High shutter value for shiny balls might mean that bad pickup quality of your shiny balls was most likely caused by combination of lack of subsurface texture and insufficient amount of light that could be picked up by the sensor. (Since shiny balls reflected most of the light away from the sensor, and consequently there was not enough diffuse light to be picked up.)

- Unfortunately, the custom balls are all pretty low quality so I cannot actually test how fast motion fares in practice. This makes the analysis an entirely theoretical thing. 

- Custom balls are all shiny resin balls with some added pigment to give them a grainy glittering subsurface texture. I will probably make some higher quality ones sometime in the future...

kareltucek:
(If anyone wants to try it out - [uhk-firmware-8.10.8.squalv2.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/6345758/uhk-firmware-8.10.8.squalv2.tar.gz) . ~ = squal, 1 = min squal, 2 = shutter, 3 = rating)

kareltucek:
The squal branch is now able to stream data from the sensor and show them on screen in "real" time. Some shots either at:

- http://ktweb.cz/uhk_balls/
- or https://github.com/kareltucek/uhk-squal_research-agent/tree/master/packages/usb/Balls

The code now lives in separate repositories:

- https://github.com/kareltucek/uhk-squal_research-agent
- https://github.com/kareltucek/uhk-squal_research-firmware

PR #352, opened 2021-04-14
 Apply additional acceleration to caret mode. 

kareltucek:


mondalaci:
I've been playing with various values as suggested, but the master branch feels much better, so I'm closing this PR.

Thanks for tweaking it regardless!

PR #356, opened 2021-04-19
Fix caret mode

### LINKED ISSUE #354: Navigation mode bugs
@kareltucek

- When scrolling with the mini trackball of the key cluster, it scrolls quite a large distance at a time. It feels like it emits double or triple scroll actions on every pulse.
- Vertical direction is reversed for the touchpad and for the mini trackball of the key cluster in caret mode.

kareltucek:
Closes #354.


mondalaci:
Thanks so much for the quick fix!

PR #358, opened 2021-04-19
Port keystroke delay.

### LINKED ISSUE #353: Feature request:  Add modifier key delays to keys bound with modifiers
Mappings of keys with modifiers such as Alt+Tab or Ctrl+Shift+Esc using the direct, naïve approach of a key plus mod keys directly in the UI is currently next to useless when using Windows Remote Desktop Protocol.  This is an issue in this time of pandemic-enforced homebound work.  I have observed similar issues when working through TeamViewer, so it seems to not be isolated to the RDP.

From what I can gather, the issue is that the UHK sends the keys "too quickly" and that the RDP program or protocol, or some layer in-between, gets confused about which one was pressed first, or only registers one or some of the keys.  A solution, or at least work-around, would be to delay each key event slightly to allow the RDP or similar program to parse the key events correctly.  From testing with a custom firmware, 5-10ms delay seems to do the trick.

Your stance stance on this, from an earlier report of the same issue, was along the lines of "The UHK adheres to the specs of the keyboard USB protocol and the issue is in the RDP implementation.  We will not do firmware work-arounds for software issues in other products."

While I do agree with the analysis and can see the merrit in the stance, the fact of the matter is that Microsoft is not going to address this for the insignificant minority of their customers who has this issue, and who frankly have no choice but to use their OS for work anyway.  So no fix is forthcoming for the software bug in other products.

As such, this issue does detract greatly from the value of your product for anyone using RDP for their daily work, to a point where I would advise people to avoid your product if they use RDP extensively.  Had I known of this issue, I certainly would have thought twice before purchasing your product, which is a shame because I absolutely love every other aspect of it.

I would implore you to reconsider your previous stance on this issue and address it, such that the full range of features of your keyboard can become available for those of us who have to use RDP for work.

kareltucek:
Closes #353.

### DIFF
--- a/right/src/postponer.c
+++ b/right/src/postponer.c
@@ -52,6 +52,13 @@ bool PostponerCore_IsActive(void)
 void PostponerCore_TrackKeyEvent(key_state_t *keyState, bool active)
 {
     uint8_t pos = POS(bufferSize);
+
+    //if the buffer is totally filled, at least make sure the key doesn't get stuck
+    if (bufferSize == POSTPONER_BUFFER_SIZE) {
+        buffer[pos].key->current = buffer[bufferPosition].active;
+        consumeEvent(1);
+    }
+
     buffer[pos] = (struct postponer_buffer_record_type_t) {
             .key = keyState,
             .active = active,
--- a/right/src/usb_report_updater.c
+++ b/right/src/usb_report_updater.c
@@ -28,6 +28,8 @@ static key_action_t actionCache[SLOT_COUNT][MAX_KEY_COUNT_PER_MODULE];
 
 volatile uint8_t UsbReportUpdateSemaphore = 0;
 
+static uint16_t keystrokeDelay = 0;
+
 // Holds are applied on current base layer.
 static void applyLayerHolds(key_state_t *keyState, key_action_t *action) {
     if (action->type == KeyActionType_SwitchLayer && KeyState_Active(keyState)) {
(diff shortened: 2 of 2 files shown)


kareltucek:
(This, if set to non-zero value, currently slows down all firmware scancode actions such as macros or caret mode. Atm., I consider it a feature...)

mondalaci:
I set the delay to 300. The first character appeared instantly, then rapidly typed characters afterward appeared with significant delay. The delay wasn't 300 ms, but more like a second, and shortly, some of the keys behaved as they got stuck.

kareltucek:
> I set the delay to 300. The first character appeared instantly,

This is correct.

> then rapidly typed characters afterward appeared with significant delay. The delay wasn't 300 ms, but more like a second, and shortly, 

300ms is interval between report changes. Since every keystroke consists of  two such changes (keydown and keyup) or more (extra modifier reports), the observed delay should be 600ms. Give or take 30% on inaccuracies of the firmware timer.

> some of the keys behaved as they got stuck.

The postponer's queue is 32 events deep, which is 16 keystrokes. If it gets almost filled, it will usually start replaying keys one by one even if the postponing mode is kept active, except for the keystroke delay which is honored in all situations. 

So if you set this ridiculously long delay and overfill the buffer, then I guess loosing some of he events is an almost correct behaviour. Except that having keys stuck is not good -> fixed.

mondalaci:
The stuck keys issue got resolved, indeed, but

1. When I use the 300ms value, type "asdf" quickly, and don't hold the "f" key, I get "asdffffffffffffffffff".
2. When the postponer kicks in, module pointer movement becomes jaggy.

I'm unsure whether the above is considered normal, but they're not ideal.

kareltucek:
> The stuck keys issue got resolved, indeed, but
> 
>     1. When I use the 300ms value, type "asdf" quickly, and don't hold the "f" key, I get "asdffffffffffffffffff".

Standard "autorepeat" delay of most operating systems is 400ms. Which means that yeah - by definition the behaviour is correct. (And at 5ms resolution is a non-issue.)

>     2. When the postponer kicks in, module pointer movement becomes jaggy.
> 
> I'm unsure whether the above is considered normal, but they're not ideal.

This is again specific to the keystroke delay, not to the postponer. Since I never move cursor while writing heavily at the same time, I consider it a non issue. Of course the situation can be improved, but would probably need some refactors and would make the code more complex.

mondalaci:
I've been testing more in the meantime. Even a 10ms delay interferes with the caret mode substantially, as caret navigation becomes slower, and I often change lines in the IDE inadvertently.

kareltucek:
Sorry for the delay - this inquiry needed a bit testing and thinking on my part... unfortunately while testing the 10ms delay + caret mode, I have ran into strange problems (namely caret mode getting stuck) that so far withstanded all my debugging attempts.

But to the point:

> I've been testing more in the meantime. Even a 10ms delay interferes with the caret mode substantially, as caret navigation becomes slower, and I often change lines in the IDE inadvertently.

The keystroke delay inherently does slow down the caret navigation. That is known since by definition it is supposed to make firmware send at most one usb state change per the delay interval. After looking into the code, I am afraid I am not able to make an exception for the caret mode in a simple manner. 

Also, axis locking breaks due to the keystroke delay, since over the 10ms chunks of time, much larger deltas are fed into the algorithm - namely sufficiently large to override current lock. I can deal with this by resetting the module deltas over the delayed time. This is also likely to solve the problem which I mentioned in the beginning, but unfortunately without uncovering the root cause.

The question is whether such solution is desired.

mondalaci:
@kareltucek Thanks for your PR and for addressing the pointer delta issue!

PR #367, opened 2021-05-06
Request for testing: Full N-Key support

### LINKED ISSUE #70: NKRO with 6KRO BIOS fallback
Currently the firmware exposes a basic 6KRO keyboard USB interface. We'll expose an NKRO keyboard interface, but its compatibility will be limited (no BIOS support) so it'd probably be better to to leave the 6KRO keyboard USB interface instead of replacing it, and add an NKRO interface, too. Then the firmware would detect which interface is polled. If the NKRO interface is polled, then use that. Otherwise fall back to the 6KRO interface.

steamraven:
Resolves #70.

This provides full, true nkey support.  This is one of three ways to implelemnt:

This implemented via a bitfield with one bit for every hid usage (scan-code).  All 222 of them. Plus the 8 modifier keys. 

 N-key is implemented only on the non-boot interface, so if a device uses the boot keyboard interface (like your bios firmware) it will get a normal, compatible keyboard.

Firmware bundle available here:
https://github.com/steamraven/firmware/releases/tag/v8.10.9

Pros: 
* Full n-key
* Fast path in the keyboard driver

Cons:
* Does not support scancodes above 0xE8
* Unknown compatibility (especially with smaller embedded devices), though they should use the boot keyboard interface.

eh:
* Middle of the pack packet size: 30 bytes

There are two other implementations:

## 62 key Scancode list

This basically takes the 6 key scancode list, and expands it to the max usb full-speed packet size.  This enable 62 simultanious keys.  Since there are 28 (left) + 30 (right) + 3 (module)  =61 non-modifier keys, this gives full n-key support
Pros:
* N-key for normal operation
* supports full 8-bit scancodes

Cons:
  * Slow path in the driver.  It executes a loop 2 * 62 * 62 = 7688 times per report
  * Large packet size - 64 bytes
  * Barely n-key. If you use modifier keys for real-scancodes, may get past 62 real keys
 
## Hybrid:

This is hybrid of the two. It has bit field for the first 0x6C scancodes, and a 6 key list of scan-codes.  This covers all of the PC-AT and Mac scancodes within the bitfield
Pros:
* Fast path in the keyboard driver: bit-field + only 6 scancodes
* Smallest packet size: 21 bytes
* n-key for normal operation
* supports full 8-bit scancodes

cons:
* If you use a lot of keys with scancodes over 0x6C, you may not have full n-key

If you would like to test these versions, I can put a together a firmware release


mondalaci:
@steamraven Thanks so much for your PR and the detailed description! Your N-key rollover implementation works great on Linux and Windows 8, but it triggers all kinds of characters, keyboard combinations, and media keys on Mac.

@abcminiuser @xxxajk You guys are tremendously knowledgeable about USB. How do you suggest implementing N-key rollover in a cross-platform way and with boot protocol support?

abcminiuser:
Boot protocol support (the old legacy fixed report structures used to support simple hosts that lack a full report parser) is entirely orthogonal to the regular USB HID report protocol, and won't be affected by this.

I was actually going to start work on last last weekend only to find the only machine in the house old enough to have the USB 2.0 ports and controller needed to run the compliance test suite has died, so I need to source some other test hardware first to ensure I get all the implementation details right.

When the keyboard or mouse boot protocol is advertised on a HID interface, and the host enables it, the normal HID report structure of that interface is disabled, and the fixed report format specified in the HID documentation takes its place instead. The one further non-spec caveat to this is that MacOS (at least up to Catalina) will break horribly on boot protocol interfaces that have an OUT endpoint for host-to-device data, instead of using the control endpoint.

For N-Key rollover, I'll need to do some research as it's not something I've had to implement myself (always using the standard 6 key rollover support along with a key matrix that uses a diode per key to prevent ghosting). I'm certain compatibility will be the sticking point more than anything else, as I've had horrible troubles only recently trying to make "standard and well specified" implementations work across all the various OSes.

steamraven:
@mondalaci It is unfortunate that it is not working on Macs. Unfortunately, I was not able to test Macs, but I was really hopeful based on my Windows and Linux test.

All is not lost, though.  I found the the openbsd driver code, and it hasn't had much change in 5 years. And I bet it mostly the same in macOS ( I really wish more oses were open source).   Looking through the code, there is a hard limit of 128 bit fields and 6 scancodes.  The hybrid approach falls with this range, using 112 bit fields, and 6 element scancode list. 

Here is that branch rebased to the latest on master. I wish I had a Mac to test on, but this works under windows 10 and linux
https://github.com/steamraven/firmware/releases/tag/v8.10.9-nkey-hybrid

The branch this is build from is:

https://github.com/steamraven/firmware/tree/nkey-hybrid

If testing works, I will update this pull request.

mondalaci:
@steamraven I've just tested your branch, and the issue persists on OSX.

How about implementing NKRO [this way](https://groups.google.com/g/lufa-support/c/xXztpOFT624/m/-8x93W537WcJ)?

mondalaci:
I'm unsure whether I should have tested this PR, but I've just tested it, and it exhibits the same issues on MacOS as previously.

steamraven:
@mondalaci That is pretty weird. It is working fine on MacOS BigSur 11.4. Could it be a language or config setting?

Things I tested on MacOs:
* All standard keystrokes
* All Modifiers
* Media Keys
* Mouse Keys
* Macro
* Keys with modifiers
* High scancodes (106,107, 145)

Here are the firmwares using all three methods (including the one linked by @mondalaci ). Each have a relevant branch.
https://github.com/steamraven/firmware/releases/tag/v8.10.10-nkey-bitfield
https://github.com/steamraven/firmware/releases/tag/v8.10.10-nkey-hybrid
https://github.com/steamraven/firmware/releases/tag/v8.10.10-nkey

mondalaci:
@steamraven Your change actually works! Sorry, it was totally my bad. My git repo was messed up, I re-cloned it, flashed the firmware, and it works flawlessly with NKRO on, even on MacOS.

@kareltucek @abcminiuser It'd give me peace of mind if you guys could review this PR.

kareltucek:
> It'd give me peace of mind if you guys could review this PR.

Oops, did not notice your request.

Looks good, just two notes:
- `processTextAction` (macro text action) does not seem to be converted, so it is probably going to do funny things.
- It might be a good idea to abstract the pieces of code that add/remove basic scancodes - for better readability.

benedekkupper:
The commits look good from USB HID point of view. There is a possible synchronization bug when getting the HID protocol mode, I have submitted #439 which has the fix commits from this PR and another on top to facilitate getting the correct protocol mode. In addition, it would probably make sense to skip sending an assembled packet while the protocol has changed (when the packet has different layout in different protocol modes).

I'm not sufficiently familiar with the overall codebase to perform the rebase myself, as there was a major macro overhaul (53535a9) merged in recently that brings in a lot of conflicts.

mondalaci:
@steamraven @benedekkupper @kareltucek Can any of you guys resolve the conflicts? Then I'd test it thoroughly and merge it.

kareltucek:
I will take a look, I guess they are consequence of the fork squash merge.

kareltucek:
@mondalaci @benedekkupper https://github.com/kareltucek/firmware/tree/nkey-bitfield_rebased is this PR with uhk/master merged in. I have resolved conflicts, refactored report handling and fixed logic around the macro engine, but I lack deeper understanding of the USB subsystem, so it would be ideal if you @benedekkupper could take over it again. 

It seems to work fine, except first scancode doesn't get sent.

mondalaci:
@kareltucek What do you mean by that the first scancode doesn't get sent?

kareltucek:
@mondalaci :

- powercycle the UHK
- (optional) give it a few seconds
- (optional) tap a key to switch keymap (works as expected)
- (focus text editor)
- tap 'a' (nothing happens)
- tap 'a' ("a" appears in the text editor)

mondalaci:
@kareltucek Thanks for elaborating! The mentioned bug is a show-stopper. Now that common trunk is merged, would you please rebase this PR? It'd be much easier if we could work with this repo. Afterward, @benedekkupper will do his magic.

kareltucek:
Here you go: https://github.com/UltimateHackingKeyboard/firmware/tree/nkey-rollover

mondalaci:
@kareltucek Thank you! I can reproduce the issue with the `nkey-rollover` branch.

benedekkupper:
I have done my "magic", see #448. I revised the commits and put them in order for easy bisecting. Not tested at all yet, some testing / reviewing would be appreciated.

mondalaci:
Superseded by #448 which just got merged.

Issue #314, opened 2020-11-01
USB resets on Ryzen via USB 2.0 happen frequently on Linux with firmware 8.10.1


lo48576:
## Environment
* Gentoo Linux, kernel 5.9.2
    + Also with 5.8.14, 5.8.17 and 5.9.0, same problem happens.
* UHK is connected to USB 2.0 port, and XHCI driver is used (xhci_pci, xhci_hcd)
* UHK firmware 8.10.1

## Expected result
UHK works normally.

## Actual result
With linux kernel built with `CONFIG_HIGH_RES_TIMERS=y` **OR** `CONFIG_NO_HZ_IDLE=y` (or both), UHK is frequently disconnected (i.e. key inputs are ignored intermittently) and `dmesg` reports errors:

> [  286.223211] usb 1-10.3: reset full-speed USB device number 7 using xhci_hcd

![screenshot-2020-10-27-214313+0900](https://user-images.githubusercontent.com/1246590/97801137-f31a2a00-1c7d-11eb-8513-3a90d1004f0e.png)

### Workaround
If you are suffering from this, build the linux kernel with `CONFIG_HIGH_RES_TIMERS=n`, `CONFIG_NO_HZ_IDLE=n`, **AND** `CONFIG_HZ_PERIODIC=y`.

### Cause
~~I'm almost sure the patch for #269 caused this problem.~~
(I tried to confirm that, but I couldn't. The USB reset happened during firmware flashing, and left half of the keyboard seems to be bricked :rofl:)

**EDIT**: With firmware 8.10.0 (only with right half of the keyboard), this error still happens. Now I don't know why this error started to happen recently...
**EDIT2**: The error still happens with frimware 8.10.2 (most recent release).

lo48576:
I plugged the keyboard to another Gentoo Linux machine (laptop), and then the error does not seem to happen, even with `CONFIG_HIGH_RES_TIMERS=y` and `CONFIG_NO_HZ_IDLE=y` and `xhci_hcd`...
I don't know what to check more.

mondalaci:
What happens if you connect your UHK via a USB 3 port or a powered USB hub?

lo48576:
Powered USB 2 hub (which is connected to USB 2 port of M/B): problem still happens.
USB 3 port: The problem doesn't seem to happen.

Unplugging from USB 3 port, this warning is reported by `dmesg`, but I don't know this is related.

[...]

PR #369, opened 2021-05-08
Fix up mouse HID report descriptors.

abcminiuser:
I've noticed the key cluster trackball (mouse X/Y scrollwheel) is quite janky on my system, with movements being rather chaotic. I haven't had any luck with fixing that yet today - first I assumed the trackball was actually dual quadrature rather than pulsed axis, then I tried using ISRs to handle the pin changes just in case the MCU was missing them.

While testing, I did find a few oddities in the HID descriptors that should be corrected:

1) Constant padding entries should be marked as CONSTANT only, without DATA or other flags
2) The mouse X/Y wheel reports had some bizarre feature reports listed for configurable resolution multipliers that weren't actually implemented in the firmware
3) The mouse HID descriptors have some PUSH/POP HID entries in them that are both not required, and will have some very "interesting" (read: incorrect) behaviors on some systems.

This PR cleans up the HID report descriptors so we at least advertise to the host reports that we will actually send, or will actually consume.

The `bcdDevice` bump is to force Windows and other hosts to re-fetch the descriptors when the device re-enumerates, rather than using any cached versions.

This PR also changes the reported endpoint size of the system HID interface from 1 byte to the more standard 8 bytes; even if the extra bytes aren't used, it's better to report standard endpoint sizes for broader compatibility.


mondalaci:
@abcminiuser Thanks so much for these fixes! We would have never noticed the USB-related issues. Do you think it was the cause of #314?

abcminiuser:
Hard to say, but I suspect that's more of an issue on the Linux end. The descriptors for the device were advertising reports that the device wouldn't accept, which could cause all sorts of issues, as well as using push/pops which are frequently implemented poorly.

More likely the device lacking support for boot protocol support is what's broken support with some KVMs and BIOS, and the lack of support for the mandatory HID interface requests like SetIdle and GetReport will break various hosts.

The best place to start would be to download and run the keyboard firmware (as it is now) against [the USB-IF compliance test tool](https://www.usb.org/document-library/usb20cv-x64-bit) to see how many tests we fail. I've been trying to do that to see what needs to be fixed, but lack a Windows host with the required EHCI controller at the moment. If you *do* have suitable hardware, try running the full Chapter 9 tests along with the HID tests and send me the generated report - that will greatly help.

Getting 100% compliance with the tool doesn't guarantee no compatibility issues, but it at least gives us a fighting chance to work against most hosts.

mondalaci:
@abcminiuser You piqued my interest regarding the USB compatibility tool, which I was unaware of. I think my ThinkPad T520 has the required EHCI controller, and I've installed the tool, but I get the following upon pressing the Run button:

usb

Looks like a permission issue. When right-clicking on the icon of the tool, the "Run as Administrator" option is unavailable. I'm not exactly a Windows guru. Any ideas?

abcminiuser:
Hrm, your user directory shouldn't be protected against apps run with your user account - more likely it's the (lack of) UTF-8 awareness in that tool that's breaking it (note the wonky characters in the path). The tool is fairly...errr...engineering quality, so I wouldn't be surprised. It might be worth making a new account with a latin name and see if that fixes it.

There's a newer 3.0 compliance tool that will run on newer xHCI USB host controllers, but I suspect it'll have the same issues even  if you have compatible hardware. It's been a frustrating week for me; of all the systems in my house, they either have a single USB controller (and I don't have any PS2 mice to control the app while the tester takes over the controller), have hardware failures since I last tried to boot them, or both.

I've ordered a cheap USB3 PCI Express card from eBay, so I should (finally) be able to run the tests myself later this week.

At a minimum you're going to fail the HID specific tests due to missing mandatory HID class requests. I've started work on those in https://github.com/abcminiuser/uhk-firmware/tree/feature/hid_class_requests but need to understand the timer system in the codebase before I can implement the SetIdle request. Once those are fixed, there's still legacy boot protocol to consider, as well as fixing any other failures the compliance tool points out.

mondalaci:
Thank you! Then I'm abandoning my attempts at the USB compliance too, and waiting for you. Excited about your findings!

PR #370, opened 2021-05-09
Ensure LEDs are updated on the key cluster when it is initialized.

### LINKED ISSUE #366: Key cluster has wrong colors when attached
1. Set up keymap for key cluster so that it has different colors
2. Detach your key cluster module
3. Reattach your key cluster module
4. Observe that key cluster has default (?) colors, not the same colors before detach
5. Switch layer, observe that colors are correct now

abcminiuser:
Fixes #366.

When the key cluster is hot-swapped in, the LED driver wasn't pushing the initial LED driver states over correctly. This fixes that (although it's not immediately obvious to me why the module ID checks are needed at all) and updates the slave scheduler logic to initialize the modules when the firmware boots, and once when a module is connected (rather than continuously when disconnected).

I've tested by setting the cluster keys to various special modifiers, then hot-swapping it in and out. Without this patch the cluster LEDs are all a off-shade white (not the same as regular key white) and with the patch all the initial colours are set correctly.

### DIFF
--- a/right/src/slave_drivers/is31fl3xxx_driver.c
+++ b/right/src/slave_drivers/is31fl3xxx_driver.c
@@ -208,7 +208,7 @@ void LedSlaveDriver_Init(uint8_t ledDriverId)
     currentLedDriverState->ledIndex = 0;
     memset(LedDriverValues[ledDriverId], KeyBacklightBrightness, currentLedDriverState->ledCount);
 
-    if (ledDriverId == LedDriverId_Left) {
+    if (ledDriverId == LedDriverId_Left || ledDriverId == LedDriverId_ModuleLeft) {
         UpdateLayerLeds();
         LedDisplay_UpdateAll();
     }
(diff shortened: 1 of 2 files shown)


abcminiuser:
@mondalaci happy to split this into just the fix and a separate pull for the refactoring, if you prefer.

mondalaci:
@abcminiuser Thanks, but I've noticed an issue with your PR. When reconnecting the left keyboard half, quite often, its LEDs don't turn on. In the meantime, I've fixed the key cluster LED initialization issue with 5d4645179bca5edb5e1c605c37293d541abfa46a.

Theoretically, your fix should work, and I don't understand why it doesn't because the state machine is rather complex. Feel free to fix your PR, or we can abandon it if you choose so.

abcminiuser:
If you have a better fix, let's abandon this as I fully admit I don't entirely understand the full update state machine yet.

Issue #219, opened 2019-04-25
Unable to wake up OS X


mrekucci:
I'm unable to wake up OS X from sleep after the time set in `Turn display off after` from `System Preferences -> Energy Saver` elapses (see the screenshot below). I'm able to do it with another keyboard.

Tested with the following HW:
macOS High Sierra Version 10.13.6
MacBook Pro (15-inch, 2017); 2,9 GHz Intel Core i7; 16 GB 2133 MHz LPDDR3

macOS Mojave Version 10.14.2
Mac mini (Late 2018); Intel Core i7-8700B; 8 GB 2666 MHz SODIM DDR4
 
UHK Firmware: 8.5.4
UHK Agent: 1.2.12

Tested with different values of `Turn display off after`.

<img width="667" alt="Screenshot 2019-04-24 at 22 03 40" src="https://user-images.githubusercontent.com/4932785/56738267-8b238300-676c-11e9-8d07-5dfbaa72b8ae.png">

mondalaci:
I can't reproduce this issue. I'm running OSX 10.9.5 on a 2.4 Ghz Intel Core 2 Duo MacBook Pro.

@eltang @Lauszus Any luck with reproducing this?

jcostlow:
I can duplicate this.

MacOS 10.14.4, Macbook Pro
UHK 8.5.3

The UHK is connected to a USB-C hub which also has another keyboard and an HDMI monitor plugged in. If just the screensaver goes off, the UHK operates normally and I have to enter the password. However if the machine falls asleep, UHK key presses do not wake the computer. Pressing any keys on the other keyboard does wake it.
I have not tried disconnecting the other keyboard yet.

mondalaci:
@mrekucci Do you use dongles / hubs for connecting the UHK to your computers? If so, which types?

mrekucci:
With the Mac Mini I don't. With the MacBook Pro I use [Apple USB-C Digital AV Multiport Adapter](https://www.apple.com/shop/product/MJ1K2AM/A/usb-c-digital-av-multiport-adapter?fnode=8b).

forficate:
Same issue.

Agent: 1.2.12
Firmware: 8.5.3
OSX: 10.14.4

Imac 27" 2017.

Keyboard is connected to USB port directly on the Mac. To wake up I need to move/click a mouse button, keyboard will not wake up the machine.

mondalaci:
Thanks for the reports! We'll look into this.

[...]

PR #379, opened 2021-05-23
Replace incorrect/hard-coded GetDeviceStatus flags with correct remote wake-up information.

### LINKED ISSUE #373: USB 2.0 Compliance: Incorrect DeviceStatus request handling.
USB 2.0 Compliance Failure: TD 9.12 - Remote Wakeup Test (Configuration Index 0) - Disabled

```
INFO
Could not disable remote wake up on device
FAIL

(USB: 1.2.102) After a successful ClearFeature(DEVICE_REMOTE_WAKEUP) Bit 1 of status word returned in response to a GetStatus request must be set to a zero.
```

I've tracked this down in the Kinetis SDK to the `kUSB_DeviceControlGetDeviceStatus` handler in the USB Device driver being hard-coded to always return self powered and remote wakeup enabled bits. This is a big no-no -- we need to return the current remote wake state as set by the host, and only set the self-powered bit if the device is battery powered (which it is not).

abcminiuser:
Fixes #373.

*Note that this will not build as-is, and requires the following one-liner Kinetis K22 patch (which I am too time-poor to PR today, but can be manually applied if desired)*:

```
commit 3d73483096cc5a4387fa581486411ec01b1bb4ab (HEAD -> feature/add-device-get-status-event)
Author: Dean Camera <dean@fourwalledcubicle.com>
Date:   Sun May 23 10:56:16 2021 +1000

    Patch in a new device status callback event so the application layer can provide accurate information.

diff --git a/middleware/usb_1.0.0/device/usb_device.h b/middleware/usb_1.0.0/device/usb_device.h
index f60ac64..fff0622 100644
--- a/middleware/usb_1.0.0/device/usb_device.h
+++ b/middleware/usb_1.0.0/device/usb_device.h
@@ -110,6 +110,7 @@ typedef enum _usb_device_event
     kUSB_DeviceEventSetRemoteWakeup,            /*!< Enable or disable remote wakeup function. */
     kUSB_DeviceEventGetConfiguration,           /*!< Get current configuration index */
     kUSB_DeviceEventGetInterface,               /*!< Get current interface alternate setting value */
+    kUSB_DeviceEventGetDeviceStatus,            /*!< Get current device status flags (self powered, remote wake-up enabled) */
 } usb_device_event_t;

 /*! @brief Endpoint callback message structure */
```

This corrects the bafflingly hard-coded responses to `GetDeviceStatus` control requests from the host. This request is supposed to return the current device status flags including self/bus powered state and remote wakeup enable state, but right now just returns hard-coded `SELF_POWER` and `REMOTE_WAKE_ENABLED` flags.

This makes us fail USB compliance testing as the host is unable to see the remote wake feature being disabled after a `SetFeature` control request. That's mostly just a technical failure with no serious real-world implications, but it's worth getting right so we finally pass the compliance suite.

The other flag is more serious; we advertise that we are self powered all the time. We are an exclusively bus powered device with no internal or external means of power other than the host, and an over-zealous host may detect that we are drawing more than the 50mA budget allotted to self-powered devices and shut down the port. Most hosts are cheap and just use a 1A fuse per-port, but it's entirely possible that more correct implementations exist.

@mondalaci do you know the maximum real-world current draw of the keyboard with maximum brightness and all the LEDs on white? We will need to set that correctly in the USB descriptors -- but note that if it is more than ~150mA the keyboard may be rejected if plugged into unpowered USB hubs.


mondalaci:
Thanks for your contribution!

Please see the [power consumption](https://ultimatehackingkeyboard.com/knowledgebase?q=power) of various UHK versions.

abcminiuser:
Uh oh, 700mA is out of spec - the only hosts that can (legally) supply that are USB3.x ports, with special negotiation. For USB 2 ports, 500mA is the max current that can be negotiated between the host and device, with 100mA max draw being allowed before the configuration has been set.

I propose we merge in this fix so we accurately report the remote wake and power states to the host, even if we are still non-compliant for the power draw. We can then look into the power issue.

One immediate partial fix would be to specify 500mA max draw. This will prevent the keyboard from working in unpowered hub situations, but would be closer to the actual spec (especially since with the current code we would just draw wildly over spec current anyway).

We can then look into two areas to fix this up:
1) See if we can limit the overall max current to something close to 500mA. That would reduce the overall max brightness, but probably not by as much as you would imagine due to the non-linear response of the human eye
2) Potentially add a second configuration with a 100mA max current draw specified, so the keyboard can work in some limited fashion (no backlights?) when plugged into an unpowered hub.

mondalaci:
I actually meant to set the default brightness of the UHK so that it draws 500 mA and almost forgot it. Then users would still be able to max it out via Agent, but at least it'd be in spec by default.

In practice, a number of USB keyboards draw more than 500 mA. I talked with Jacob Alexander of the Input Club, who designed many keyboards, and he said that drawing above 500 mA is rarely an issue in practice.

As for the choice you mentioned, I'm unsure which solution would be the best. What do you suggest in light (pun clearly intended) of the above?

I'll test and merge your PR soon.

abcminiuser:
I've updated the review to include removal of the self-powered attribute and increase of the max current consumption to 500mA. That's the quickest way to make this closest to what is actually going on, as right now we effectively report a 50mA total draw to the host. You are correct that most hosts won't care and I am very aware of some commercial products that violate this - most hosts just use a 1A or so polyfuse for limiting without any smart monitoring.

If you want to try to support the use case of being inserted into an unpowered hub (without back-lighting at all) I can start doing some testing, although frankly I don't think it will be of much practical use. Doing so would require multiple alternative configuration descriptors which might decrease compatibility.

mondalaci:
Thanks for the update! I'd much prefer not to support the more sophisticated use case at the expense of reduced compatibility.

As for the exact power consumption, how about branching based on `DEVICE_ID` and reporting 130 mA for `DEVICE_ID_UHK60V1` and 500 mA for `DEVICE_ID_UHK60V2`?

abcminiuser:
Yep, fixed (off to bed now, so that was good timing). I used 50mA units since that's what hosts will compute the loads in, practically speaking.

It'll be worth re-testing #219 with all these compliance fixes, as I know various MacOS versions are finicky about USB remote wake-up for devices that don't implement all the standard and HID class requests properly.

PR #393, opened 2021-08-07
Fix overflow of the HID idle period, when a period above 16 (64ms) is set.

abcminiuser:
I did this originally, locally, but later changed it to `us` units to avoid the division and forgot to update the size of the intermediate types. As a result, this causes an arithmetic overflow on HID idle periods above 64ms, which is very likely.

If set, the actual idle period would be extremely uneven, and much faster than intended - so BIOS and other reduced feature hosts would see many key repeats rather than just one or two per second.


mondalaci:
Thank you for the fix!

PR #397, opened 2021-08-12
Extended macro engine.

kareltucek:
(once more, since the original branch caught flame...)

So here we go... There are some things to be polished later, but this should be the minimal version.

Some notes here - https://github.com/kareltucek/firmware/issues/61 - most notably - should the language version be actually represented in the firmware and queriable via usb? (I sort of assume so, but did not implement it so far...)


mondalaci:
1. When I hold Mod+D, it gets mapped to Alt+Tab, but Alt is not held.
2. Extended macros should be enabled by default, and consequently, --extendedMacros, CUSTOM_CFLAGS=-DEXTENDED_MACROS, and processStockCommandAction() should be removed.

I keep testing your PR. This feature is super exciting, so please keep it up, @kareltucek!

mondalaci:
Actually, I can't keep testing this PR until the Alt+Tab issue is fixed because I rely so heavily on it.

mondalaci:
Regarding smart macro versioning, please extend `protocolVersions` with `SMART_MACROS_{MAJOR,MINOR,PATCH}_VERSION` and let's start with version 1.0.0

`scripts/package.json` should also be extended with `smartMacrosVersion` and so should `generate-versions-h.js` accordingly.

kareltucek:
>     1. When I hold Mod+D, it gets mapped to Alt+Tab, but Alt is not held.

Fixed. Caused by a typo which broke the ActiveLayerHeld flag. Only one of the two switcher types was affected.

mondalaci:
Thanks for the fix! I'll be testing your PR for a while before merging it.

kareltucek:
Some more code implemented on separate branch, please have a look at it and let's merge it into this branch. (I would create the PR in this repo, but github does not allow me to create a PR here that goes between two branches living in my repository...)

https://github.com/kareltucek/firmware/pull/65

kareltucek:
Similarly with 12-layer support -> https://github.com/kareltucek/firmware/pull/66

kareltucek:
One more update - proper multiline support. Pushing directly into the PR branch since it deals with requirements of 351 directly, and does not touch original stock code much.

This was a bit troublesome affair, due to which I have after all refactored return values in entire macros.c. I also dropped the old `$` notation, so extended commands are now available only by UltimateHackingKeyboard/agent#1545 agent. This also allows me to easily rectify the scheduling related problems. But for the moment, I don't plan to clutter this PR any more. 

So atm., my suggestion is to review and merge kareltucek/firmware#65 and kareltucek/firmware#66 into this PR first, and then continue with testing and discussions of this PR.

mondalaci:
@kareltucek Please run `$onInit`, then `$onKeymapChange $currentKeymap` every time the configuration gets saved. The rationale is that many folks will be experimenting with various macro commands and parameters, and power cycling the UHK from time to time would be very cumbersome.

kareltucek:
@mondalaci done.

mondalaci:
@kareltucek Thanks for the quick commit, but only `$onInit` gets called. I'd also like `$onKeymapChange $currentKeymap` to get called afterward.

kareltucek:
@mondalaci what exactly do you mean by `$onKeymapChange $currentKeymap`? I take `$currentKeymap` to stand for a  keymap abbreviation, like `QTY`, in full, `$onKeymapChange QTY`. That I believe works correctly.

mondalaci:
I mean that when saving the configuration, I expect `$onInit` to be called followed by `$onKeymapChange QTY` assuming that QTY is the current keymap. The same should happen upon firmware startup.

kareltucek:
Retested. On save, both `$onInit` and `$onKeymapChange QTY` get executed. Their commands are interleaved, so if your two macros are overwriting the same value, result is either of the two.

On firmware startup, the same happens (both macros get executed), except any produced usb reports don't get sent, probably because USB stack is not initialized yet at that point? I assume that happens in the background via interrupt routines. I didn't pay much attention to this, but it is true that it should be probably fixed, even if only to avoid confusion. The update loop should probably not start until all interfaces are running properly. How do I determine the moment when everything is set up and working?

 If you are still having issues with the "on save" behaviour, would you share your UserConfig?

mondalaci:
I tried to overwrite the same navigation mode value in `$onInit` and `$onKeymapChange QTY`, so that must be the culprit. I strongly feel that this use case should work, and I expect users to specify general settings in `$onInit` and keymap-specific settings in `$onKeymapChange $keymapName`, so the latter should be fully executed after the former, and the two macro events shouldn't interfere.

I agree that the USB stack is probably uninitialized at startup. There must be some variables, or better yet, callbacks that could be used to detect the initialization of the stack, but I don't know which ones, and I don't have time to look into this. In any case, this is not so important as the first issue.

kareltucek:
Well... there is the other macro executor on https://github.com/kareltucek/firmware/tree/common_trunk_upstream (a branch which contains the other PRs as already merged and some more functionality that has accumulated since this PR was opened). Currently, it is considered experimental feature enabled by `set macroEngine.scheduler blocking`. So the proposal is to merge `common_trunk_upstream` directly and make that other executor default. It will need some little changes, but should fit the bill even without ugly hacking. In that case, execution order will be preserved as long as the `$onInit` macro event does not enter any waiting/sleeping state (that is, `delay` or `holdKey`, `holdLayer` and similar commands). 

Otherwise I am open to suggestions.

kareltucek:
Or I can alter current execution strategy to execute as many commands as possible in each cycle. Not sure how good idea that is however.

mondalaci:
What's the practical drawback of altering the current execution strategy?

In any case, after merging this PR, I'll make a release before merging additional PRs since it's already huge, and I want folks to be able to test the individual releases when encountering a bug.

kareltucek:
Hard to tell. Except touchpad-related problems (which are targeted by one of those additional PRs), nothing serious I think. One drawback may be that if multiple macros are running, they will tend to fall in sync on output operations. Also, If anyone relies on stepping logic, they might be surprised. But those are corner cases that probably don't have practical relevance.

In the end the biggest practical drawback I can foresee is that the further your codebase is going to be from my own setup, the more troublesome may providing support be. (But this relates much more to the overall decision to go step-by-step than to the execution strategy itself.)

kareltucek:
There you go.

mondalaci:
I've been sitting on this PR for ages, and it's time to merge it. Let me give some feedback.

First up, let's not merge any more branches into this one, unless necessary because this PR is already quite overwhelming. Please resolve the conflict in this branch, though.

I haven't encountered any touchpad-related bugs. In case there are any known bugs that cause practical issues, please let me know how to reproduce them.

I'd much rather only have a minimal set of commands for the 1.0.0 version of smart macros, namely only `set` and `printStatus`. Please put all other commands into ifdefs, so you can enable them in your fork. I'd like to review the inactive commands one by one later.

I'd expect the error buffer to get flushed when fixing the macro commands in macro events and saving the configuration, so `printStatus`shouldn't output anything in such cases.

As for `set`, the bodies of some parameters are unimplemented, so the firmware doesn't error out when encountering them. I strongly feel that every unimplemented parameter should throw errors, so please don't only uncomment the bodies of unused parameters, but their branching, too.

`set mouseKeys.move.*` values should reflect the values of the Mouse speed page, so specified values should be transformed accordingly.

I strongly prefer to only support the `#` macro syntax. The `//` syntax adds redundancy.

I'd much prefer to use `true` vs `false` for boolean `set` values instead of `1` and `0` because it makes their boolean nature apparent.

I'd prefer using `diagonalSpeedCompensation` instead of `compensateDiagonalSpeed`. As a rule of thumb, using nouns instead of verbs feel right for `set` variable names.

I prefer `stickyModifiers` over `stickyMods` because it's more explicit and one can't mistake this command for the Mod layer this way. The values should be `never` instead of `0` and `always` instead of `1`. But even more importantly, what's the practical significance of this setting? I'd only use the `smart` option and can't see why the other values are practical.

Do you know anyone who ever benefitted from adjusting `debounceDelay`? I'm afraid that some folks will be messing with sane factory defaults causing additional support burden.

What's the practical use of `set chording`?

In which pratical cases can `set emergencyKey` be useful?

Can you easily deactivate conditions, modifiers and labels for the time being. It'd keep smart macros 1.0.0 minimal?

kareltucek:
Allright:

- [x] - revert the revert of extended macros parameter.
- [x] - clear error buffer when configuration is written
- [x] - fill missing unimplemented throws.
- [x] - fix `mouseKeys.move.*` multiplier

> I strongly prefer to only support the `#` macro syntax. The `//` syntax adds redundancy.

`#` syntax is nice, but ambiguous. The language will probably need a proper tokenizer at some point, which will among other things deal with this ambiguity. I would prefer not to discuss it here and now.

- [x] - extend boolean parsing to also accept `true` and `false`
- [x] - rename  `compensateDiagonalSpeed` to `diagonalSpeedCompensation`.
- [x] - rename `stickyMods` to `stickyModifiers`.
 
> But even more importantly, what's the practical significance of this setting? I'd only use the `smart` option and can't see why the other values are practical.

I can't see a setting which makes "random" keys stuck on purpose as practical. The significance is that I am not using these keybindings and want to be able to disable a feature which is going to make my keyboard misbehave "at random".

> Do you know anyone who ever benefitted from adjusting `debounceDelay`?

Yes. Me and p4elkin, for development purposes.

> What's the practical use of `set chording`?

If I want to be able to strike complex chords as a single motion. 

E.g., without having to pay attention to correct order and without having to mentally split the motion into multiple phases.

> In which practical cases can `set emergencyKey` be useful?

Again, debug and development. Postponer-related features may cause the keyboard to lock up in case of either bug, or certain (although by definition correct) macro sequences. `emergencyKey` allows both debug and recovery from such conditions.

> Can you easily deactivate conditions, modifiers and labels for the time being. It'd keep smart macros 1.0.0 minimal?

Yes, I can.

kareltucek:
Done. Still, please note that I do not consider the decision to release several months old and deprecated code as an unquestionably good one.

mondalaci:
Thanks for the quick changes!

The 0 and 1 values are still allowed as booleans. I'd like to restrict the use of false and true as the only accepted boolean values. I know they're a mouthful compared to 0 and 1, but they're more explicit, and users wouldn't mistake them for integer values.

Please give me some practical examples regarding the `chording` variable.

> Done. Still, please note that I do not consider the decision to release several months old and deprecated code as an unquestionably good one.

What do you mean? Would you merge more commits into this PR to make it more up-to-date? I'd much rather merge this huge PR first and then merge separate PRs one by one afterward.

kareltucek:
> The 0 and 1 values are still allowed as booleans. I'd like to restrict the use of false and true as the only accepted boolean values. I know they're a mouthful compared to 0 and 1, but they're more explicit, and users wouldn't mistake them for integer values.

Yes they are, and there is a reason for it - maintaining integer compatibility provides regular integer features - for instance substituting register values.

> Please give me some practical examples regarding the `chording` variable.

If I hit `A+Ctrl` instead of `Ctrl+A`, UHK will still send `Ctrl+A`.
 
> > Done. Still, please note that I do not consider the decision to release several months old and deprecated code as an unquestionably good one.
> 
> What do you mean? Would you merge more commits into this PR to make it more up-to-date? I'd much rather merge this huge PR first and then merge separate PRs one by one afterward.

Those other three PRs are in fact open just for your convenience - so that you can read them and review them separately from other things if you wish. They are already merged into `common_trunk_upstream`, together with some other new features and refactors, which have been accumulating for some time already. As a consequence, from my point of view, this code is already quite old and partially irrelevant.

So yes, alternative is to jump straight to `common_trunk_upstream` (which - just as you say - contains more commits and is more up-to-date). I admit I am not entirelly sure which way to go - code in this PR is more stable, but since you want (non-set) commands disabled anyway, this (only) benefit does not apply. Apart from that, I can see only downsides to releasing outdated code.

mondalaci:
I'm not proposing getting rid of integer values. I'm proposing not using integers for booleans. Any objections?

I'll merge the rest of the PRs separately. I only plan to make a release afterward.

mondalaci:
As for the `chording` option, I'm sure that 50 ms is a reasonable default, but it may be practical to make this value configurable. Also, `chordingDelay` may be a more explicit name.

kareltucek:
> As for the `chording` option, I'm sure that 50 ms is a reasonable default, but it may be practical to make this value configurable. Also, `chordingDelay` may be a more explicit name.

Both good suggestions.

> I'm not proposing getting rid of integer values. I'm proposing not using integers for booleans. Any objections?

Technically taken still the same one - features provided by the integer subsystem which do benefit boolean settings this way too. Although to be fair, practical concerns are not very grave and the most proper solution is probably to introduce proper variable (register) typing (which then could also work with things like floats, keymap abbreviations, and scancode actions).

More importantly, I have realized that I actually do like the brevity of `1|0`, and do not think that folks are going to have problems with this specific thing. (I mean - sure, smart macros do carry some support burden, but I think folks are going to be confused about actual features rather than about messing up integers with booleans.)

mondalaci:
I'm fine with using `0` and `1` for booleans, but then please remove `false` and `true` for the sake of minimizing redundancy.

The speed settings of the key cluster module don't seem to affect its mini trackball except for the acceleration setting, which makes the mini trackball less responsive when set to 10, for example.

For the trackball, acceleration=1 works very well, but other values make me feel that the pointer follows my trackball movements poorly.

For the trackpoint, acceleration values such as 1.1 or 3 make pointer movements jaggy.

For the touchpad acceleration values, my experience is similar to the trackball.

I think the acceleration driver is a blessing because it allows for smooth movement at low speeds, but it can run wild quite early when experimenting with various values. Maybe the acceleration values should be limited to a range, to certain values, or for some modules, not even exposed, or else users will get confused and file bug reports. Interested in your opinion.

kareltucek:
> I'm fine with using `0` and `1` for booleans, but then please remove `false` and `true` for the sake of minimizing redundancy.

Allright - is it fine like this, or do you have something else in mind?

> The speed settings of the key cluster module don't seem to affect its mini trackball except for the acceleration setting, which makes the mini trackball less responsive when set to 10, for example.

Yes. This branch still contains your hardcoded speed settings (which is 5 for cursor, 1 otherwise). My reworked handling waits in https://github.com/kareltucek/firmware/pull/65 .

> For the trackpoint, acceleration values such as 1.1 or 3 make pointer movements jaggy.

I see. The pow does some crazy things to performance with exponent values greater than one. I have thrown in a fast approximate computation instead...

> I think the acceleration driver is a blessing because it allows for smooth movement at low speeds, but it can run wild quite early when experimenting with various values. Maybe the acceleration values should be limited to a range, to certain values, or for some modules, not even exposed, or else users will get confused and file bug reports. Interested in your opinion.

I think we should wait whether or not people really get confused, and if they do, then see why. Then fix docs. Limiting the acceleration range might make sense, but I am afraid that just limiting it will not improve user experience a bit - it will cause the same non-intuitive effects (and consequent headaches) even within reasonable bounds.

In any case, recommended ranges should be clearly indicated in documentation (preferably worded as "valid" rather than "recommended").

(Btw., I indicate 0-1 as recommended range for the acceleration exponent.)

> For the trackball, acceleration=1 works very well, but other values make me feel that the pointer follows my trackball movements poorly.
> 
> For the touchpad acceleration values, my experience is similar to the trackball.

This is very nondescript - If you have been trying acceleration values greater than 1, bad experience is expected since (1) such values are crazy high (2) the above pow performance issue.

As a rule of thumb, Acceleration exponent controls shape of the curve (i.e., low speed to high speed ratio), but does not serve to increase pointer speed. (I suspect that this misunderstanding will be a common one.) The curve sort of pivots around midSpeed (which currently is defined in code only) - increasing acceleration will increase the highest speeds and decrease the lowest ones, but will not really affect overall perception of speed/responsiveness.

Speed should be used to tune either the accelerated speed or the overall speed. 

When Acceleration and Speed are tuned, BaseSpeed should be used to finetune the near-zero speed multiplier.

(Difference between BaseSpeed and Speed is that Speed is affected by Acceleration, while BaseSpeed is not. Therefore, nonzero BaseSpeed can be used as a counterweight to high Acceleration values.)

mondalaci:
> > I'm fine with using `0` and `1` for booleans, but then please remove `false` and `true` for the sake of minimizing redundancy.
> 
> Allright - is it fine like this, or do you have something else in mind?

Currently, `false` and `true` are supported as valid boolean values in this branch. Please remove them, and only leave `0` and `1` as valid boolean values.

> As a rule of thumb, Acceleration exponent controls shape of the curve (i.e., low speed to high speed ratio), but does not serve to increase pointer speed. (I suspect that this misunderstanding will be a common one.) The curve sort of pivots around midSpeed (which currently is defined in code only) - increasing acceleration will increase the highest speeds and decrease the lowest ones, but will not really affect overall perception of speed/responsiveness.

For this very reason, this setting shouldn't be named acceleration. We should come up with a more intuitive name. The way I see it, the primary role of this feature is to allow precise movement at lower speed while allowing fast movement at higher speed. The term "precision smoothing" is the best that comes to my mind. If you don't have a better alternative, please rename `acceleration` to `precisionSmoothing` in smart macro names and in the source. I agree that we should advise the [0, 1] interval as legit values for this setting.

kareltucek:
> > > I'm fine with using `0` and `1` for booleans, but then please remove `false` and `true` for the sake of minimizing redundancy.
> > 
> > 
> > Allright - is it fine like this, or do you have something else in mind?
> 
> Currently, `false` and `true` are supported as valid boolean values in this branch. 

No, they are not - now `0` and `1` are documented as valid booleans and other integers are accepted.

> The term "precision smoothing" is the best that comes to my mind. If you don't have a better alternative, please rename `acceleration` to `precisionSmoothing` in smart macro names and in the source. I agree that we should advise the [0, 1] interval as legit values for this setting.

I am not convinced. I do not have a perfect name, but acceleration still describes the resulting effect of increasing speed gradient pretty well, and in linux the very same mouse settings is also called acceleration. (And in linux, it features the same non-intuitiveness as well.)

Maybe `accelerationExponent`? But then other values should probably be changed to `speedMultiplier` and `baseSpeedMultiplier`.

kareltucek:
Or maybe `accelerationExponent`, `accelerationMultiplier` and `baseSpeedMultiplier`?

//Edit: not a very good one, since it does not cover that `accelerationMultiplier` will still affect speed even if accelerationExponent is set to zero.

kareltucek:
Or we could switch to an actual exponential. That would remove some degree of freedom, but would make configuration a bit more intuitive.

mondalaci:
We shouldn't append `Multiplier` to the names of `speed` and `baseSpeed`. The actual variable names are simpler and describe the role of these variables well.

In the same vein, I don't think it's a good idea to change `acceleration` to `accelerationExponent`. We shouldn't bother users with math terms and make variable names as intuitive as possible. My main gripe with the `acceleration` name is that it affects both acceleration and deceleration. Actually, I can notice deceleration much more in the case of the trackpoint, and it helps greatly to hit small targets. Let's think about a better name a bit more.

Please only accept `0` and `1` for boolean values and throw an error is if any other value is specified.

I think the acceleration feature works well in the [0, 1] interval and shouldn't be modified.

mondalaci:
I've just made up `xceleration` hinting at both deceleration and acceleration.

kareltucek:
> Please only accept `0` and `1` for boolean values and throw an error is if any other value is specified.

Ok.

> I've just made up `xceleration` hinting at both deceleration and acceleration.

I like this idea.

---------------

Done

mondalaci:
Looks like you only switched to `xceleration` in the docs, but not in the code.

kareltucek:
> Looks like you only switched to `xceleration` in the docs, but not in the code.

Indeed. Fixed.

mhantsch:
If you want something more user understandable, how about naming `xceleration` -> `speedboost` or `speedfactor` ? (just tossing in my ideas.. ignore me if you don't like it)

mondalaci:
@mhantsch Thanks for the suggestion. `speedboost` suggests acceleration, even though deceleration is also involved, and `speedfactor` is quite vague. I'd rather stick to `xceleration`, which hints at both acceleration and deceleration.

kareltucek:
@mondalaci btw, I do agree that one single commit looks nice in the uhk/master repository for PRs like this one, but do you realize that this way the commits are probably going to haunt you till the end of the days in all subsequent PRs?

mondalaci:
@kareltucek I wasn't even conscious that I merged this PR as a single commit. It's probably not a good idea, although I'm unsure of the best practice in various cases.

PR #407, opened 2021-09-15
Fix hardware config allignment.

### LINKED ISSUE #406: Wrong detection of ISO vs ANSI
@kareltucek Just committed aebe2080b135c38a198adb00fbf579dd7fa91be9, but always ISO gets detected, even after running `agent/packages/usb/write-hardware-config.ts uhk60v2 ansi` and power-cycling the UHK. Any ideas what's wrong?

kareltucek:
@mondalaci alignment.

Closes #406.

### DIFF
--- a/right/src/config_parser/config_globals.h
+++ b/right/src/config_parser/config_globals.h
@@ -4,6 +4,7 @@
 // Includes:
 
     #include "fsl_common.h"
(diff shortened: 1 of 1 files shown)


mondalaci:
Great catch, and thanks so much for the quick fix!

PR #419, opened 2021-11-12
Trackpoint jumps fix

### LINKED ISSUE #383: Trackpoint module pointer jumping
When the trackpoint module is used, the pointer sometimes jumps. That is, its position suddenly changes. According to my experience, the relative distance of the jump is constant. I think the PS/2 protocol parser of the trackpoint is buggy which results in this issue.

kareltucek:
Closes #383 .

### DIFF
--- a/right/src/mouse_controller.c
+++ b/right/src/mouse_controller.c
@@ -387,6 +387,7 @@ void MouseController_ProcessMouseActions()
     MouseScrollState.yOut = 0;
 
     if (Slaves[SlaveId_RightTouchpad].isConnected) {
+        // TODO: this is still unsafe w.r.t interrupts
         processTouchpadActions();
         processModuleActions(ModuleId_TouchpadRight, (int16_t)TouchpadEvents.x, (int16_t)TouchpadEvents.y);
         TouchpadEvents.x = 0;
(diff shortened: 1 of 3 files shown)


right/src/mouse_controller.c
```diff
@@ -399,9 +400,17 @@ void MouseController_ProcessMouseActions()
             continue;
         }
 
-        processModuleActions(moduleState->moduleId, (int16_t)moduleState->pointerDelta.x, (int16_t)moduleState->pointerDelta.y);
+        __disable_irq();
```
kareltucek, line 403: This place is important.

shared/module/slave_protocol_handler.c
```diff
@@ -99,10 +99,16 @@ void SlaveTxHandler(void)
             uint8_t messageLength = BOOL_BYTES_TO_BITS_COUNT(MODULE_KEY_COUNT);
             if (MODULE_POINTER_COUNT) {
                 pointer_delta_t *pointerDelta = (pointer_delta_t*)(TxMessage.data + messageLength);
+                __disable_irq();
```
kareltucek, line 102: And this place is important.

trackpoint/src/module.c
```diff
@@ -21,22 +21,19 @@ void Module_Init(void)
                       &(port_pin_config_t){/*.pullSelect=kPORT_PullDown,*/ .mux=kPORT_MuxAsGpio});
     GPIO_PinInit(PS2_CLOCK_GPIO, PS2_CLOCK_PIN, &(gpio_pin_config_t){.pinDirection=kGPIO_DigitalInput, .outputLogic=0});
 
-    PORT_SetPinInterruptConfig(PS2_CLOCK_PORT, PS2_CLOCK_PIN, kPORT_InterruptEitherEdge);
+    PORT_SetPinInterruptConfig(PS2_CLOCK_PORT, PS2_CLOCK_PIN, kPORT_InterruptFallingEdge);
```
kareltucek, line 24: All following changes (all in the `module.c`) are not so important:

- They are great deal aesthetic changes.
- The PS2 interrupt will now be called only on falling edge. This means lower probability of timing issues as there is fewer calls overall. (All related changes should be aesthetic only.)
- It should now be more robust and be able to deal even with severe problems with either timing or garbled GPIO read values.

During testing, it seems that some evidence of timing-induced race conditions was encountered - i.e., some PS2 interrupts arriving late - but the evidence was not very conclusive and fixing the other race condition (the one on 16-bit assignments) seems to have fixed all the problems in practice, so while changes in `module.c` might have some effect on overall robustness, they are by no means mandatory part of the PR. (I.e., feel free to pick whichever set of the changes seems reasonable.)

mondalaci:
@kareltucek Great job on this one! This issue was highly nontrivial, and your persistence is much appreciated.

PR #420, opened 2021-11-16
Common trunk upstream

kareltucek:


mondalaci:
@kareltucek This PR contains so many changes, I'm getting quite lost. In any case, I'm eager to merge it and catch up with your repo. Would you please write a summary of this PR?

kareltucek:
Relevant changes:
- more layers feature support
- bug fixes (esp. related to stability and correctness of the macro engine in context of the new macro-sleeping performance optimizations - macro engine now should not eat performance on "hold"/"delay" features)
- new elegant way to queue/chain macros (relevant to the onInit issues)
- some timeouts were changed (performance optimizations made the internal clock run somewhat faster, resulting in doubletaps not being correctly registered)
- some bugs fixed relating to layer switching and to sticky modifiers 

Semi-relevant features:
- rebinding both keys and navigation mode slot actions (only to macros)
- adding monochromatic LED backlight mode (via `set` command)
- doubletapDelay can now be set via smart macros

Other features:
- `tapKeySeq` command that allows specifying more complex scancode sequences easily (codename "emoji support")
- `ifKeymap` and `ifLayer` conditions and `ifShortcut/ifGesture orGate` option  (thanks to @soraxas)
- `fork` command added, which allows spawning new macro without blocking
- the blocking scheduler is now default 
- The blocking scheduler now uses a circular linked list to keep track of order of execution - this allows more deterministic behaviour when spawning/calling new macros. 
- decreased macro state pool size to 16.
- `activateKeyPostponed` can now trigger actions from other layers

(I assume that the commit history is not relevant due to reasons discussed in `common_trunk` PR. Since then, I have emloyed new merging scheme thanks to which I hope I should be able to deal with squash merges without any more problems.)

For some reason, I see the mouse-related functionality shown in current PR as new, although it should already be merged and therefore same on both sides. 

Also, some changes from past PRs seem to be reverted - this is **not intentional from my side and needs further investigation** -> please do not merge this yet.

My todos:
- [x] check that semver versioning was bumped whenever it should have been 
- [x] check health of the PR (relating to above note of unexpected things showing in the diff)
- [x] check that changing led `rgb` also switches driver automatically

Todos for @mondalaci :
- [x] please resolve https://github.com/UltimateHackingKeyboard/firmware/issues/438

mondalaci:
Thanks for the detailed list! I'm very impressed, and I'd love to merge this as soon as possible, but the configuration transformation in Agent for the "more layers" feature will be time-consuming to get right. If you can remove the "more layers" feature from this PR with reasonable effort, please do it because it can block the merging of this PR for a while.

kareltucek:
>  If you can remove the "more layers" feature from this PR with reasonable effort, please do it because it can block the merging of this PR for a while.

You don't have to enable it. The firmware is backwards compatible with the old config - it is up to you when you decide to bump the version.

mondalaci:
Maybe I'm mistaken then. So does this PR nothing have to do with https://github.com/UltimateHackingKeyboard/agent/issues/617, and is the latter implemented by #422 on the firmware side?

kareltucek:
#422 is a courtesy / pro-forma PR. It's purpose is to allow you to review code changes which I made to the code which was originally written by you and other contributors. (Contrary to this PR, which is mostly my changes to my own code.)

#422 is fully included in this PR and implements UltimateHackingKeyboard/agent#617. However, it does so in a backward compatible manner. 

Maybe I'm mistaken, but I believe that config user version is actually included within the flashed user config. This gives the firmware ability to switch depending on the version of config which is actually written in the ROM (rather than which is expected to be written). In other words, if Agent writes a 5.x user config and correctly tags it as a 5.x version user config, the firmware will read it correctly as a 5.x config. If the Agent writes a 4.x user config and correctly tags it as 4.x, the firmware will read it correctly as a 4.x config.

In other words, if you release the firmware with 5.x, it will act as a 5.x version. If you release it as 4.x, it will act as 4.x. (More precisely, it will be treated as the corresponding version by Agent.)

(Also please note that this is exactly the reason why I have opened the #422 PR - because it is important that you know how I implemented the version transition and can point out any misunderstanding about the version transition process.)

mondalaci:
Thanks for separately submitting #422! It's much appreciated, I'll review it and give it a try as soon as its conflicts are resolved.

I planned to modify the configuration parser for user config 5.0.0 in a non-backward compatible manner to minimize the amount of legacy code, but if backward compatibility doesn't add much cruft, then your implementation is preferred.

Does this PR contain (almost) all the code of your fork? It'd be nice to catch up and only receive smaller PRs from that point forward.

kareltucek:
> Does this PR contain (almost) all the code of your fork? It'd be nice to catch up and only receive smaller PRs from that point forward.

Yes, it is a long-living branch which is intended to contain the most up-to-date code.

kareltucek:
Conflicts resolved. As for the health, I have not found any real problem.

PR #421, opened 2021-11-16
Mouse controller refactors and extensions

kareltucek:


mondalaci:
First things first, I'm officially blown away by these configuration options, and I'm sure our customers will be, too. Keep on rocking! Let me provide some insights.

`zoomSpeedDivisor` should behave just like `scrollSpeedDivisor`, but for the zoom navigation mode for the sake of consistency. 

I suggest adding `module.touchpad.pinchToZoomSensitivity` for setting touchpad-specific zoom sensitivity.

I don't think divisors are intuitive. Let's rather use multipliers exclusively. This way, people don't have to think about inversely proportional relationships, only proportional relationships. As for the related variable names, I wouldn't feature "multiplier" instead of "divisor" for the sake of brevity. Accordingly, let's rename `caretSpeedDivisor` to `caretSpeed`, `scrollSpeedDivisor` to `scrollSpeed`, and `zoomSpeedDivisor` to `zoomSpeed`, and let's make their default values 1/defaultValue of their current value.

It'd be worth introducing `mediaSpeed` to complement `caretSpeed`, `scrollSpeed` and `zoomSpeed`.

`*AxisLockEnabled` should be renamed to `*AxisLock` for brevity's sake.

`cursorAxisLock` should be removed as it doesn't make practical sense, but `caretAxisLock` does make sense, so it should be added.

`scrollAxisLock` is very useful, so it should be enabled by default.

Axis locking worked for me very reliably regardless of the value of `axisLockStrength`. How about you? This variable may not be needed and a reasonable default suffices.

As for `axisLockStrengthFirstTick`, I don't know what "tick" means, and I don't understand the use case of this variable. Please give me an example.

`swapAxes` only makes practical sense for the key cluster, so an error should be thrown when using this setting with other modules. When using `swapAxes` with the key cluster, horizontal scrolling works backwards than what is intuitive for me, so I think it'd be worth to reverse it.

The Ctrl- and Ctrl+ don't work on Mac on which Cmd- and Cmd+ are the relevant shortcuts. First, I wanted to introduce a pc|mac smart macro variable for this purpose, but it'd be a better solution to send both Ctrl- and Cmd- versus Ctrl+ and Cmd+ respectively. Based on my testing, the extra shortcuts don't trigger unwanted behavior in any platform. If they will, we can still introduce a pc|mac variable.

kareltucek:
> I don't think divisors are intuitive.

Let's see:

- Assume `scrollSpeedDivisor = 16`. Interpretation is simply "It takes covering 16 pixels to produce one scroll event."

- Assume `scrollSpeedMultiplier = 0.0625`. What is your interpretation of that? I really don't think that human brain is good at dealing with numbers that inherently begin with two leading zeroes.

> This way, people don't have to think about inversely proportional relationships, only proportional relationships.

Yes, that's exactly the reason why I introduce divisors and not multipliers.

-----

> I suggest adding module.touchpad.pinchToZoomSensitivity for setting touchpad-specific zoom sensitivity.

That's exactly the intended meaning of `zoomSpeedDivisor`, except I have packed it the way I have packed it in order to avoid introducing unnecessary special-case branchings into the code which would be going to cause confusion an bugs.

Also, other ways of packing it would/will void regularity of the grammar.

-----------

> It'd be worth introducing mediaSpeed to complement caretSpeed, scrollSpeed and zoomSpeed.

You are actually suggesting expanding `caretSpeed` into three new config values. What is the practical benefit of doing such a thing? Any caret binding which I can think of needs sensitivity which will allow control up to a single event, which means that they will need the same coefficient anyway.

I am probably rather going to figure out how remove the offending options since they clearly cause misunderstandings...

-------

> The Ctrl- and Ctrl+ don't work on Mac on which Cmd- and Cmd+ are the relevant shortcuts. First, I wanted to introduce a pc|mac smart macro variable for this purpose, but it'd be a better solution to send both Ctrl- and Cmd- versus Ctrl+ and Cmd+ respectively. Based on my testing, the extra shortcuts don't trigger unwanted behavior in any platform. If they will, we can still introduce a pc|mac variable.

If you mean producing `Ctrl + Alt + +/-`, that does not work on linux.

If you mean to send first one and then the another, how exactly do you propose to send the two events? Let me remind you that your standard way of encoding actions has no support for macro-like trains of actions.

kareltucek:
Let me be clear on one thing: from design, zoom navigation mode was intended to serve the touchpad pinch gesture *only*, and to allow implementation of the pinch gesture using already available features. The documentation (which you have clearly not read) explains that.

Ability to reference the mode and change its properties, rebind it, etc., is from my side meant just as "When we have the options, lets interface them (just as they are) rather than to try to hide them.". I.e., to interface them without introducing any more complexity.

What you are suggesting is taking it and expanding it into a new set of features, which I see as complex, unnecessary and awkward to implement and maintain.

mondalaci:
My comment regarding proportional versus inversely proportional relationships still stands. When increasing non-divisor variables like `module.trackpoint.speed`, the higher the value, the faster the movement. But when increasing divisor values, the higher the value, the slower the movement.

In practice, nobody would use values like 0.0625. Users would rather use values like 0.1, perhaps 0.06, and numerous other smart macro variables also use fractional numbers.

---

Anyone who's aware of all the different navigational modes, including `zoom`, will potentially confuse `module.touchpad.zoomSpeed` as a general module setting. The `set module.MODULEID.{... | zoomSpeedDivisor | ...}` part in MACRO_DOCS.md further contributes to this confusion.

Module-specific values, such as the pinch to zoom value should be explicitly marked as such. `swapAxes` should also be module-specific according to my previous comment, and I'm sure there will be other module-specific values, too. A mechanism is needed to handle module-specific versus general module variables at the code and documentation level.

The way I see it, it'd be practical to configure `module.MODULEID.zoomSpeed`, just as all the other navigation modes, and use a separate module-specific variable for pinch-to-zoom.

---

> You are actually suggesting expanding `caretSpeed` into three new config values.

I don't see how. If we remove the `Divisor` part of the related variables, we'll end up with `scrollSpeed`, `mediaSpeed`, and `zoomSpeed`. `caretSpeed` would be the fourth in the line.

---

As for the zoom shortcuts, I meant sending one, then sending the other. I'm unsure what you mean by "the standard way of encoding actions". It's not like binding a shortcut to a key. This is a special case that could be separately implemented for the zoom navigation mode.

kareltucek:
> My comment regarding proportional versus inversely proportional relationships still stands. 

My arguments against it still stand too.

> In practice, nobody would use values like 0.0625. Users would rather use values like 0.1, perhaps 0.06.

Still exactly my point. How do you intuitively explain to me how fast a `0.06` is? Is it really obvious and intuitive that a step between 0.1 and 0.05 is the same size as the step between 0.02 and 0.01? Shouldn't the first one be 5x more pronounced?

Also, operation of the mini trackball is discreet in nature, so if you want it to behave deterministically (without pseudorandom skips), you actually need to feed it exact fractions (like 0.0625 rather than 0.06). 

Furthermore this will make people mix the values up with the actual `speed` and `baseSpeed` which have somewhat different meaning.

----

> > You are actually suggesting expanding `caretSpeed` into three new config values.
> 
> I don't see how. If we remove the `Divisor` part of the related variables, we'll end up with `scrollSpeed`, `mediaSpeed`, and `zoomSpeed`. `caretSpeed` would be the fourth in the line.

Internally, there are just three modes. Cursor, scroll and "discrete action" mode. The last one can be bound to arbitrary discrete actions - at the moment three bindings of the last mode exist (caret, media, zoom).

I still don't see any practical reason why would anyone want to use different values for different bindings of the mode. Btw., this claim takes just one good counter-example to tear down.

I think it would be much more productive to try to figure out some way to distinguish between "the caret" mode and set of all the three "action" modes (which is a problem no matter what the resolution of this one point is - see bellow).

-----------

> As for the zoom shortcuts, I meant sending one, then sending the other. I'm unsure what you mean by "the standard way of encoding actions". It's not like binding a shortcut to a key. This is a special case that could be separately implemented for the zoom navigation mode.

I am saying that I do not know how to describe the thing with an instance of `key_action_t`.

> It's not like binding a shortcut to a key.

Currently, it is exactly like that. Performing actions is usb_report_updater's domain. Interpreting mouse events is mouse_controller's domain. Anything can be bound in navigation modes. Then, mouse_controller takes care of transforming the coordinate signal into regular usb_report_updater input, and then feeds it into the usb_report_updater, which executes it including features like the "extra usb reports". Currently, pinch zoom actions even can be changed from userspace.

It surely *can* be implemented the way you suggest, and if you take special care, you might even get it right without glitches (like the missing extra usb report), but I don't like the idea because it means slowly transforming the code into a minefield.

-----

> As for axisLockStrengthFirstTick, I don't know what "tick" means, and I don't understand the use case of this variable. Please give me an example.

- `tick` means an output event. (Or more precisely one trigger of an action which is bound to a navigation mode.) (*Edit: On semantics - I use it in context when discrete actions/events are triggered by accumulation of some continuous (aka non-discreet) signal.*)
- `first` means the first tick, which happens to be the one tick which is evaluated while an axis is not locked yet. Also, first tick after any 500ms of inactivity.

Usecase - if you want mini-trackball to require two "pushes" (roll by two units) to produce the first event, yet then to register 1:1. Or on the contrary, if you want first event to register with a very slight push of the big trackball and just then start counting the "gap" given by `scollSpeedDivisor`.

-----

> cursorAxisLock should be removed as it doesn't make practical sense,

Of course it makes sense - testing purposes (both for sake of development and for folks to be able to try out what exactly different values of the axis locks do), and maybe some uses of mini-trackball.

-----------------

> but caretAxisLock does make sense, so it should be added.

Interesting point. Axis lock is always on with caret modes (caret, media, zoom), and "disabling" it is same as setting both skews at 1.0, but for sake of intuitiveness, it should indeed be overloaded.

------------------------

> ...we'll end up with scrollSpeed, mediaSpeed, and zoomSpeed. caretSpeed would be the fourth in the line.
>
> but caretAxisLock does make sense, so it should be added.

These two patterns are becoming dangerously similar. I hope it does not mean that every navigation mode should have its own `axisLock` and a pair of `axisLockStrength` variables. If I count correctly, it would mean 12 new fields per module.

mondalaci:
Instead of divisors, we could use percentage values like `scrollSpeedPercent`. It'd allow for proportional behavior like for other variables.

The mini trackball of the key cluster module emits merely 8 pulses over 360 degrees, so it's very low resolution. Consequently, I don't think we should provide a divisor value for it.

---

It'd probably be a better idea not to expose `caret`, `media`, and `zoom` navigation modes, but only the discreet mode. I'd name it `custom` mode and make `up`, `down`, `left`, and `right` directions configurable as arbitrary key actions.

I do think it has merit to configure the sensitivity of the custom mode on a per-module and per-layer basis. There are very different use cases, such as cycling through workspaces, windows, characters (caret), adjusting volume (media), and a hardcoded solution won't be satisfactory for all these cases.

---

As for the double (PC and Mac) zoom shortcuts, I wasn't aware of their implementation challenges, but I strongly favor them because they just work.

---

`axisLockStrengthFirstTick` is perfectly suited for the mini trackball of the key cluster module. Some users reported accidentally activating the mini trackball, so it'd definitely be useful.

However, I can't see this option being useful for any other module. I think it should be made specific to the mini trackball of the key cluster.

---

We shouldn't expose developer options like `cursorAxisLock` via smart macros along with other options. I really can't think of any practical uses of this option, and we shouldn't burden users with it. If users want to experiment with axis locking, we should enable them via navigation modes for which it's useful, such as `scrollAxisLock`.

---

How about ditching axis lock enabling and only exposing axis lock strength? The way I see it, it'd reduce redundancy and allow for the same configuration possibilities.

mondalaci:
@kareltucek I'm eager to merge this PR after resolving the standing issues. Is this your focus nowadays?

kareltucek:
> Is this your focus nowadays?

Not at the moment.

mondalaci:
@kareltucek Would you resolve the conflicts? I want to merge this PR and release a new firmware version because it'd already provide great value to our customers. We'll keep tweaking and adding macro commands.

mondalaci:
I could merge this PR via the command line.

PR #434, opened 2021-12-23
Sign CLA

soraxas:
I have read the Agreement, and fully agree to it by signing it with my GitHub username.


soraxas:
For https://github.com/kareltucek/firmware/pull/76

PR #450, opened 2022-02-18
Implement hybrid pinch zoom mode.

kareltucek:


mondalaci:
As for the zoom shortcuts, Shift is actually not needed for "+", so rather, "=" needs to be invoked. Also, Mac uses Super (Gui) for the zoom shortcut, not Alt.

Perhaps even more importantly, I've just realized that emulating zooming via keyboard shortcuts is fundamentally problematic and may result in a lot of support burden; I've tested this feature with a laptop using Hungarian layout, and unsurprisingly, the scancodes didn't map to the +/- characters.

@benedekkupper Do you know of a platform-independent way to implement zooming via USB?

benedekkupper:
To answer that I just jump into the [HID usage tables document](https://www.usb.org/sites/default/files/hut1_22.pdf):
> 22D AC Zoom In Sel
22E AC Zoom Out Sel
22F AC Zoom LC
230 AC Full Screen View Sel
231 AC Normal View Sel
232 AC View Toggle Sel
233 AC Scroll Up Sel
234 AC Scroll Down Sel
235 AC Scroll LC

Where the "Sel" elements are the ones that can be mapped to a key as scancode, the LC means it's a linear value, so that would need modification of the report desciptor and layout.

mondalaci:
@benedekkupper I've just tried the two referenced zoom scancodes by mapping two keys to M557 and M558 in Agent, and they didn't zoom in the browser. Anything else that might work?

benedekkupper:
Nope, this is the best we have. The concept is defined by HID, but it's not enforced that each OS has to implement it (we can see in Linux source code that KEY_ZOOMIN exists and bound to this scancode), and that each application running on these OSes handle these key events. We can only say: "Hey, we can zoom in-out with our keyboard, but your shitty OS / application doesn't handle it as it should, please fix it."

mondalaci:
@benedekkupper Gotcha. How do touchpads communicate pinch to zoom toward the OS?

benedekkupper:
Some googling tells me that touch pads have a very different level of abstraction on HID, see the Digitizers usage page. Such device just tells the host where the physical contacts are made on the surface, and the host OS/application interprets gestures such as pinch zoom. There is some [vague Windows-specific guide](https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/windows-precision-touchpad-required-hid-top-level-collections#selective-reporting-feature-report) to implement such HID devices. I myself never had the chance to test a touchpad, so I cannot really give you further insight.

mondalaci:
Thanks for looking into this! We should implement the planned solution then.

kareltucek:
@mondalaci this means that you just want the ALT and SHIFT fixed?

kareltucek:
(Done.)

kareltucek:
@mondalaci just out of curiosity, which ones?

(My original line of thought was exactly that Ctrl+= should reset zoom to default rather than act as Ctrl++, but I failed to find a single example on my workstation.)

mondalaci:
Gnome Terminal didn't work with Ctrl + =

PR #460, opened 2022-03-17
Add git properties.

kareltucek:


right/src/debug.c
```diff
@@ -17,11 +17,8 @@ static uint32_t watchInterval = 500;
 
 void AddReportToStatusBuffer(usb_basic_keyboard_report_t *report)
 {
-    Macros_SetStatusNum(report->modifiers);
-    for (int i = 0; i < 6; i++) {
```
mondalaci, line 21: I'd rather not use `int`, but a specific type like `uint8_t`.

mondalaci: I'm retarted as this has been removed.

kareltucek: Still good point.

PR #470, opened 2022-04-22
Rename doubletapDelay to doubletapTimeout.

kareltucek:


mondalaci:
I'd much rather prefer branching based on smart macro version and explicitly disallow deprecated language elements.

kareltucek:
How?

I can't efficiently branch based on semver in the macro engine given that stock and fork use separate semver spaces. (Which cannot be avoided as long as you keep insisting on changes to already released features while also refusing to inherit macro versioning of the fork.)

I would also much rather not break people's setups when there is no reason to do so.

mondalaci:
Then please comment out the old naming for the official UHK firmware build.

kareltucek:
Like this?

PR #485, opened 2022-06-08
Fix zoom quirks

### LINKED ISSUE #481: Touchpad pinch-to-zoom bug
When pinch-to-zooming with the touchpad module, the UHK sends Ctrl +/-, and afterward, it sends a Super scancode by itself, activating the menu on Linux and Windows.

Besides fixing this bug, it'd be important to have a zoom/zoomPc/zooomMac configuration option for pinch-to-zoom.

kareltucek:
Closes #481.


mondalaci:
Zooming is way more sensitive in `zoom` mode than in `zoomPc` mode.

kareltucek:
Good catch. Fixed.

PR #486, opened 2022-06-08
Common trunk upstream

kareltucek:
Macro engine updates that have accumulated over last few months...


kareltucek:
Actually, let's do it from common_trunk so that this PR does not block further work...

PR #487, opened 2022-06-08
Common trunk sync

kareltucek:
Macro engine updates that have accumulated over last few months...


kareltucek:
```
    COMMAND = set navigationModeAction.NAVIGATIONMODECUSTOM.DIRECTION ACTION
```

This change is significant as action is now `macro MACRONAME | keystroke SHORTCUT`, which means that the html documentation will need to be updated. Should I try to keep html docs up to date, or should I leave them to you (as you have it in fresh memory @mondalaci )?

mondalaci:
Thanks for the heads-up! I'll take care of it.

Issue #348, opened 2021-04-08
Reapply keymap upon module change 


mondalaci:
The current keymap has to be reapplied when attaching new modules, otherwise the configuration of the previous module remains.

mhantsch:
Reminder to myself to re-test the scenario in #389 once this issue has been resolved.

mondalaci:
Resolved by #497.

PR #497, opened 2022-07-02
Reload keymap when module is swapped.

### LINKED ISSUE #361: Some Keybindings not working with 8.10.9
I have Mod + F bound to L_Ctrl + L_Super + Right Arrow
and Mod + S bound to L_Ctrl + L_Super + Left Arrow.

When I updated to firmware 8.10.9 these bindings stopped working.
Also, Fn + 1, 2, 3, 4, 5, and 6 failed to cycle through the layout mappings. 

I downgraded to 8.10.8, and functionality has been restored.

I'm running the latest agent on Linux (OpenSUSE Tumbleweed) Kernel 5.12.0-1

I've also connected to a 2016 MBP, and the keybindings are also not working there, with 8.10.9

kareltucek:
Closes #361, #348.

### DIFF
--- a/right/src/slave_drivers/uhk_module_driver.c
+++ b/right/src/slave_drivers/uhk_module_driver.c
@@ -79,7 +79,7 @@ void UhkModuleSlaveDriver_Init(uint8_t uhkModuleDriverId)
 // moduleId. However, this triggers macro events, and we don't want to trigger
 // the same macro multiple times if we can detect that another module is also
 // going through the initialization sequence.
-void ReloadKeymapIfNeeded()
+static void reloadKeymapIfNeeded()
 {
     bool someoneElseWillDoTheJob = false;
     for (uint8_t i = 0; i < UHK_MODULE_MAX_SLOT_COUNT; i++) {
@@ -195,7 +195,7 @@ status_t UhkModuleSlaveDriver_Update(uint8_t uhkModuleDriverId)
             bool isMessageValid = CRC16_IsMessageValid(rxMessage);
             if (isMessageValid) {
                 uhkModuleState->moduleId = rxMessage->data[0];
-                ReloadKeymapIfNeeded();
+                reloadKeymapIfNeeded();
             }
             status = kStatus_Uhk_IdleCycle;
             *uhkModulePhase = isMessageValid ? UhkModulePhase_RequestModuleKeyCount : UhkModulePhase_RequestModuleId;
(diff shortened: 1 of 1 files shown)


right/src/slave_drivers/uhk_module_driver.c
```diff
@@ -73,6 +75,25 @@ void UhkModuleSlaveDriver_Init(uint8_t uhkModuleDriverId)
...
+// the same macro multiple times if we can detect that another module is also
+// going through the initialization sequence.
+void ReloadKeymapIfNeeded()
+{
+    bool someoneElseWillDoTheJob = false;
+    for (uint8_t i = 0; i < UHK_MODULE_MAX_SLOT_COUNT; i++) {
+        uhk_module_state_t *uhkModuleState = UhkModuleStates + i;
+        uhk_slave_t* slave = Slaves + i;
+
+        someoneElseWillDoTheJob = uhkModuleState->moduleId == 0 && slave->isConnected;
```
mondalaci, line 89: What's the logic behind this line? moduleId==0, hence the module is disconnected, but the slave is connected.

right/src/slave_drivers/uhk_module_driver.c
```diff
@@ -73,6 +75,25 @@ void UhkModuleSlaveDriver_Init(uint8_t uhkModuleDriverId)
     uhkModuleState->pointerDelta.y = 0;
 }
 
+// When module is swapped, we need to reload its Keymap once we know its
+// moduleId. However, this triggers macro events, and we don't want to trigger
+// the same macro multiple times if we can detect that another module is also
+// going through the initialization sequence.
+void ReloadKeymapIfNeeded()
```
mondalaci: Should be lowerCamelCase because it's a local function.

right/src/slave_drivers/uhk_module_driver.c
kareltucek: If I read the code correctly, then `isConneted` is set to true after first successfull transfer. However, moduleId is not known until roughly 4th exchange - that is after 8 transfers are finished. 

This means that after startup, or whenever left half (with a dangling keycluster module) is reconnected, the modules will go through their initiation sequences in parallel. I.e., being connected, obviously knowing slot id, but not yet knowing the module id.

right/src/slave_drivers/uhk_module_driver.c
kareltucek: Indeed.

mondalaci:
@kareltucek Excellent job, especially preventing multiple macro event triggers.

kareltucek:
Actually, poor job. Idea is correct, but the `=` operator is clearly wrong 🤦.

Issue #399, opened 2021-08-15
Inaccurate module attached state


mondalaci:
When attaching or detaching a module, it can temporarily or permanently affect the attached/detached state of other modules, including the left keyboard half. According to my experience, only the key cluster module is permanently affected sometimes. This bug is easy to notice in Agent, and with get-device-state.ts

kareltucek:
Is this still relevant?

mondalaci:
Yes, it is. There's a minor and major issue.

The minor issue is that when attaching or detaching a module, other modules or the left keyboard half often momentarily disappears.

The major issue is that when detaching the key cluster, sometimes it stays attached in Agent.

These issues are reproducible not only with Agent, but with `agent/packages/usb/get-device-state.ts`, too.

kareltucek:
> The minor issue is that when attaching or detaching a module, other modules or the left keyboard half often momentarily disappears.

This results from i2c reinitializations.

From a quick glance at I2C_MasterTransferHandleIRQ and I2C_MasterTransferRunStateMachine it is not at all clear that the i2c subsystem always calls the slaveSchedulerCallback after a (possibly failed) i2c transfer. So I suspect some problem there.

Should I dig deeper?

> The major issue is that when detaching the key cluster, sometimes it stays attached in Agent.

This results from slaves not being properly disconnected on such reinitialization. Fix in #499 .

mondalaci:
Major issue beautifully solved.

As for the minor issue, the I2C bus is not designed to be hot-swappable. When attaching or detaching a module, it likely interferes with the actual communication of the bus and makes other modules disconnect as a side-effect. As a workaround, maybe the updated connection state of the modules affected by hot-swapping shouldn't be exposed via USB for a short period, like about half a second, only afterward.

On a related note, it may be worth checking the various error counters of modules, such as `arbitrationLost` and `timeout` by running `watch agentDir/packages/usb/get-i2c-health.ts` to more explicitly detect which modules are affected by the above side-effect.

kareltucek:
Still, it should be a basic decency for the KSDK library to report failed transactions flawlessly.

Unless we want to actually figure out cause of the issue, I would rather not clutter the code by workarounds as long as there is not a strong reason to do so.

mondalaci:
Does the KSDK not report failed transactions with `arbitrationLost` or `timeout` status when the minor issue happens?

[...]

PR #499, opened 2022-07-04
Properly disconnect slaves on i2c reinitialization.

kareltucek:


right/src/slave_scheduler.c
```diff
@@ -114,6 +116,9 @@ void InitSlaveScheduler(void)
 
     for (uint8_t i=0; i<SLAVE_COUNT; i++) {
         uhk_slave_t *currentSlave = Slaves + i;
+        if (currentSlave->isConnected && currentSlave->disconnect) {
```
mondalaci, line 119: Shouldn't all slaves be disconnected upon scheduler initialization?

mondalaci: Sorry, I guess https://github.com/UltimateHackingKeyboard/firmware/issues/399#issuecomment-1173870244 is the explanation.

kareltucek: > Sorry, I guess https://github.com/UltimateHackingKeyboard/firmware/issues/399#issuecomment-1173870244 is the explanation.

Is it?

> Shouldn't all slaves be disconnected upon scheduler initialization?

Should they? Those which are not "connected" are guaranteed to have already been properly disconnected, or never even were connected.

kareltucek: Oh, sorry, I see the misunderstanding now. Yes, we are talking watchdog-induced re-initializations.

mondalaci:
Very well done! Stuck modules fixed.

PR #502, opened 2022-08-10
Add resetTrackpoint command.

### LINKED ISSUE #382: Trackpoint module drift
Sometimes, the trackpoint module drifts. In other words, the mouse pointer keeps moving in a direction without any input being applied. This is a known issue of original IBM/Lenovo TrackPoints, too.

Given that the direction and speed of such drifts are constant, I plan to detect drifts on the firmware level, then execute a PS/2 reinitialization sequence. The trackpoint firmware will contain the above fix without affecting anything else.

kareltucek:
Resolves #382.

### DIFF
--- a/right/src/slave_drivers/uhk_module_driver.c
+++ b/right/src/slave_drivers/uhk_module_driver.c
@@ -14,6 +14,8 @@
 
 uhk_module_state_t UhkModuleStates[UHK_MODULE_MAX_SLOT_COUNT];
 
+static bool shouldResetTrackpoint = false;
+
 uint8_t UhkModuleSlaveDriver_SlotIdToDriverId(uint8_t slotId)
 {
     return slotId-1;
@@ -26,7 +28,7 @@ uint8_t UhkModuleSlaveDriver_DriverIdToSlotId(uint8_t uhkModuleDriverId)
 
 void UhkModuleSlaveDriver_ResetTrackpoint()
 {
-    UhkModuleStates[UhkModuleDriverId_RightModule].phase = UhkModulePhase_ResetTrackpoint;
+    shouldResetTrackpoint = true;
 }
 
 static uint8_t keyStatesBuffer[MAX_KEY_COUNT_PER_MODULE];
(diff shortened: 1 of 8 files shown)


trackpoint/src/module.c
```diff
@@ -133,6 +154,38 @@ static bool readByte()
     return false;
 }
 
+int16_t errorCode = 0;
+
+#define RQ_TO_SEND(AT)                  \
```
kareltucek, line 159: I have made these macros in order to make adding new commands into the loop easier. (For testing purposes.)

Should I leave them here, or revert to the original?

right/src/slave_drivers/uhk_module_driver.c
```diff
@@ -24,6 +24,11 @@ uint8_t UhkModuleSlaveDriver_DriverIdToSlotId(uint8_t uhkModuleDriverId)
     return uhkModuleDriverId+1;
 }
 
+void UhkModuleSlaveDriver_ResetTrackpoint()
+{
+    UhkModuleStates[UhkModuleDriverId_RightModule].phase = UhkModulePhase_ResetTrackpoint;
```
kareltucek: I guess writing phase directly could mess up module initialization (if resetTrackpoint is called during initialization phase). Will add an indirection in order to fix it...

shared/slave_protocol.h
```diff
@@ -23,6 +23,7 @@
         SlaveCommand_RequestKeyStates,
         SlaveCommand_SetTestLed,
         SlaveCommand_SetLedPwmBrightness,
+        SlaveCommand_ResetTrackpoint,
```
mondalaci, line 26: I feel uneasy about polluting `slave_command_t` with module-specific commands. We may also run out of command ids sooner, eventually this way. I first thought about adding a generic module reset command, but I'm unsure whether we'd use it for other modules later. I think having `SlaveCommand_ModuleSpecificCommand` is the best way forward, and using the second byte with its own enum per module as the actual module-specific command id.

trackpoint/src/module.c
```diff
@@ -29,6 +29,26 @@ void Module_Init(void)
...
+    GPIO_PinInit(TP_RST_GPIO, TP_RST_PIN, &(gpio_pin_config_t){.pinDirection=kGPIO_DigitalOutput, .outputLogic=1});
+}
+
+bool shouldReset = false;
+
+void resetBoard()
+{
+    GPIO_WritePinOutput(TP_RST_GPIO, TP_RST_PIN, 0);
+
+    for (volatile uint32_t i=0; i<10000; i++);
```
mondalaci, line 44: We'll optimize the firmware for low power consumption, especially for wireless, and busy loops hurt battery life, so please let's not use them, even if they're rarely invoked. Let's rather use the state machine to pull TP_RST low and then high.

trackpoint/src/module.c
mondalaci: For me, it's harder to digest the code with the macros included, so I'd prefer not to have them.

mondalaci:
@kareltucek Did the reset actually work? I can reproduce the drift rarely; that's why I ask.

kareltucek:
> @kareltucek Did the reset actually work? I can reproduce the drift rarely; that's why I ask.

Yes, it did!

PR #510, opened 2022-08-20
Add resetTrackpoint command.

### LINKED ISSUE #382: Trackpoint module drift

kareltucek:
Closes #382.

(Reworked according to requests in #502.)

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -60,6 +60,7 @@ The following grammar is supported:
     COMMAND = statsActiveKeys
     COMMAND = statsActiveMacros
     COMMAND = statsRegs
+    COMMAND = resetTrackpoint
     COMMAND = diagnose
     COMMAND = printStatus
     COMMAND = {setStatus  | setStatusPart} <custom text>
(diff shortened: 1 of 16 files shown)


mondalaci:
Now the trackpoint module is not immediately useable right after attaching it due to the explicit recalibration. Please don't reset it upon firmware startup.

kareltucek:
Oops. Fixed.

mondalaci:
The resetTrackpoint command seemingly doesn't do anything now. I can move the pointer with the trackpoint module uninterruptedly when triggering it.

mondalaci:
I've just realized that this command only works when extended macros are enabled. Please enable it always.

kareltucek:
Good catch! Fixed.

Issue #505, opened 2022-08-15
Firmware 9.0.0 blocking issues


mondalaci:
- [x] #500
- [x] #501
- [x] Add `macroEngine.extendedCommands`
- [x] Hide `debounceDelay` behind `macroEngine.extendedCommands`
- [x] Laci: update informal docs regarding axis locks. By mistake, the parser was using *AxisLockEnabled instead of *AxisLock. Add caretAxisLock.
- [x] Update informal documentation according to the above (Laci's part)
- [x] Review informal documentation

mondalaci:
@kareltucek I plan to catch up with my part this week. In the meantime, feel free to implement `macroEngine.extendedCommands`.

PR #511, opened 2022-08-22
Common trunk upstream

kareltucek:
Some bug fixes:

- state split for left and navigation mode was broken by one fix, so it should be fine now.
- tapKeySeq fixed. (It had uninitialized modifiers, which caused problems time to time).
- parser fixed for `*AxisLock*` patterns.
- add some missing led updates in set macros
- fix macro freezes on noOp commands

Todo @mondalaci : 
- [x] update informal docs regarding axis locks. By mistake, the parser was using `*AxisLockEnabled` instead of `*AxisLock`. Also, `caretAxisLock` was added some time ago, and I apparently forgot to connect it in the parser.


mondalaci:
Transferred the item to be checked to #505.

PR #515, opened 2022-09-02
A bunch of unrelated smaller fixes

benedekkupper:
Today I just wanted to check the current USB descriptors of the UHK.

... but then I noticed that all the interrupt endpoint max packet sizes (MPS) are the same, even though the report lengths are different. So I went ahead and fixed these sizes. The gain is at the basic keyboard interface, which has to split its reports into multiple transfers without this change.

... but then I couldn't compile the FW anymore. So I found out about the `versions.h` generated header. It's best to not change header content frequently, as all source files including it will be rebuilt during incremental builds. Therefore I split the header's contents and minimized the use of the generated header.

... but then I saw that the CI builds are failing, so I checked that code area, and made some optical and code size tuning on it. [edit: the CI build is now passing]

Now I have 3 commits that are very unrelated but I don't want to open separate PRs for them.


mondalaci:
@benedekkupper These are very welcome changes. Thank you for caring about the general health of the codebase!

PR #517, opened 2022-09-07
Fix touchpad oscillation.

kareltucek:


kareltucek:
However, this will probably skip a phase avery time a nack is encountered 😑.

PR #518, opened 2022-09-07
[WIP]: Set touchpad charging mode to manual in order to make it respond to I2c.

### LINKED ISSUE #364: Touchpad module oscillation
The touchpad module oscillates in Agent.

I noticed this issue a while ago. The right keyboard half keeps querying the touchpad over I2C, and the touchpad NAKed from time to time which caused the above phenomenon in Agent. I set MANUAL_CONTROL to 1 and MODE_SELECT to Active mode in bb9bb29 which resolved the issue.

Afterward, I had a chat with Azoteq to improve the tracking accuracy of the touchpad and changed some values in a2af6d994015cc3a2dd4f9b91ca1e004ea7df95a which apparently reintroduced the NAKs. The following values were changed in the Trackpad Hardware Setup values of the Prox Hardware Settings section:

1. Pass Length: 101 -> 010
2. UpLength: 101 -> 010
3. VTrip: 0011 -> 1000
4. Opamp Bias: 2.5 uA -> 5.0 uA

According to my findings, changing either 1. or 2. negates the effect of the `MANUAL_CONTROL=1` and `MODE_SELECT=Active mode` settings.

We have to figure out which settings to use to maximize tracking accuracy and get rid of the NAKs.

kareltucek:
Closes #364

### DIFF
--- a/right/src/slave_scheduler.c
+++ b/right/src/slave_scheduler.c
@@ -93,13 +93,16 @@ static void slaveSchedulerCallback(I2C_Type *base, i2c_master_handle_t *handle, 
             currentSlave->init(currentSlave->perDriverId);
         }
 
-        status_t currentStatus = currentSlave->update(currentSlave->perDriverId);
+        bool shouldYield = true;
+        status_t currentStatus = currentSlave->update(currentSlave->perDriverId, &shouldYield);
         if (IS_STATUS_I2C_ERROR(currentStatus)) {
             LogI2cError(currentSlaveId, currentStatus);
         }
         isTransferScheduled = currentStatus != kStatus_Uhk_IdleSlave && currentStatus != kStatus_Uhk_IdleCycle;
 
-        if (currentStatus != kStatus_Uhk_IdleCycle) {
+        bool progressToNextSlave = shouldYield || !currentSlave->isConnected;
+
+        if (progressToNextSlave) {
             previousSlaveId = currentSlaveId++;
             if (currentSlaveId >= SLAVE_COUNT) {
                 currentSlaveId = 0;
--- a/right/src/slave_drivers/is31fl3xxx_driver.c
+++ b/right/src/slave_drivers/is31fl3xxx_driver.c
@@ -235,7 +235,7 @@ void LedSlaveDriver_Init(uint8_t ledDriverId)
     currentLedDriverState->ledIndex = 0;
 }
 
-status_t LedSlaveDriver_Update(uint8_t ledDriverId)
+status_t LedSlaveDriver_Update(uint8_t ledDriverId, bool *yield)
 {
     status_t status = kStatus_Uhk_IdleSlave;
     uint8_t *ledValues = LedDriverValues[ledDriverId];
--- a/right/src/slave_drivers/is31fl3xxx_driver.h
+++ b/right/src/slave_drivers/is31fl3xxx_driver.h
@@ -81,6 +81,6 @@
     void LedSlaveDriver_DisableLeds(void);
     void LedSlaveDriver_UpdateLeds(void);
     void LedSlaveDriver_Init(uint8_t ledDriverId);
-    status_t LedSlaveDriver_Update(uint8_t ledDriverId);
+    status_t LedSlaveDriver_Update(uint8_t ledDriverId, bool *yield);
 
 #endif
--- a/right/src/slave_drivers/kboot_driver.c
+++ b/right/src/slave_drivers/kboot_driver.c
@@ -23,7 +23,7 @@ void KbootSlaveDriver_Init(uint8_t kbootInstanceId)
 {
 }
 
-status_t KbootSlaveDriver_Update(uint8_t kbootInstanceId)
+status_t KbootSlaveDriver_Update(uint8_t kbootInstanceId, bool *yield)
 {
     status_t status = kStatus_Uhk_IdleSlave;
 
--- a/right/src/slave_drivers/kboot_driver.h
+++ b/right/src/slave_drivers/kboot_driver.h
@@ -52,6 +52,6 @@
 // Functions:
 
     void KbootSlaveDriver_Init(uint8_t kbootInstanceId);
-    status_t KbootSlaveDriver_Update(uint8_t kbootInstanceId);
+    status_t KbootSlaveDriver_Update(uint8_t kbootInstanceId, bool *yield);
 
 #endif
--- a/right/src/slave_drivers/touchpad_driver.c
+++ b/right/src/slave_drivers/touchpad_driver.c
@@ -9,45 +9,45 @@
 #include "key_states.h"
 #include "usb_interfaces/usb_interface_mouse.h"
 #include "touchpad_driver.h"
-
+#include "debug.h"
+#include "timer.h"
 
 /*
-Actually produced sequences:
-
-1       single tap
-2       tap and hold
-      1 number of fingers
-
-Tap:
-
-0  0  0
-0  0  1
-1  0  0
-
-
-tap and hold:
-
-0  0  0
-0  0  1
-2  0  1
-
-tap and hold (separate):
-
-0  0  0
-0  0  1
-1  0  0
-0  0  0
-0  0  1
-2  0  1
-
-two taps:
-
-0  0  0
-0  0  1
-1  0  0
-0  0  0
-0  0  1
-1  0  0
+|  Actually produced sequences:
+| 1       single tap
+| 2       tap and hold
+|       1 number of fingers
+|
+| Tap:
+|
+| 0  0  0
+| 0  0  1
+| 1  0  0
+|
+|
+| tap and hold:
+|
+| 0  0  0
+| 0  0  1
+| 2  0  1
+|
+| tap and hold (separate):
+|
+| 0  0  0
+| 0  0  1
+| 1  0  0
+| 0  0  0
+| 0  0  1
+| 2  0  1
+|
+| two taps:
+|
+| 0  0  0
+| 0  0  1
+| 1  0  0
+| 0  0  0
+| 0  0  1
+| 1  0  0
 */
 
 
@@ -75,6 +75,8 @@ touchpad_events_t TouchpadEvents;
 uint8_t phase = 0;
 static uint8_t enableEventMode[] = {0x05, 0x8f, 0x07};
 static uint8_t enableManualMode[] = {0x05, 0x8e, 0xec};
+/* The touchpad wil NAK if we ask it more often than the configured report rate. */
+static uint8_t setReportRate[] = {0x05, 0x7b, 0x01};
 static uint8_t getGestureEvents0[] = {0x00, 0x0d};
 static uint8_t getRelativePixelsXCommand[] = {0x00, 0x12};
 static uint8_t closeCommunicationWindow[] = {0xee, 0xee, 0xee};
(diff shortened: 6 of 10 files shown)


kareltucek:
I have pushed some new commits to this PR...

Turns out that if we try to communicate with the touchpad sooner than the requested report rate, it also NACKs -> I have set the report rate to 1ms. It should therefore no longer NACK when left half is not connected. If you wish to set it to other value, please let me know. 

Furthermore, I have added a parameter which allows slaves to decide whether they want to yield or block. I have made first 4 slaves to "always" block. This way, touchpad update loop length was reduced from 23ms to 7ms. Other modules should have now the same 7ms response time (from i2c point of view). I didn't measure their time before, but it should be quite similar to what it was before. (Their "relative priority" increased w.r.t. to LED drivers, yet decreased w.r.t. to the touchpad driver.)

I am not sure about some details of the proper i2c logic for the kboot slave -> will followup in code comments.

right/src/slave_scheduler.c
```diff
@@ -93,13 +93,16 @@ static void slaveSchedulerCallback(I2C_Type *base, i2c_master_handle_t *handle,
...
 
-        status_t currentStatus = currentSlave->update(currentSlave->perDriverId);
+        bool shouldYield = true;
+        status_t currentStatus = currentSlave->update(currentSlave->perDriverId, &shouldYield);
         if (IS_STATUS_I2C_ERROR(currentStatus)) {
             LogI2cError(currentSlaveId, currentStatus);
         }
         isTransferScheduled = currentStatus != kStatus_Uhk_IdleSlave && currentStatus != kStatus_Uhk_IdleCycle;
 
-        if (currentStatus != kStatus_Uhk_IdleCycle) {
```
kareltucek, line 102: I have removed this condition effectively, since keeping the logic intact would significantly decrease code readability, and I didn't see a reason to keep it.

kareltucek:
Yet, it seems like this scheduling change has broken module flashing. Any ideas why?

right/src/slave_scheduler.c
kareltucek: -> Should kboot idle cycles at https://github.com/UltimateHackingKeyboard/firmware/pull/518/files#diff-0f726b9e03a7f7b68ce24085975f37582511def44889dc80637f55b7b930b25aL44 be forced to not yield?

mondalaci: `kStatus_Uhk_IdleCycle` is used deliberately, and it shouldn't be ignored. When a slave returns it, the scheduler should not move to the next slave. It's used when no I2C transaction is done in a phase, only validation, to jump to the next phase quickly. Have you tried to use it instead of `yield`? Maybe it's a viable alternative.

Please only use `kStatus_Uhk_IdleCycle`, `yield`, or any similar solution to alter what's really necessary, which I believe is only the touchpad slave driver, to keep this change minimal. My memories about the codebase are a bit foggy, and the smaller the scope of the change, the easier for me to understand it.

Great finding about touchpad timing!

kareltucek: > kStatus_Uhk_IdleCycle is used deliberately, and it shouldn't be ignored. When a slave returns it, the scheduler should not move to the next slave. It's used when no I2C transaction is done in a phase, only validation, to jump to the next phase quickly. Have you tried to use it instead of yield? Maybe it's a viable alternative.

Shouldn't I return the I2C status when a phase consists of I2C communication?

Also, am I supposed to return kStatus_Success in order to yield when no I2C communication took place?

(The required yield/block behaviour will have to be composed somewhere anyways, so I might as well extend the bool to a 3 state enum "yield", "block", "dontcare".)

> Please only use kStatus_Uhk_IdleCycle, yield, or any similar solution to alter what's really necessary, which I believe is only the touchpad slave driver, to keep this change minimal. My memories about the codebase are a bit foggy, and the smaller the scope of the change, the easier for me to understand it.

Shouldn't we try to reduce the overall response time? 

I mean, if I make touchpad process its messages atomically, but leave the module slave driver intact, the left half/module update will take twice the time of touchpad update.

mondalaci: Indeed, when the phase consists of an I2C communication, the status code code should be returned, so `kStatus_Uhk_IdleCycle` is only useful when no communication happens.

I think the most elegant way is to return the status value and the yield value in a struct by slaves.

Instead of the `yield` name, I'd rather use the more intuitive `hold` name to hold control and not allow the scheduler to move to the next slave. I don't understand why 3 states would be useful, and I'd use a boolean `hold`.

When no communication takes place, I'd use a new `kStatus_None` value, and then instead of using `kStatus_Uhk_IdleCycle`, it'd make more sense to return `{.status=kStatus_None, .hold=true}`.

I don't see a way to reduce the overall response time. We can make practical compromises, but we have only so much I2C bandwidth.

kareltucek: > I don't understand why 3 states would be useful, and I'd use a boolean `hold`.

Because if you want to have the noop-cycle-skipping scheduling behavior coupled to `kStatus_Uhk_IdleCycle` (which actually sounds reasonable), then you need to let the slave scheduler know whether it should respect the `hold` or the `kStatus_Uhk_IdleCycle`?

Also, it might be better to explicitly specify whether or not I (as a human being and author of the commit) care about the scheduling policy rather than having to specify the `.hold = true` everywhere, because if I explicitly have to state the value everywhere, then I erase the information of whether I (as the commit author) do care or do not and why so.

> When no communication takes place, I'd use a new `kStatus_None` value, and then instead of using `kStatus_Uhk_IdleCycle`, it'd make more sense to return `{.status=kStatus_None, .hold=true}`.

What about `kStatus_Uhk_IdleSlave`... ? You are breaking the pattern...

> I don't see a way to reduce the overall response time. We can make practical compromises, but we have only so much I2C bandwidth.

Well: make sure that all modules update at the same pace w.r.t. the number of actually completed update cycles (proposed behavior) rather than w.r.t. completed I2C exchanges (current behavior)

To model the impact, assume:
- each I2c transfer takes 1ms (not realistic)
- module update cycle takes 2 transfers
- touchpad update cycle takes 6 transfers
- assume only 2 uhk modules and 1 touchpad module to be in the update loop
- (I also assume that in module code, data to be sent are prepared at receiving the request rather than when sending them, which means that the message actually sits one slave scheduler cycle without being updated, therefore increasing the actual latency by a further factor of 1.5x.)

strategy | current scheduling | make just touchpad processing atomic | atomic touchpad and atomic modules
--|--|--|--
module update rate | 6ms      | 16ms | 10 ms
module latency | 9ms          | 24ms | 10 ms
touchpad update rate | 18ms  | 8ms   | 10 ms

For instance, If I apply the blocking/holding policy to the touchpad only (as you seem to suggest), then the update loop will look like:

- left half 1 I2c transfer
- left module 1 I2c transfer
- touchpad 1 update cycle == 6 I2c transfers
- (touchpad update loop completed here)
- left half 1 I2c transfer
- (left half update loop completed here, since it takes 2 I2c transfers)
- left module 1 I2c transfer
- (left module update loop completed here, since it takes 2 I2c transfers)
- touchpad 1 update cycle == 6 I2c transfers
- (touchpad update loop completed here)

mondalaci: Thanks for elaborating. I agree that we should cluster module I2C transactions for the modules to have the same refresh rate.

I forgot that there's both `kStatus_Uhk_IdleSlave` and `kStatus_Uhk_IdleCycle`. I'd ditch them both and use `kStatus_None` instead with and also use `.lock = true | false`.

I'd always respect `.hold` regardless of `.status`.

`.hold = true` shouldn't be "explicitly stated everywhere", only at the beginning of slave update functions, which is already done for `status`.

kareltucek: > `.lock = true | false`.

What is `.lock`?

> I'd always respect `.hold` regardless of `.status`.
> 
> `.hold = true` shouldn't be "explicitly stated everywhere", only at the beginning of slave update functions, which is already done for `status`.

~~I mean, this implies that `.hold = true` has to be manually included everywhere where `kStatus_Uhk_IdleCycle` is returned. Furthermore, the idea of IdleCycle phases being contracted is entirely removed from the code, and is required to be filled in by the user (by either including `.hold = true` or by deliberately omitting it).~~

~~This means doing exactly what I meant by removing the `kStatus_Uhk_IdleCycle` condition in the first post.~~

~~Edit: I guess I know what you want me to do with the code, it is just that your two requirements contradict themselves, because your requirement of IdleCycle logic remaining in the code is directly broken by "always respecting `.hold`". Furthermore, if you expect the logic of IdleCycle not to be ignored (for instance during future maintenance), then you have a code that is "broken by default".~~

Edit2: Nevermind. I guess if I just ignore all your requirements and implement it just as I intended, you will get exactly the code you want to have...

mondalaci: > What is `.lock`?

Sorry, I meant `.hold`.

What's the problem with always respecting `.hold`? Please give me an example.

kareltucek: There is no problem. 

It is just that you stated that:

> kStatus_Uhk_IdleCycle is used deliberately, and it shouldn't be ignored. When a slave returns it, the scheduler should not move to the next slave. 

Which is in direct contradiction with respecting `.hold`, because respecting `.hold` means that `kStatus_Uhk_IdleCycle` is no longer semantically coupled with the requirement of directly skipping to the next phase of the same slave.

mondalaci: Ok, I think we're on the same page then.

mondalaci:
I've just realized that I haven't answered your last question. I have a lot on my plate, so can you bisect your commits to find the culprit?

kareltucek:
> I've just realized that I haven't answered your last question. I have a lot on my plate, so can you bisect your commits to find the culprit?

I believe the last commit of this PR is the culprit. In other words - either increased kboot slave delay or some other logic that I don't know about that is going on during flashing sequence.

May I have a short summary of how flashing works from the firmware's perspective? 

(My current code is available at https://github.com/kareltucek/firmware/tree/fix_touchpad_oscillation2_wip .)

mondalaci:
Short summary of how module flashing works:

1. The right half makes the chosen module jump to its bootloader with `UsbCommand_JumpToModuleBootloader()`
2. There's a short time (several milliseconds) after which the module bootloader timeouts, so it must be pinged quickly to stay in the bootloader and not jump to the firmware. This is done by `UsbCommand_SendKbootCommandToModule()`, and the mechanism is implemented by `kboot_driver.c` which I guess is not getting scheduled (frequently enough) in your branch, so the module bootloader quickly jumps to the firmware.

Actually, the above order may be backward to ensure that the kboot slave driver spins the ping command by the time the module jumps to the bootloader, but the order is implemented in Agent and doesn't affect the firmware.

kareltucek:
@mondalaci can the "short time" simply be extended?

mondalaci:
@kareltucek [The bootloader timeout is actually 100 ms](https://github.com/UltimateHackingKeyboard/firmware/blob/af9683012ca53e0d7eb8d593a0813ae930df47ce/shared/bootloader.h#L14), and it can be changed, but please, let's not change it. The value was selected so that the timeout is not noticeable for the user, yet, it's sufficient for pinging the bootloader.

kareltucek:
Well, 100 should be sufficient. Also, increasing it is not helping.

-> will post new PR

PR #533, opened 2022-10-09
Fix touchpad oscillation3

kareltucek:


mondalaci:
@kareltucek I'm unsure when I should test this PR. Please let me know.

right/src/slave_scheduler.c
```diff
@@ -93,13 +93,14 @@ static void slaveSchedulerCallback(I2C_Type *base, i2c_master_handle_t *handle,
...
 
-        status_t currentStatus = currentSlave->update(currentSlave->perDriverId);
+        slave_result_t res = currentSlave->update(currentSlave->perDriverId);
+        status_t currentStatus = res.status;
         if (IS_STATUS_I2C_ERROR(currentStatus)) {
             LogI2cError(currentSlaveId, currentStatus);
         }
         isTransferScheduled = currentStatus != kStatus_Uhk_IdleSlave && currentStatus != kStatus_Uhk_IdleCycle;
 
-        if (currentStatus != kStatus_Uhk_IdleCycle) {
```
kareltucek, line 102: The `KbootDriverState.command != KbootCommand_Idle` is not very nice, however my attempts to fix it by changing scheduling orders while also respecting `hold` failed miserably. And debugging it is sort of problematic since both halves are resetted during flashing process...

mondalaci: Would having a logic analyzer help? If so, I'd send you one.

kareltucek: > Would having a logic analyzer help? If so, I'd send you one.

With analyzing what is wrong probably yes.

Not sure whether knowing what's wrong would help to actually fix it.

> I'm unsure when I should test this PR. Please let me know.

Unless you are significantly unhappy about that Kboot state condition in the slave scheduler, please do test the PR.

kareltucek:
Actually, there is still a couple of problems. -> no need to test atm.

PR #534, opened 2022-10-21
Fix touchpad oscillation

kareltucek:
Please test this PR and let me know your findings.

### DIFF
--- a/right/src/slave_drivers/is31fl3xxx_driver.c
+++ b/right/src/slave_drivers/is31fl3xxx_driver.c
@@ -235,9 +235,9 @@ void LedSlaveDriver_Init(uint8_t ledDriverId)
     currentLedDriverState->ledIndex = 0;
 }
 
-status_t LedSlaveDriver_Update(uint8_t ledDriverId)
+slave_result_t LedSlaveDriver_Update(uint8_t ledDriverId)
 {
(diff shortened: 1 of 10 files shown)


mondalaci:
According to my findings, things are working perfectly with this PR: I couldn't reproduce the oscillation issue, the touchpad pointer movement is buttery smooth, and jump to bootloader works.

That being said, I can't reproduce the oscillation with 9.1.0 either now. What are your findings regarding this PR?

kareltucek:
According to my testing NACKs are gone and jumping to bootloader works. Refresh rate (measured at the level of the slave scheduler) is approximately 14ms, which is an improvement for both smoothness and latency.

Subjective feeling is that the latency might have gone lower (but that may be a placebo effect). Unfortunately, movement seems uneven, and seems to not follow straight lines.

mondalaci:
I think this PR is as good as it can get. The code also looks tight, with a minor exception of an empty else clause. Given that this PR is marked as a work-in-progress, is there anything you want to improve before I merge it?

kareltucek:
Feel free to merge then.

cvkline:
Just downloaded this and it works a treat!  Thanks for sticking with this issue, it looks to have been a nasty one.

Issue #380, opened 2021-06-02
USB XInput interface


mondalaci:
It'd be cool to transform the UHK to a gamepad for which an XInput interface should be implemented. In specific, we should make the trackpoint module configurable to behave like a gamepad joystick. I've exchanged emails with @steamraven and @abcminiuser in this respect, and I'd like to continue the discussion here.

benedekkupper:
So let's get closer to some actual specification of desired functionality here. Which controls of a gamepad are required? Which SW components are producing the data that is to be represented as gamepad actions? What transformation is needed on the data before it is sent over USB?

mondalaci:
@benedekkupper We want to expose every XInput control. Please provide a list of the controls, and we'll move forward from that point.

mondalaci:
@benedekkupper Let me be more specific. For testing the XInput USB interface, we could hardcode several keys to behave as XInput controls, like the keys of the Mouse layer. For example, Mouse+{i, j, k, l} could move the joystick via XInput. Alternatively, we could use the trackpoint module to do the same. We only need the XInput interface exposed in the firmware, and we'll implement the transformation of the data and integration with Agent from that point.

We're happy to assist you to hook to the above actions for testing purposes. It'd be worth starting with a hardcoded mapping scheme for development purposes, but first, we need the list of controls that can be implemented via XInput.

benedekkupper:
So I'll need some clarification on whether we specifically want an XInput simulated, or just an HID gamepad. For a differentiation of features, see [this link](https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput).

**XInput** is a completely custom USB protocol, and requires the entire USB device to act as an XBOX controller. See some details [here](https://www.partsnotincluded.com/understanding-the-xbox-360-wired-controllers-usb-data/) and [here](https://espruino.microco.sm/api/v1/files/11349d480e49ef1d74b3b76f600b016e7a91a0e6.pdf).

We can easily implement an HID device that is picked up by the **DirectInput** driver, for the possible HID usage codes see the [Microsoft reference](https://docs.microsoft.com/en-us/windows/win32/xinput/directinput-and-xusb-devices#gamepad). I believe this link should be your starting reference point to understand what controls can be exposed meaningfully as an HID gamepad. (In theory tools such as [DsHidMini](https://github.com/ViGEm/DsHidMini) could be used to mask the HID interface(s) as an XInput device, but this would require further investigation.)

[...]

PR #535, opened 2022-10-21
Add USB gamepad interface

benedekkupper:
Here is the working branch for the USB gamepad interface (solving #380). I was able to track down the issue pretty quickly once I got the dev board up.

Some technical details:
The gamepad interface is a normal HID one in terms of software design, pretty much like the others. Using Windows' magic, this HID interface is described as an XBOX360 interface instead. The `MsAltEnumMode` flag tells us if the device is plugged in to Windows or not. The HID reports are designed to mirror the XBOX360 report layout, so the way the report is filled out doesn't change. There is a divergence from the official report layout, `USB_GAMEPAD_REPORT_IN_PADDING` is set to 0 - if the report is not recognized on Windows correctly, this needs to be set to 6.

Even though the report layout is fixed in the current state, I have created a `UsbGamepadSetProperty()` method that allows setting the gamepad report properties without considering the layout. There is currently no code that sends any data to the host, so that should be implemented as well.


mondalaci:
Quite a PR!

I've tested it, and my UHK properly enumerates on every OS. dmesg shows the gamepad interface on Linux, and device manager shows the Xbox game controller device on Windows. (I'm clueless about how to check the gamepad on Mac.)

Does the Xbox gamepad only work on Windows, and does the firmware automatically fall back to HID gamepad on other OSes? If so, should we send a kernel commit to having our VID-PID registered on Linux?

Will you implement data sending to the host before I should merge this PR? We'll need to have all the gamepad controls exposed via variables so that we can use them. I wonder if the controls differ on the HID gamepad vs Xbox gamepad.

benedekkupper:
The way it works is this: by default, the gamepad is a standard HID interface. Every host other than Windows will see that. Windows will find its special descriptors, and finds that it can switch the device into alternate enumeration mode. When Windows does this, the gamepad HID interface is swapped with an XBOX360 interface in the config descriptor. The same endpoints are used, and the HID report descriptor is constructed to be equivalent to how the XBOX360 report is laid out. So there is no need to patch the Linux or any other kernel (as you have seen as well, it was already identified correctly).

The data sending itself is there, but since there is no code part that puts any button presses into the reports, so far nothing is sent. As mentioned above, the `UsbGamepadSetProperty()` API is recommended to be used to set the different values in the report - but at the moment using the `usb_gamepad_report_t::X360` fields directly are also a valid way, the former is added to avoid future refactoring in case the report layout changes.

mondalaci:
Thanks, I *think* I understand how enumeration works now.

If the Linux kernel contained our VID-PID, would it treat the UHK as an Xbox gamepad instead of an HID gamepad? Assuming that the Xbox gamepad is more capable in some ways, it could be beneficial.

Can I merge this PR as it is? If we cannot interface with your API, I'll summon you. :)

benedekkupper:
> If the Linux kernel contained our VID-PID, would it treat the UHK as an Xbox gamepad instead of an HID gamepad?

No, because for any other host than Windows, the gamepad will appear as an HID interface.

> Assuming that the Xbox gamepad is more capable in some ways, it could be beneficial.

There is nothing revolutionary about the XBOX360 gamepad interface, in fact it could be considered a downgrade from HID, since all the data elements are completely hardcoded, so it's not possible to add new buttons, or extend the scale of some analog controls. The only benefit of this interface vs HID is that it's clear and unambiguous. HID gamepads have at least 3 different common ways to represent the D-Pad buttons:
1. Using the Buttons / 13-16 usages
2. Using the Generic Desktop / D-Pad up-down-left-right
3. Using the Generic Desktop / Hat switch usage, with 360 degrees split into 8 directions
But there is also widespread confusion about which axes should be used for the analog left and right triggers. So in the case of HID gamepad interface I'm expecting some compatibility issues with certain games, but this seems inevitable as there is really no industry standard for some of these usages. We will need to rely on user feedback and adjust based on that. (Or guide the users to a gamepad remapping daemon they can use. Let's see.)

> Can I merge this PR as it is? If we cannot interface with your API, I'll summon you. :)

Honestly I would get the whole feature integrated before merging this branch (unless you plan to do development on the same code areas), it would be confusing that a gamepad interface shows up without any way to make use of it. But it's up to you, it won't do any harm.

mondalaci:
@benedekkupper I've merged this PR to see whether the presence of the gamepad interface will cause compatibility issues. We'll expose the gamepad controls to users, but it can take a while.

Excellent work!

PR #539, opened 2022-10-31
Fork merge

kareltucek:


mondalaci:
What improvements do the first two commits implement? The comments are rather vague, so I'm unsure.

kareltucek:
In a multiline command, write would terminate at the beginning of next line, rather than the end of its own line, resulting in an extra newline character.

As to the missing memset and missing delay deinitialization, it would result in undefined behaviour in certain contexts (e.g., delay followed by a write would result into dumping entire user configuration on screen). It is rather serious bug. Honestly, I am very surprised that it went unnoticed so long.

PR #552, opened 2022-11-13
Fix scheduling error in call command, and add control bit.

kareltucek:
Fixed scheduling problem in call command. This could make subsequent macros stop responding.

### DIFF
--- a/right/src/debug.h
+++ b/right/src/debug.h
@@ -48,6 +48,12 @@
     // Always show string (no timing logic is applied), but respect slot logic.
     #define SHOW_STRING(V, N) if(CurrentWatch == N) { ShowString(V, N); }
 
+    #define ERR(E) Macros_ReportError(E, NULL, NULL);
+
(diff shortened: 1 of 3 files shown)


mondalaci:
Please always provide a test case for such commits. Otherwise, I can't be sure whether the PR fixes the issue.

kareltucek:
A test case... well...
 
[UserConfig.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/9998218/UserConfig.tar.gz)

Tap left Mod key. Observe keymap switch from INI to QT3. (This calls `call` command, which messes up scheduling state.)
Tap the Mouse key. (This calls `holdKeymapLayer`, which does not finish.)

Expected behaviour: Mouse press, UHK switches to QTM, when released, switches back to QT3
Actual behaviour: Mouse press, UHK switches to QTM, when released, remains there.

(Optionally: invoke `diagnose`: you will see the macro hanging there, apparently not getting called, yet not getting killed either.)

PR #558, opened 2022-11-18
Fix size of values in ifReg* macro commands.

hxv:
Macro registers hold 32-bit values, but macro commands `ifRegEq`, `ifNotRegEq`, `ifRegGt` and `ifRegLt` parse `value` argument as 8-bit – this change should fix commands to use 32-bit values.


kareltucek (approved):
Perfect! Thanks for the fix!

PR #572, opened 2022-12-06
Rename super to gui and control to ctrl, for sake of consistency.

### LINKED ISSUE #566: Cannot set navigation mode for ctrl layer: Unrecognized layer.: ctrl
It looks like there's an issue with setting navigation modes with the ctrl layer.

I tried to have ctrl+touchpad zoom in & out (this is analagous to ctrl + mousewheel for me).

Typing it myself or using the editor I tried:

`set module.touchpad.navigationMode.ctrl zoom`

The LED display shows ERR and the resultant status shown from printStatus is:

`$onInit: 0: 2: Unrecognized layer.: ctrl zoom`

Thing I tried around this issue:

- Create `ctrl` layer (yes, at first I didn't have this layer so I thought this would fix it. It did not)
- Other variants of the command - they seem to work e.g. `set module.touchpad.navigationMode.base zoom` works just fine
- Even the default command for the ctrl layer (`set module.touchpad.navigationMode.ctrl cursor`) yields the same error
- Turning it off and on again (obviously)

I'd be happy to be wrong but this smells like the `.ctrl` part is not being correctly parsed as a layer, especially since in-place editing yields exactly this command.

kareltucek:
This change is mostly about `control` -> ctrl, but when refactoring, I have also noticed the 'super' discrepancy. I think we should prefer a single consistent naming, at least in the macro language. Due to 's' mnemonic being taken by shift, I have changed it to 'gui', which we are already using anyways (`ifGui`, shortcut notation...).

Original syntax is still accepted.

Closes #566.

@mondalaci do you have any strong opinion on the super vs gui? 

Since 'g' has to be used in the shortcut notation and super is more intuitive in agent, I guess we can't fully unify the terminology anyways.

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -122,7 +122,7 @@ The following grammar is supported:
     COMMAND = set leds.enabled BOOLEAN
     COMMAND = set leds.brightness <0-1 multiple of default (FLOAT)>
     COMMAND = set leds.fadeTimeout <minutes to fade after (NUMBER)>
-    COMMAND = set modifierLayerTriggers.{shift|alt|gui|ctrl} {left|right|both}
+    COMMAND = set modifierLayerTriggers.{shift|alt|super|ctrl} {left|right|both}
     CONDITION = {ifShortcut | ifNotShortcut} [IFSHORTCUTFLAGS]* [KEYID]+
     CONDITION = {ifGesture | ifNotGesture} [IFSHORTCUTFLAGS]* [KEYID]+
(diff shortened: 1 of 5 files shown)


mondalaci:
I agree about the `control` to `ctrl` change, but I'd prefer to use `super` instead of `gui` in the macro language to be as consistent with Agent as possible, except for the shortcut notation.

If you insist, I can merge your PR as it is.

kareltucek:
Allright, reverted to the original state of the affairs.

PR #584, opened 2023-01-13
Remove extended commands switch.

kareltucek:
- [ ] Informal documentation needs to get updated. (I assume you prefer to do it yourself.)


mondalaci:
Please ignore `set macroEngine.extendedCommands` instead of throwing an error to minimize support requests and not have to increase the major version.

kareltucek:
Good catch! Done.

PR #606, opened 2023-04-15
Docs: fix U+1f605 explanation

kenhys:
U+1f605 should be "SMILING FACE WITH OPEN MOUTH AND COLD SWEAT", not shrugging person (U+1f937)

See https://util.unicode.org/UnicodeJsps/character.jsp?a=1F605


mondalaci:
Thanks, bug "smiling face with open mouse and cold sweat" is quite a mouthful. Would you rather change the scancode to represent a shrugging person?

kenhys:
ok, fixed now.

PR #621, opened 2023-06-21
Add onLayerChange and onKeymapLayerChange macro events.

### LINKED ISSUE #588: $onLayerChange smart macro event
At the moment, the display shows the active keymap, and layers are indicated with the "mod", "fn" and "mouse" LEDs.

Now that the firmware also has "fn2-fn5" (as well as "shift/alt/ctrl/super") layers, there are no indicators for there layers.

I would like to be able to show different text on the segment display when layers are switched. My use case: the secondary role of my Tab key switches to a layer with a num pad. I would like to display "NUM" when that layer is active.

kareltucek:
Closes #588

### DIFF
--- a/right/src/keymap.c
+++ b/right/src/keymap.c
@@ -1,6 +1,7 @@
 #include "arduino_hid/ConsumerAPI.h"
 #include "arduino_hid/SystemAPI.h"
 #include "keymap.h"
+#include "layer_switcher.h"
 #include "led_display.h"
 #include "ledmap.h"
 #include "config_parser/parse_keymap.h"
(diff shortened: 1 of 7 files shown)


mondalaci:
$onKeymapLayerChange QWR fng:
setLedTxt 300 FN

$onLayerChange md:
setLedTxt 500 MOD

Tapping Mod makes MOD and then FN momentarily appear. Macro event arguments should be validated, and onKeymapChange may be affected, too.

kareltucek:
Good catch! Should be fixed now.

mhantsch:
Documentation for onLayerChange and onKeymapLayerChange macro events seems to be missing in doc-dev/reference-manual.md

PR #622, opened 2023-06-23
Add prepend activateKeyPostponed option.

### LINKED ISSUE #111: Extend activateKeyPostponed [ append | prepend ]
Currently, there is `activateKeyPostponed`, which puts a new virtual key activation at the end of the postponing queue.

It would be good to extend the command with an optional `[ append | prepend ]` argument, which would allow the user to specify whether the activation is to be put at the beginning or end of the queue.

kareltucek:
Closes https://github.com/kareltucek/firmware/issues/111 . I lost original ticket which contained reasonable usecase...


mondalaci:
I'm all for making the macro engine more powerful, but I feel uneasy about merging this PR. A practical use case would put my mind at ease.

kareltucek:
As I wrote above, I lost the original ticket. 

Given that we already have appending, there is no reason to be _more_ concerned about prepending.

mondalaci:
Fair enough. I trust your judgment.

PR #638, opened 2023-07-15
Fix connection of ifSecondary to the secondary role driver.

kareltucek:
Clearly, I didn't do very good work when testing the alphanumeric-friendly strategy macro support...

Steps to reproduce:
- Try `ifPrimary advancedStrategy final holdKey p`, it gives you an error straight away.
- Also, the advanced strategy did not work due to forgotten postponed calls...


kareltucek:
Superseeded by #639

PR #639, opened 2023-07-16
Fix secondary roles for mouse

### LINKED ISSUE #334: [bug] Navigational modes do not react to secondary role layers.
(Does this need any more elaboration?)

Will think about how to deal with it later...

### LINKED ISSUE #476: Module cursor movement does not trigger secondary role.
1. Install the latest Agent:
   https://github.com/UltimateHackingKeyboard/agent/releases/latest
2. Use Agent to update to the latest firmware:
   https://github.com/kareltucek/firmware/releases/tag/v8.10.12.kt.2

I've done everything from the template.

Here's my issue. When I use a key with Esc as primary and MOD as secondary role, modules (trackball, trackpad, and key cluster) behave as I expected in the mod layer, but then after releasing the Esc signal is sent. The problem is that the keyboard behaves halfway, it doesn't switch to a second role completely.

Previously (not sure until when) it didn't work at all, and keys w/ secondary role didn't affect modules' behavior at all. IMHO, it's worse now, because it's inconsistent.

### LINKED ISSUE #565: doubletap to select a word
First i wanted to say that the new touchpad gestures are really fantastic many many thanks for this !

In the previous firmware, i was able to doubletap (single finger) the touchpad to select a word or a line. This seems to no longer work with the latest firmware, is there a trick to enable this again?

kareltucek:
This also includes #638 which I therefore close. These should be 3 separate PRs, but they share a ton of code... ...so read them commit by commits.

Changes:
- Fix `ifPrimary` / `ifSecondary` which was not functional at all after reconnection onto default secondary role mechanism.
- Any module pointer action now triggers secondary roles.
- Artificial delay added between touchpad presses and releases, fixing doubletap-to-select-word for systems which ignore short (<20ms) mouse button releases (e.g., linux). 

Closes #565, Closes #334, Closes #476

### DIFF
--- a/right/src/macros.c
+++ b/right/src/macros.c
@@ -1789,6 +1789,17 @@ static macro_result_t processResolveSecondaryCommand(const char* arg1, const cha
 
 static macro_result_t processIfSecondaryCommand(bool negate, const char* arg, const char* argEnd)
 {
+    secondary_role_strategy_t strategy = SecondaryRoles_Strategy;
+
+    if (TokenMatches(arg, argEnd, "simpleStrategy")) {
+        arg = NextTok(arg, argEnd);
+        strategy = SecondaryRoleStrategy_Simple;
+    }
+    if (TokenMatches(arg, argEnd, "advancedStrategy")) {
+        arg = NextTok(arg, argEnd);
+        strategy = SecondaryRoleStrategy_Advanced;
+    }
+
     if (s->as.currentIfSecondaryConditionPassed) {
         if (s->as.currentConditionPassed) {
             goto conditionPassed;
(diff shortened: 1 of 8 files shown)


likern:
@kareltucek Not sure is this related or not and should I submit a bug.

Could you show how to use `ifPrimary advanced` with CTRL+SHIFT?
Like in screenshots, but for `CTRL+SHIFT`. Cant make it working.

Screenshot from 2023-07-24 22-51-57
Screenshot from 2023-07-24 22-51-44

kareltucek:
Well, if you are trying to use `ifPrimary advancedStrategy` with official 10.0.0, then it does not work. We know it, and this PR fixes it, so you just need to wait for release.

In case you are on a manually flashed master and by CTRL+SHIFT you mean that you want the secondary role to be CTRL+SHIFT, then this should work:

2023-07-24-231901_1229x1010_scrot

likern:
> Well, if you are trying to use `ifPrimary advancedStrategy` with official 10.0.0, then it does not work. We know it, and this PR fixes it, so you just need to wait for release.
> 
> In case you are on a manually flashed master and by CTRL+SHIFT you mean that you want the secondary role to be CTRL+SHIFT, then this should work:
> 
> 2023-07-24-231901_1229x1010_scrot

Got it, thanks. Yes, I'm on release version. Will wait new release version.

PR #640, opened 2023-07-17
Implement overlayLayer, replaceLayer and overlayKeymap macro commands.

### LINKED ISSUE #567: shared layers across keymaps - feature request
while the concept of keymaps is quite a unique feature of UHK firmware, some layers do not really need to stay separate keymaps since they are redundant to some extent. The proposal is to have also shared layers or the option to share a specific layer across all keymaps.

kareltucek:
Closes #567.


mondalaci:
Excellent job, but I'd rather have `dst` instead of `tgt` because the latter doesn't ring a bell when reading it.

kareltucek:
Ok! Will refactor that shortly,!

kareltucek:
Done!

PR #644, opened 2023-07-20
Make fw report a module as connected if it was connected in last 500ms.

### LINKED ISSUE #399: Inaccurate module attached state

kareltucek:
Closes #399.

### DIFF
--- a/right/src/slave_drivers/touchpad_driver.c
+++ b/right/src/slave_drivers/touchpad_driver.c
@@ -100,6 +100,7 @@ int16_t deltaY;
 
 void TouchpadDriver_Init(uint8_t uhkModuleDriverId)
 {
+    phase = 0;
 }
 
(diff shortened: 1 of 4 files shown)


mondalaci:
The shown state is accurate at 500ms, but responsivity feels rather sluggish. Lowering the timeout to 100ms would be nice, but then the flicker kicks in. Can you make it both accurate and quick?

kareltucek:
Well, if flicker kicks in at 100ms, I can drop the timeout to something like 200ms... I don't think I can do much more.

mondalaci:
I've been testing the timeout in 50ms increments and found 350ms to be the smallest value that doesn't produce flicker.

PR #645, opened 2023-07-23
Refactor invertScrollDirection.

### LINKED ISSUE #526: Make the `invertScrollDirection` invert in both x and y axis?
Transfering from https://github.com/kareltucek/firmware/issues/108 :

Currently the `module.MODULEID.invertScrollDirection` only inverts the y axis. Inverting the x axis as well makes it feel more like the so called natural scrolling, when there is overflow on both axes.

@mondalaci, what do you think? (I have no idea about natural scrolling conventions.)

kareltucek:
Let's do it this way. Leave `invertScrollDirection` as is, but add explicit settings for both axes.

Closes #526.

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -97,6 +97,8 @@ The following grammar is supported:
     COMMAND = set module.MODULEID.caretAxisLock BOOLEAN
     COMMAND = set module.MODULEID.swapAxes BOOLEAN
     COMMAND = set module.MODULEID.invertScrollDirection BOOLEAN
+    COMMAND = set module.MODULEID.invertScrollDirectionX BOOLEAN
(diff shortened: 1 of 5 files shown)


mondalaci:
I also prefer explicit X and Y values for clarity.

But the 11 arguments of `processAxisLocking()` read very badly. A struct should be used instead unless you have a better idea.

Also, we should introduce a mechanism to signify to-be-deprecated smart macro elements. We should eventually deprecate `invertScrollDirection` because we're making it redundant, and it hurts the clarity of the language long-term.

kareltucek:
> But the 11 arguments of processAxisLocking() read very badly. A struct should be used instead unless you have a better idea.

Allright!

> Also, we should introduce a mechanism to signify to-be-deprecated smart macro elements.

Well, there is no difference from just marking it as deprecated right away. ..

kareltucek:
Done.

PR #646, opened 2023-07-23
WIP: Add macro control of perKeyRgb backlight.

kareltucek:
Written this almost accidentally before realizing that it doesn't do what I wished it to do...

Still, sounds like a useful command, so either merge it or close it...

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -124,6 +124,7 @@ The following grammar is supported:
     COMMAND = set keymapAction.LAYERID.KEYID ACTION
     COMMAND = set backlight.strategy { functional | constantRgb | perKeyRgb }
     COMMAND = set backlight.constantRgb.rgb <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)><number 0-255 (NUMBER)>
+    COMMAND = set backlight.perKeyRgb.LAYERID.KEYID.rgb <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)><number 0-255 (NUMBER)>
     COMMAND = set leds.enabled BOOLEAN
     COMMAND = set leds.brightness <0-1 multiple of default (FLOAT)>
     COMMAND = set leds.fadeTimeout <seconds to fade after (NUMBER)>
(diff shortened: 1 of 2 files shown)


kareltucek:
Actually, it is a necessary building block to accomplish one of the caps-lock related feature requests, so probably better merge it...

doc-dev/reference-manual.md
```diff
@@ -124,6 +124,7 @@ The following grammar is supported:
     COMMAND = set keymapAction.LAYERID.KEYID ACTION
     COMMAND = set backlight.strategy { functional | constantRgb | perKeyRgb }
     COMMAND = set backlight.constantRgb.rgb <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)><number 0-255 (NUMBER)>
+    COMMAND = set backlight.perKeyRgb.LAYERID.KEYID.rgb <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)><number 0-255 (NUMBER)>
```
mondalaci: Shouldn't there be 3 numbers instead of 4?

kareltucek: Oops. Yes.

kareltucek: Fixed.

mondalaci:
All good now, but I'm wondering about the exact use case. It may be useful to override the key color regardless of the backlighting mode. Currently, the command switches to per-key RGB mode, which I did not expect.

kareltucek:
Well, that was my original idea too, but that would require a bigger refactor, so I left it as-is. Should I invest the time into it?

mondalaci:
I think it'd be worth the extra time as this command would be more generally useable.

kareltucek:
Allright! Happy to do that!

kareltucek:
Just to be clear, I intend to refactor it so that any explicit change to `set backlight.strategy { functional | constantRgb | perKeyRgb }` always resets overrides. An implicit change of strategy will reset overrides only when it actually changes the strategy. I.e., just changing the color of constantRgb won't reset the overrides.

This means that I will need to:
- alter parser so that it is possible to reparse just per-key rgb values
- probably allocate one more byte per action to signify whether a value is overriden
- (add code that resets the overrides appropriately)

kareltucek:
Rebase was needed, so I have opened new PR.

PR #648, opened 2023-07-24
Implement key states

### LINKED ISSUE #585: [Macros] Expose the state of capslock, scroll lock, and numlock to use in macro conditions
I noticed that on Windows OS, when I turned on/off the caps lock on a non-uhk keyboard, the caps lock led icon on the uhk is also turning on/off. Apparently Windows OS updates all the connected keyboards to be in the same state for modifiers like caps lock, regardless on which keyboard they are toggled. We can use this feature to allow non-uhk devices and uhk interact with each other. For example, I wish to toggle the scroll mode of UHK trackball when I turn on the numlock on my macropad. This is already possible in qmk firmware where they provide the led states of scroll lock / num lock / caps lock `host_keyboard_led_state().num_lock`. Can you provide the same so that we can use them in macro conditions. Also is it possible to change the navigation modes of modules using these conditions?

kareltucek:
Closes #585 .

Introduces `$onCapsLockStateChange, ...` smart macro event and `$ifCapsLockOn...` conditions for caps lock, num lock and scroll lock.

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -24,6 +24,10 @@ Macro events allow hooking special behaviour, such as applying specific configur
     $onNumLockStateChange
     $onScrollLockStateChange
 
+Please note that:
+  - under Linux, scroll lock is by disabled by default. As a consequence, the macro event does not trigger.
+  - under MacOS, scroll lock dims the screen but does not toggle scroll lock state. As a consequence, the macro event does not trigger.
+
 I.e., if you want to customize acceleration driver for your trackball module on keymap QWR, create macro named `$onKeymapChange QWR`, with content e.g.:
 
     set module.trackball.baseSpeed 0.5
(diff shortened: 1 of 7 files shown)


mondalaci:
`$onScrollLockStateChange` doesn't trigger for me. Does it for you?

kareltucek:
No, but I believe it is OS limitation (linux here) rather than firmware. 

I can make UHK send scroll lock scancode. I can verify in the OS that it receives a scancode that it recognizes as scroll lock. I can make OS print its scroll lock, num lock and caps lock states into terminal, and I see that receiving the scroll lock does not change it, while for the other two it works as expected.

mondalaci:
I also use Linux and just tried to reproduce the issue on macOS and Windows.

On macOS, scroll lock dims the screen, and `$onScrollLockStateChange` is not triggered.

On Windows, the event is triggered and works as expected.

We should make the above clear in the documentation to prevent false bug reports.

kareltucek:
> We should make the above clear in the documentation to prevent false bug reports.

Good point, and thanks for testing!

kareltucek:
Done.

PR #649, opened 2023-07-24
Allow overriding key rgb.

kareltucek:
In the end, it turns out that resetting the overrides is too tricky. (Now I wonder if, on the contrary, it would be better to make these overrides persist across keymap switches.)

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -128,6 +128,7 @@ The following grammar is supported:
     COMMAND = set keymapAction.LAYERID.KEYID ACTION
     COMMAND = set backlight.strategy { functional | constantRgb | perKeyRgb }
     COMMAND = set backlight.constantRgb.rgb <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)><number 0-255 (NUMBER)>
+    COMMAND = set backlight.keyRgb.LAYERID.KEYID <number 0-255 (NUMBER)> <number 0-255 (NUMBER)> <number 0-255 (NUMBER)>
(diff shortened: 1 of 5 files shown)


mondalaci:
I think this is a niche feature, and using an extra byte per key action is not a good tradeoff. Users could use macro events to refill the colors they desire.

kareltucek:
You mean you want to write ledmap directly? That would significantly limit the usecases. My usecase is that I want to mark keys that have a runtime macro recorded on them. I do so by marking them when the macro gets recorded 
 Admittedly I could introduce new command that checks whether a key has a macro recorded on it... ...bu that would be a very messy solution since I would have to check all keys from a loop.

Admittedly, the tradeoff is questionable, but there are plenty of good counterarguments:

- current implementation of per key rgb takes up three times more space. In comparison that 1 byte is insignificant.
- current implementation of per key rgb is just as wasteful for just as bad reasons, saving just a bit of code clutter and allowing usecases with colour palettes bigger than 256 different colours. *That* I call niche.
- technically just 1 bit per action is actually used, the rest is padding which can be useful in the future for other purposes.
- as you frequently remind me, we do have the memory at the moment, so let's deal with it when we actually run out.

Don't get me wrong, I am not happy about it either, but given above, I think it justified.

mondalaci:
Fair enough, and thanks for elaborating!

In functional backlighting mode, every key gets black except the set key, which I think is incorrect.

Typo: Overriden -> Overridden

kareltucek:
> In functional backlighting mode, every key gets black except the set key, which I think is incorrect.

Ouch. Indeed, wil fix that layer in the afternoon

kareltucek:
Fixed!

PR #652, opened 2023-07-25
Fix some initialization-related bugs.

### LINKED ISSUE #547: Initialization-related bugs
- Macro output in macro events is lost because main update loop (including macro events) runs even before USB starts working.

- If first scancode originates from the macro engine, it is garbled. The reason is late reactive update of `usbBasicKeyboardProtocol`. (Report handling does not know yet that NKRO is used, and treats the memory as the boot protocol instead.)
  - macro reports need to get zeroed too when protocol changes
  - in order not to lose macro output, we need to update `usbBasicKeyboardProtocol` sooner (unlike standard keys, we won't get it entered again)

@benedekkupper / @mondalaci any ideas on how to detect when USB becomes ready?

kareltucek:
Closes #547

### DIFF
--- a/right/src/macros.c
+++ b/right/src/macros.c
@@ -3565,3 +3565,10 @@ bool Macros_MacroHasActiveInstance(macro_index_t macroIdx)
     }
     return false;
 }
+
+void Macros_ResetBasicKeyboardReports()
+{
+    for(uint8_t j = 0; j < MACRO_STATE_POOL_SIZE; j++) {
+        memset(&MacroState[j].ms.macroBasicKeyboardReport, 0, sizeof MacroState[j].ms.macroBasicKeyboardReport);
+    }
+}
--- a/right/src/macros.h
+++ b/right/src/macros.h
@@ -273,6 +273,7 @@
     uint8_t Macros_ParseLayerId(const char* arg1, const char* cmdEnd);
     int32_t Macros_ParseInt(const char *a, const char *aEnd, const char* *parsedTill);
     bool Macros_ParseBoolean(const char *a, const char *aEnd);
+    void Macros_ResetBasicKeyboardReports();
 
 #define WAKE_MACROS_ON_KEYSTATE_CHANGE()  if (Macros_WakeMeOnKeystateChange) { \
                                               Macros_WakedBecauseOfKeystateChange = true; \
--- a/right/src/main.c
+++ b/right/src/main.c
@@ -70,17 +70,6 @@ static void sendFirstReport()
     }
 }
 
-static void waitUntilUsbInitialized()
-{
-    while (ReportDescriptorsTouched < USB_DEVICE_CONFIG_HID && !UsbBasicKeyboard_ProtocolChanged && CurrentTime < 5000) {
-        __WFI();
-    }
-    uint32_t waitUntil = ReportDescriptorsTouched == USB_DEVICE_CONFIG_HID ? CurrentTime + 200 : CurrentTime;
-    while (CurrentTime < waitUntil) {
-        __WFI();
-    }
-}
-
 int main(void)
 {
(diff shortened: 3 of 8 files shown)


kareltucek:
Namely:
- If first scancode produced is from macro system, it is garbled.
- If there is a protocol switch during runtime, we now clear macro states. This cancels all scancodes held via macro system, but prevents garbage from getting outputted.
- Add a delay on UHK bootup. This should hopefully fix processing any write commands in `$onInit`.

mondalaci:
I feel uneasy about an added delay upon boot because speed is a feature. It'd be nice to measure the added delay compared to the master.

There are some conflicts to be resolved.

kareltucek:
With my system, initialization takes approximately 870ms. I guess it will be different with different systems, hence the extra 130ms.

kareltucek:
(Conflicts resolved.)

kareltucek:
(From my side, the delay is optional. In my opinion, this is the best tradeoff, but if you prefer missed $onInit writes over the delay, I am fine with it.)

mondalaci:
Even though on my computer, the actual delay is so small that I can barely tell the difference, I feel uneasy because, on other systems, the difference may be noticeable.

Let me sleep on this.

kareltucek:
Also one more option: add shorter delay (200ms) since those report descriptors are queried.

mondalaci:
How about introducing a new command for this purpose? It's not a clean solution, but this way, an artificial delay wouldn't be introduced upon boot.

kareltucek:
What do you mean by a new command?

($onInit is not parsed at that time yet.)

mondalaci:
> What do you mean by a new command?
> 
> ($onInit is not parsed at that time yet.)

Then your "Add a delay on UHK bootup. This should hopefully fix processing any write commands in $onInit." comment doesn't make sense to me.

kareltucek:
Why so?

The point is exactly that macro engine should not start running until UHK is properly initialized (usb interfaces included), so that when it finally starts running, everything works properly (including all commands present within $onInit).

kareltucek:
Anyway, what new command did you have in mind?

(I think the answer may clear up some of the confusion.)

mondalaci:
How should I reproduce this issue, to begin with? I put `write test` into `$onInit`, and nothing is written even after flashing this PR.

mondalaci:
Correction: I can reproduce this issue according to the above, but this PR doesn't seem to fix it.

kareltucek:
I see. That's yet another issue then.

mondalaci:
Then how can I reproduce the original issue? :)

kareltucek:
Please try now. I have also changed logic of the delay so that it waits just 200 ms after one of the events suggested by benedek takes place. (Should work more reliably and yield smaller lag.)

mondalaci:
It now works in about half of the cases.

I'm fundamentally against using arbitrary timeouts, as they add to the bootup delay, and we can't be certain if they're sufficient on all OSes.

kareltucek:
I also dont like arbitrary timeouts, but this is the best I can offer which still fixes those writes.

If you prefer writes to not work from $onInit then let me know and I will throw the delay out of the PR.

If you choose to take the delay, then please finetune the value on line 78 untilcit works reliably.

mondalaci:
If you throw out the arbitrary delay, how can I test that this PR works compared to master? I'm not sure how to test it.

kareltucek:
Create a macro `holdKey a`, bind it in your default keymap, powercycle UHK, tap that macro.

With this pr, you should see a outputted. With master, you get some garbage instead.

mondalaci:
Please remove the delays, and I'll give it a shot.

kareltucek:
Done.

PR #657, opened 2023-07-30
Refactor leds, handle led text more intelligently, introduce calendar.

kareltucek:
- Introduce discrete calendar for scheduled activities. The calendar can easily be reconnected onto an interrupt-based timer, thus eliminating need of active waiting for events based on repetitive time comparisons. I intend to use it for pretty much everything that needs to wait for something.
- Refactor ledmap functions to always contain `Ledmap_` prefix. (Previous "led update" functions were impossible to distinguish from one another at glance.)
- Fix some led updates which updated all LEDs instead of just backlight.
- Introduce a manager for segment led display:
  - Prevent information from being accidentally overwritten:
    -  ERR now blinks until consumed, instead of being stably lit until anything overwrites it.
    - Runtime macro recording is now indicated by blinking "REC" text instead of (stolen) adaptive mode led
    - Debug texts are now never overwritten by another text.
    -  `setLedTxt` command is now never overwritten by another text (except by debug).


mondalaci:
Excellent features, but I find naming the newly introduced calendar unintuitive. When I read "calendar", I think about days, weeks, months, and years.

An alternative name like "event scheduler" would be a better fit.

kareltucek:
I see.

In czech computer-science jargon, "discrete calendar" is a well established term. Curiously, in english the corresponding term is "discrete-event simulation". I did not know this.

Regarding "event scheduler", I am a bit concerned that both words are already frequently used in the codebase, so while intuitive, it is not exactly unambiguous. But can do that.

mondalaci:
> Regarding "event scheduler", I am a bit concerned that both words are already frequently used in the codebase, so while intuitive, it is not exactly unambiguous. But can do that.

I agree, but I don't have a better idea. If you do, please suggest. Otherwise, feel free to rename it, and I'll review it.

kareltucek:
Here you go.

mondalaci:
When I make an error in `$onInit`, only ERR is displayed instead of oscillating between it and my current abbreviation. Oscillation works if I invoke a non-`$onInit` macro.

kareltucek:
Should be fixed.

mondalaci:
1. I intentionally mess up $onInit and save the configuration.
2. As expected, the display oscillates between ERR and the actual keymap.
3. I fix `$onInit` and save the configuration.
4. ERR still oscillates, and `printStatus` doesn't type anything. I'm unsure what the Right Thing to do is, but clearing ERR upon saving the fixed configuration would probably make more sense to me.

kareltucek:
Fixed.

mondalaci:
`setLedTxt 500 ABC` sets ABC permanently unless I explicitly switch to a keymap first, in which case it's working as expected.

kareltucek:
😖 Well, this was not one of my best pieces of code...

Hopefully, it should be better now.

mondalaci:
> confounded Well, this was not one of my best pieces of code...

It may very well be. It has just taken several iterations. :)

PR #659, opened 2023-07-31
Add ToggleLayer usb command.

### LINKED ISSUE #324: Expose layer toggling via USB

Might this be possible?

kareltucek:
Closes #324

### DIFF
--- a/right/src/usb_commands/usb_command_toggle_layer.c
+++ b/right/src/usb_commands/usb_command_toggle_layer.c
@@ -1,0 +1,19 @@
+#include "layer.h"
+#include "layer_switcher.h"
+#include "usb_protocol_handler.h"
+#include "usb_commands/usb_command_toggle_layer.h"
+#include "keymap.h"
+#include "debug.h"
+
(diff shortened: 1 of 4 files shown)


mondalaci:
Would it be possible rather to invoke a macro command via USB to implement this? It's quite a niche use case, and introducing a dedicated USB command doesn't seem justified.

kareltucek:
Macro engine has a separate layer switching mechanism, therefore I think a usb command that connects to the core implementation is justified, at least at the moment.

mondalaci:
What's the practical difference between the two?

kareltucek:
As far as I remember, layer locked from macro can only be unlocked from a macro.

Admittedly, another solution is to rework the layer switching mechanism..

kareltucek:
But I guess I can do that, and it should eventually be done anyways...

mondalaci:
Sounds great! I'd much prefer the latter solution.

kareltucek:
Ok!

PR #661, opened 2023-08-12
Pull status buffer via usb command

### LINKED ISSUE #626: Exposing smart macro errors to Agent
Manually dumping the status buffer into an editor is cumbersome. It'd be nice if Agent displayed it.

`UsbCommand_GetKeyboardState()` could expose the error flag, and `UsbCommand_GetVariable` could expose the buffer's content to Agent. Then Agent would display the error in an easy-to-notice top bar with a yellow background.

With the introduction of this feature (both on the firmware and Agent side), the firmware should start throwing errors for deprecated commands, such as `resolveSecondary`, so they can be obsoleted in future major firmware versions instead of keeping around deprecated commands forever.

@kareltucek Thoughts?

PS: We should use consistent naming regarding the buffer. In the reference manual, it's called as both the error and status buffer. 
I think the former is a more apt name.

kareltucek:
Closes #626.

### DIFF
--- a/right/src/macros.c
+++ b/right/src/macros.c
@@ -1160,6 +1160,7 @@ static macro_result_t processSwitchKeymapCommand(const char* arg1, const char* c
 /**DEPRECATED**/
 static macro_result_t processSwitchKeymapLayerCommand(const char* arg1, const char* cmdEnd)
 {
+    Macros_ReportWarn("Command deprecated. Please, replace switchKeymapLayer by toggleKeymapLayer or holdKeymapLayer. Or complain on github that you actually need this command.", NULL, NULL);
     uint8_t tmpLayerIdx = Macros_ActiveLayer;
     uint8_t tmpLayerKeymapIdx = CurrentKeymapIndex;
     uint8_t layer = Macros_ParseLayerId(NextTok(arg1, cmdEnd), cmdEnd);
@@ -1178,6 +1179,7 @@ static macro_result_t processSwitchKeymapLayerCommand(const char* arg1, const ch
 /**DEPRECATED**/
 static macro_result_t processSwitchLayerCommand(const char* arg1, const char* cmdEnd)
 {
+    Macros_ReportWarn("Command deprecated. Please, replace switchLayer by toggleLayer or holdLayer. Or complain on github that you actually need this command.", NULL, NULL);
     uint8_t tmpLayerIdx = Macros_ActiveLayer;
     uint8_t tmpLayerKeymapIdx = CurrentKeymapIndex;
     if (TokenMatches(arg1, cmdEnd, "previous")) {
@@ -1810,6 +1812,7 @@ static uint8_t processResolveSecondary(uint16_t timeout1, uint16_t timeout2)
 
 static macro_result_t processResolveSecondaryCommand(const char* arg1, const char* argEnd)
 {
+
     const char* arg2 = NextTok(arg1, argEnd);
     const char* arg3 = NextTok(arg2, argEnd);
     const char* arg4 = NextTok(arg3, argEnd);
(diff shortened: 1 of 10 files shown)


right/src/usb_commands/usb_command_get_variable.c
```diff
@@ -24,5 +25,10 @@ void UsbCommand_GetVariable(void)
         case UsbVariable_UsbReportSemaphore:
             SetUsbTxBufferUint8(1, UsbReportUpdateSemaphore);
             break;
+        case UsbVariable_StatusBuffer:
+            for (uint8_t i = 1; i < 64; i++) {
```
mondalaci: A define, such as `USB_PACKET_MAX_LENGTH` should be used, instead of `64`.

right/src/macros.c
```diff
@@ -460,9 +460,30 @@ static macro_result_t processScrollMouseAction(void)
...
+    static uint16_t readingPos = 0;
+    char res;
+
+    if (readingPos < statusBufferLen) {
+        res = statusBuffer[readingPos++];
+    } else {
+        res = '\0';
+        statusBufferLen = 0;
+        readingPos = 0;
+        SegmentDisplay_DeactivateSlot(SegmentDisplaySlot_Error);
```
mondalaci: We should keep showing the error/warning status on the LED display because Agent may not be in the foreground when the error triggers and it reads the error, so the user may miss the error.

kareltucek:
Allright, should be resolved.

right/src/usb_commands/usb_command_get_device_state.c
```diff
@@ -22,5 +34,6 @@ void UsbCommand_GetKeyboardState(void)
     SetUsbTxBufferUint8(4, MODULE_CONNECTION_STATE(UhkModuleDriverId_LeftModule));
     SetUsbTxBufferUint8(5, MODULE_CONNECTION_STATE(UhkModuleDriverId_RightModule));
     SetUsbTxBufferUint8(6, ActiveLayer | (ActiveLayer != LayerId_Base && !ActiveLayerHeld ? (1 << 7) : 0) ); //Active layer + most significant bit if layer is toggled
+    SetUsbTxBufferUint8(7, getErrorState());
```
mondalaci: `IsEepromBusy` or `MERGE_SENSOR_IS_MERGED` should be composed with `getErrorState()` instead because the device protocol will be optimized later to allow for multiple commands per packet, so commands should be as short as possible.

I'm unsure whether errors and warnings will be extended with additional types depending on which the error state should be shifted with 2 or more bits when composed.

kareltucek:
Done.

mondalaci:
Shouldn't you shift by 2 to not interfere with the merged state?

kareltucek:
Merged state takes just one bit, doesn't it?
```
00000011 << 1 = 00000110      //error state
                00000001      //merge state
              | --------
                00000111

```

mondalaci:
You're right; sorry for the fabulous oversight on my part.

Two minor issues:

```
set module.keycluster.invertScrollDirection 1
set adfa
set asdf
```

yields:

```
Warning at $onInit 0/6: Command deprecated. Please, replace invertScrollDirection by invertScrollDirectionY.
Error at $onInit 0/7: parameter not recognized:: adfa

Error at $onInit 0/8: parameter not recognized:: asdf

```

Ideally, the error buffer shouldn't contain empty lines.

Second, I'm thinking about how to visualize the errors/warnings in Agent. The idea is that error/warning lines will be shown with the danger/warning [Bootstrap colors](https://getbootstrap.com/docs/4.0/components/alerts/).

If so, Agent can look for "Error" vs "Warning" at the beginning of error buffer lines to display them with the appropriate color. I feel slightly bad about this solution because string matching is not exactly a well-defined interface, but other alternatives would probably take too much work. Let me know if you want to suggest an alternative.

kareltucek:
> Ideally, the error buffer shouldn't contain empty lines.

That is an independent bug -> #662 

> If so, Agent can look for "Error" vs "Warning" at the beginning of error buffer lines to display them with the appropriate color. I feel slightly bad about this solution because string matching is not exactly a well-defined interface, but other alternatives would probably take too much work. Let me know if you want to suggest an alternative.

I don't feel bad about it. Every compiler will spill out just a bunch of error lines and let IDEs to string-match them 🤷...

PR #663, opened 2023-08-13
Unify layer switching implementations.

### LINKED ISSUE #324: Expose layer toggling via USB

kareltucek:
Closes #324


kareltucek:
Actually, I think I should give this a bit more thinking.

PR #664, opened 2023-08-14
Unify layer switching implementations.

### LINKED ISSUE #324: Expose layer toggling via USB

kareltucek:
Closes #324.

This finally ensures full interoperability of macro layer switching commands and native switching commands. Most importantly, layers can be untoggled with either of the methods irrespectively of how they were locked. Also, this probably fixes some composition scenarios in which multiple layer activations were stacked on one another via different methods.

### DIFF
--- a/doc-dev/user-guide.md
+++ b/doc-dev/user-guide.md
@@ -190,9 +190,7 @@ There is also a concept of modifiers, which can modify keyboard's or command's b
 
 ### Handling layers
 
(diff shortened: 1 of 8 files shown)


mondalaci:
Works well, and I understand the logic behind the `unToggleLayer` naming, but I think `untoggleLayer` would have been more intuitive. I misspelled it first, and I vaguely recall someone else making such a mistake, too.

Maybe it'd be nice to rename the command eventually, deprecate the old naming, and provide an alias for backward compatibility.

PR #665, opened 2023-08-14
Expose keymap via the GetDeviceState USB command.

### LINKED ISSUE #328: Expose the actual keymap via the GetDeviceState USB command

kareltucek:
Closes #328

### DIFF
--- a/right/src/usb_commands/usb_command_get_device_state.c
+++ b/right/src/usb_commands/usb_command_get_device_state.c
@@ -1,4 +1,5 @@
 #include "fsl_common.h"
+#include "keymap.h"
 #include "segment_display.h"
 #include "usb_commands/usb_command_get_device_state.h"
 #include "usb_protocol_handler.h"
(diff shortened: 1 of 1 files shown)


mondalaci:
Any ideas on how to map keymap indexes to the 3-letter keymap abbreviations? From a usability standpoint, it'd be important.

kareltucek:
I don't see where this is going... Could you elaborate on the point of your question?

mondalaci:
We'll expose an API to the user that (s)he can use. In this case, the user can branch based on the current keymap. But the index of the current keymap doesn't seem to be useful for me. Its abbreviation would be.

kareltucek:
I see.

In that case, abbreviation would indeed be much more practical. I can easily expose the abbreviation instead, either in GetDeviceState command or in a new command. Just let me know which combination you prefer.

I assumed this was for Agent functionality. If the target audience are users, then this sounds like quite a niche usecase to me - do we actually have any users or specific usecases that want/need this?

mondalaci:
There are two use cases.

The major use case is for adaptive mode, for which it's necessary for Agent to retrieve the actual keymap. For this use case, the keymap index is sufficient.

The minor use case is for exposing the current keymap via the API to the user. But now that I think about it, we'll probably expose the whole configuration to the user with which he can map keymap indexes to abbreviations, so this PR looks good.

PR #666, opened 2023-08-15
Implement checksum-based builds.

### LINKED ISSUE #440: MD5 checksum-based firmware updates
For new firmware versions, Agent updates the firmware images of both halves and all modules, even though usually only the firmware image of the right half changes, which is problematic for two reasons:

1. The update process is unnecessarily long, especially due to BusPal, the USB to I2C proxy, which requires re-enumeration over USB.
2. Rarely, some customers brick their module or left keyboard half, which gets returned to us for hardware-flashing. Theoretically, the bootloader should always recover when flashing is incomplete, but practically, for some unknown reason, sometimes it doesn't.

We'll add MD5 checksums to every firmware image, including both halves and all modules. Agent will only flash firmware images whose MD5 checksums differ unless explicitly asked to flash them.

The checksums of all modules will be included in the master (right half) firmware, too, for Agent to display the current master firmware image to them whenever possible. See the [Agent-side specification of this feature](https://github.com/UltimateHackingKeyboard/agent/issues/1663) for details.

Given that the checksums are contained in the firmware images, they affect the checksums of the firmware images, so `make-release.js` will implement a two-step build process. In the first step, the 128-bit MD5 checksums will contain all zeros. In the second step, the MD5 checksums of firmware images of the first step will be used as the MD5 values contained in the final firmware images to be released.

We'll have to extend the device and module protocols to transfer the checksums and expose them to Agent.

kareltucek:
Closes #440


mondalaci:
```
/home/laci/projects/firmware/right/src/usb_commands/usb_command_get_device_property.c: In function 'UsbCommand_GetDeviceProperty':
/home/laci/projects/firmware/right/src/usb_commands/usb_command_get_device_property.c:88:64: error: 'DeviceMD5Checksums' undeclared (first use in this function)
   88 |             Utils_SafeStrCopy(((char*)GenericHidInBuffer) + 1, DeviceMD5Checksums[DEVICE_ID] , MD5_CHECKSUM_LENGTH);
      |                                                                ^~~~~~~~~~~~~~~~~~
```

kareltucek:
Well, have you ran your `./generate-versions-h.js`?

shared/device.h
```diff
@@ -6,12 +6,16 @@
     #define DEVICE_ID_UHK60V1 1
     #define DEVICE_ID_UHK60V2 2
 
+    #define DEVICE_COUNT 2
+
 #ifndef DEVICE_ID
     #error "DEVICE_ID is undefined!"
 #endif
 
+#if (1-DEVICE_ID-1+DEVICE_ID-0 == 0) //ensures that DEVICE_ID is "defined" with a specific value
```
mondalaci: This doesn't read well. What issue do you address here exactly?

kareltucek: It seems that in modules, DEVICE_ID is defined just as `#define DEVICE_ID`, which makes the next line crash when including the file. This line figures that out, as in that case, result is 2 rather than 0, since `--1 = +1`.

I included `device.h` in `versions.h` in order to know `#define DEVICE_COUNT 2`. 

`device.h` is placed in `shared`, so it looks like it is meant to be included even in module code if needed.

Happy to hear your suggestion on what to do with it!

mondalaci:
I think a `device` directory should be created in `shared` similarly to `module`.

`shared` should only contain files that are meant to be used by both devices and modules. Device-related and module-related files should be stored in their respective subdirectories of `shared`, and the includes should be revised accordingly.

kareltucek:
Well, how does that help me in knowing device count in `versions.h`?

It seems like you also propose moving the js-generated md5sum info into two new files under module and device directories. Are you sure you want it that way? (Sounds like an unnecessary complication of the `generate-versions-h.js` code/process.)

mondalaci:
I can see how my previous suggestion is impractical.

It seems to me that `DEVICE_ID` is defined for modules via `Makedefs.mk`. To resolve this issue properly, `DEVICE_ID` shouldn't be defined for modules to begin with.

Second, to be prudent, `device.h` should error out if both or neither of `DEVICE_ID` and `MODULE_ID` is specified; only one of them should be defined.

Finally, in the generated `versions.h`, we could use an `#ifdef DEVICE_ID` so that `DeviceMD5Checksums` would only be included for devices.

kareltucek:
I see. 

Refactored device.h into `device/device.h` and fixed Makefdefs as suggested.

mondalaci:
```
    CC    caret_config.c
In file included from ../src/usb_interfaces/usb_interface_mouse.h:7,
                 from /home/laci/projects/firmware/right/src/usb_descriptors/usb_descriptor_configuration.h:9,
                 from /home/laci/projects/firmware/right/src/usb_composite_device.h:6,
                 from /home/laci/projects/firmware/right/src/key_action.h:9,
                 from /home/laci/projects/firmware/right/src/caret_config.h:6,
                 from /home/laci/projects/firmware/right/src/caret_config.c:1:
../src/usb_descriptors/usb_descriptor_device.h:6:14: fatal error: device/device.h: No such file or directory
    6 |     #include "device/device.h"
      |              ^~~~~~~~~~~~~~~~~
compilation terminated.
```

kareltucek:
Oops! Fixed.

PR #667, opened 2023-08-17
Add untoggleLayer alias.

kareltucek:


mondalaci:
This should work, but I'd deprecate and eventually obsolete the `unToggleLayer` command to keep the language clean.

kareltucek:
I dont think we should break peoples setups just because of minor aesthetics 🤷

mondalaci:
I understand where you're coming from, but imagine an API where every `unCommand` function has an `uncommand` counterpart, and both do the same thing.

I'm not suggesting breaking now, but eventually, we should obsolete deprecated language elements. This should be done in larger sweeps upon certain major version changes to reduce user friction.

kareltucek:
I don't have any trouble imagining that. Many compilers are more permissive than they should be according to C standard.

On the other hand, imagine an API that constantly changes under your hands. (It's called Python, and 3 out of 4 programs written in it don't work exactly because their authors gave up long time ago.)

mondalaci:
By rarely deprecating old language elements, I don't think "the API would constantly change under users' hands".

Don't you want to trigger at least a deprecation warning for `unToggleLayer`?

kareltucek:
Deprecation warning in practice forces the user to go and rewrite all macros that use it (because otherwise he will end up burried under tons of warnings during usage of the keyboard), so I would prefer to avoid that unless there is a good technical reason or a justifiable added value. 

I am fine with switching the documentation to lowercase version, but I would really prefer not to cause all the inconveniencebto people over such a petty thing.

mondalaci:
Please change the documentation to the lowercase version, then.

kareltucek:
Done!

PR #669, opened 2023-08-18
Postponed time fix

kareltucek:
Reported in https://forum.ultimatehackingkeyboard.com/t/ifsecondary-config-to-match-deprecated-resolvesecondary/395/6 

Fixes the problem, and adds a debugging feature for similar scenarios.

Changelog line could be for instance: Fix behavior of the advanced secondary role strategy when rolling over multiple secondary role keys. Latter keys tended to produce false secondary role.

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -64,19 +64,8 @@ The following grammar is supported:
     COMMAND = noOp
     COMMAND = yield
     COMMAND = {exec|call|fork} MACRONAME
-    COMMAND = stopAllMacros
-    COMMAND = statsRuntime
-    COMMAND = statsLayerStack
-    COMMAND = statsPostponerStack
-    COMMAND = statsActiveKeys
-    COMMAND = statsActiveMacros
-    COMMAND = statsRegs
-    COMMAND = statsRecordKeyTiming
     COMMAND = resetTrackpoint
-    COMMAND = diagnose
     COMMAND = printStatus
-    COMMAND = {setStatus  | setStatusPart} <custom text>
-    COMMAND = clearStatus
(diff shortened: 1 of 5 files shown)


kareltucek:
[uhk-firmware-10.1.0-postponed_time_fix.tar.gz](https://github.com/UltimateHackingKeyboard/firmware/files/12384087/uhk-firmware-10.1.0-postponed_time_fix.tar.gz)

mondalaci:
Are you sure that the introduction of `statsRecordKeyTiming` is fully justified?

There is a value to an easy-to-comprehend, small API surface, so we should think twice when adding new commands.

kareltucek:
Well, either that, or having to write that again from scratch next time when I need to debug timing issues.

mondalaci:
I think it'd be worth removing `statsRecordKeyTiming` for now and permanently adding it back if it's useful long-term.

kareltucek:
Given you are concerned about API surface, what about just removing it from the documentation then?

mondalaci:
It'd make practical sense, but I'm also a big fan of transparency when it comes to the documentation as a whole. Are there any other development-related commands? Maybe you could extract them into a separate document.

kareltucek:
Basically the entire `stats*` "namespace".

kareltucek:
Or maybe just creating a new section in the grammar? The rest of the document is meant to be searched anyways.

mondalaci:
You mean moving the grammar to a dedicated page?

kareltucek:
No, I mean as there are 
```
############
#DEPRECATED#
############
```
and REMOVED sections, we could add a 
```
######################
#DEVELOPMENT COMMANDS#
######################
```
section

mondalaci:
Please do so.

kareltucek:
Done.

PR #671, opened 2023-08-21
Validate macros when new config is flashed.

### LINKED ISSUE #668: [Feature proposal] Macro validation
Assume I write a solution that can validate smart macros, same as the config is validated via a dry run. 

It would mean some nonnegligible amount of code clutter.

Do you find it worth the effort and the added maintenance burden (due to the code clutter)?

kareltucek:
Closes #668.

Alright, it is a bit more than 100 lines 😅, but most of it is just simple returns. No devilish algorithms.

### DIFF
--- a/right/src/macro_events.c
+++ b/right/src/macro_events.c
@@ -46,10 +46,12 @@ void MacroEvent_OnInit()
 }
 
 static void startMacroInSlot(macro_index_t macroIndex, uint8_t* slotId) {
-    if (*slotId != 255 && MacroState[*slotId].ms.macroPlaying) {
-        *slotId = Macros_QueueMacro(macroIndex, NULL, *slotId);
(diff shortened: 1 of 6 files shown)


mondalaci:
The added code size is reasonable, but I find variable names, such as `v` and `d`, hard to read, even when they're local. More descriptive names would be much welcome.

kareltucek:
So what names would you prefer? Like a low-camel-case copy of the ultimate value destination? (Like `acceleration` for `module->acceleration = acceleration`?

Could we maybe go with these names? I suspect it will make future refactor of the code much easier, as it allows easier modification via vim macros...

mondalaci:
I like the suggested naming.

kareltucek:
Alright. Done.

PR #684, opened 2023-09-11
Wake from led sleep on non-report actions, refactor activity detection.

kareltucek:
Reported in https://forum.ultimatehackingkeyboard.com/t/setting-led-fade-timeout-back-to-0-doesnt-restore-leds/455

### DIFF
--- a/right/src/usb_report_updater.c
+++ b/right/src/usb_report_updater.c
@@ -30,6 +30,8 @@
 bool TestUsbStack = false;
 static key_action_cached_t actionCache[SLOT_COUNT][MAX_KEY_COUNT_PER_MODULE];
 
+static uint32_t lastActivityTime;
+
 volatile uint8_t UsbReportUpdateSemaphore = 0;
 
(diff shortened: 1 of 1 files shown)


mondalaci:
I can't reproduce this issue. I `set leds.fadeTimeout 1` then `set leds.fadeTimeout 0`, and the LEDs are permanently enabled. Same via the LED settings.

kareltucek:
So your LEDs don't even turn off after 1 second of inactivity?

mondalaci:
They do, but after setting the value to zero, the LEDs turn on, as expected, unlike for the reporting user. In other words, how should I reproduce this issue?

kareltucek:
You must be doing something more than that. 

- Execute `set leds.fadeTimeout 5`.
- Wait 5 seconds till leds go dark.
- Execute `set leds.fadeTimeout 0`. 
- Don't press any other keys, or touch your modules. Leds should still be dark.

PR #686, opened 2023-09-11
Implement set command limits

kareltucek:


mondalaci:
Some test examples would be much appreciated.

kareltucek:
Well...

```
set module.touchpad.caretSpeedDivisor 0
```

Should now not crash the firmware, 

or
 
```
ifShift set leds.brightness ($leds.brightness * 1.5)
set leds.brightness ($leds.brightness / 1.5)
```

Should not run into problems like leds.brightness becoming equal to zero and therefore stopping working.

PR #691, opened 2023-09-12
Split macros.c into multiple files.

kareltucek:
Steps to reproduce:
- open `macros.c`
- observe whopping 4095 lines.


mondalaci:
It'd make sense to put these files into a dedicated `macros` directory under `right/src`. Eventually, a `command` directory would also make sense under `macros` to further break up `macros_commands.c`.

Let me know if you want to implement some of these suggestions before getting this PR merged.

kareltucek:
Yep, will move them into macros directory. 

As for breaking macros_commands.c more, I am not a huge fan of having dozens of super-small files that often contain just one function, so unless I see some reasonable partitioning scheme, I will probably not press breaking macros_commands.c up more.

mondalaci:
Should I merge this now?

kareltucek:
Feel free to. If you do, I will open a new PR later...

mondalaci:
Regarding `macros_commands.c`, I understand that you don't like small files, but in my mind, having a 2440-line file is a sign of bad code hygiene and discourages contributions.

Per-file USB commands also work very well the way I see it.

kareltucek:
> 2440-line file

Well, I definitely agree that it is hardly ideal, and will try to reduce it if I see a reasonable way to do that.

> Per-file USB commands also work very well the way I see it.

I find them hard to navigate.

Also, macros are quite interconnected which means that every cut across the domain exposes more and more things into global namespace, which code-hygiene-wise is also not ideal.

PR #696, opened 2023-09-18
Report line position in errors.

kareltucek:
As agreed in https://github.com/UltimateHackingKeyboard/agent/pull/2082#issuecomment-1723268913


mondalaci:
The reported character position stays unchanged if I prefix `blah` with spaces.

kareltucek:
Solved.

PR #697, opened 2023-09-18
Fix redundant "unprocessed input" warning after `if BOOLEAN` condition.

kareltucek:
Reported in https://forum.ultimatehackingkeyboard.com/t/replay-macro-n-times/311/19?u=kareltucek

- Create this macro:
```
if(false) setVar repeatCount 1
```
- Save your config
- Observe following warning:
```
Warning at tstErr 1/2: Unprocessed input encountered.
> 2 | if(false) setVar repeatCount 1
>   |           ^
```

### DIFF
--- a/right/src/macros/commands.c
+++ b/right/src/macros/commands.c
@@ -769,7 +769,13 @@ static bool processIfLayerCommand(parser_context_t* ctx, bool negate)
 
 static bool processIfCommand(parser_context_t* ctx)
 {
-    return Macros_ConsumeBool(ctx);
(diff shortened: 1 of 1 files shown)


mondalaci:
Agent doesn't display the mentioned warning, only printStatus. Is it a firmware or Agent issue?

kareltucek:
I think it is Agent issue. 

(If I save config using old Agent, dirty flag is reported correctly, which means that new Agent has to be consuming the input but not showing it.)

PR #712, opened 2023-10-16
Add delete navigation mode.

kareltucek:
Someone mentioned this on the forum, and it seems a very nice idea to me, so here we go?

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -188,7 +188,7 @@ The following grammar is supported:
     LABEL = <string identifier>
     MODMASK = [MODMASK]+ | [L|R]{S|C|A|G} | {p|r|h|t} | {s|i|o}
     NAVIGATION_MODE = cursor | scroll | caret | media | zoom | zoomPc | zoomMac | none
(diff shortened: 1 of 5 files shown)


mondalaci:
I think this feature is too niche, and line deletion doesn't work in Visual Studio Code as it's dependent on the shortcuts of the editor, making it potentially problematic.

kareltucek:
This is not a new feature. It is just adding another remappable slot to the navigation mode system ¯\\_(ツ)_/¯. 

As to line deletion, it is that, or leaving the up/down shortcuts empty...

mondalaci:
If another caret mode is justified, a more general name would make better sense to me, like caret2.

But I'm not sure if this feature is justified, to begin with.

kareltucek:
Allright, let's close it then?

mondalaci:
Yes, I prefer to close this one.

As a rule of thumb, I'd only add features to smart macros that clearly provide value to a fair number of users. Otherwise, we're adding complexity that users must absorb, and simplicity is a feature.

Admittedly, the above is not a strict definition, but I think this feature is below the acceptance threshold.

PR #718, opened 2023-10-25
Fix $keyId when used in string and not delimited by white char.

kareltucek:
Steps to reproduce:

- issue `write "$($keyId.a)"`
- observe an error


mondalaci:
After applying this PR, I get the following error:

```
Error at $onInit 1/8/10: Parsing failed, did you mean '$keyId.a)"'?
> 8 | write "$(keyId.a)"
>   |          ^
```

and `eyId.a)` gets written.

kareltucek:
Oops. 

Should be `write "$($keyId.a)`. The error is correct.

PR #721, opened 2023-10-26
Flash adaptive mode led when recording runtime macros.

### LINKED ISSUE #703: Missing triangle icon when recording macros
Until the latest firmware (10.2.0) when one records a runtime macro the triangle icon lit up. Now this icon does not light up and the phrase "REC" shows on the LED display. When there are warnings in ones macros this means that the display alternates between REC, WRN, QWM — so it can be hard to determine when one is actually recording a macro.
Can the triangle be brought by for this use case?

kareltucek:
Closes #703.

### DIFF
--- a/right/src/event_scheduler.c
+++ b/right/src/event_scheduler.c
@@ -2,6 +2,7 @@
 #include "segment_display.h"
 #include "timer.h"
 #include "macros/core.h"
+#include "macro_recorder.h"
 #include "utils.h"
 
 uint32_t times[EventSchedulerEvent_Count] = {};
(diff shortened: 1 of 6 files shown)


mondalaci:
The 600 ms blinking interval is a bit arbitrary, and I think a slightly shorter interval, such as 500 ms would be a bit better, but I've merged it anyway, and feel free to ignore this comment.

kareltucek:
Indeed it is quite arbitrary. But is 500 any less arbitrary? O:-)

Feel free to change it to any value you like ;-).

mhantsch:
> But is 500 any less arbitrary? O:-)

Choose by fair dice roll... ;)

https://xkcd.com/221/
(sorry, could not resist)

PR #724, opened 2023-11-06
Implement touchpad hold continuation timeout.

kareltucek:
As discussed in https://forum.ultimatehackingkeyboard.com/t/touchpad-double-click-to-drag-continuation/595/4

`set module.touchpad.holdContinuationTimeout 200`

Should it be enabled by default?

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -109,6 +109,7 @@ The following grammar is supported:
     COMMAND = set module.MODULEID.invertScrollDirectionY BOOL
     COMMAND = set module.touchpad.pinchZoomDivisor <1-100 (FLOAT)>
     COMMAND = set module.touchpad.pinchZoomMode NAVIGATION_MODE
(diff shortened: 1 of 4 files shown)


mondalaci:
Works great! I'm not sure whether to enable it by default. How do most touchpads work out of the box in this respect? We should meet most users' expectations.

In any case, this setting will be exposed in the dedicated touchpad module configuration page of Agent.

kareltucek:
> How do most touchpads work out of the box in this respect?

No idea. None of the laptops I work with has tap detection at all. (I.e., left click has to be done by a dedicated button.)

mondalaci:
Just merged the PR as it is, as I believe having this option disabled by default is sensible. This option will be exposed via Agent, so it'll be easy to explore.

mhantsch:
> I'm not sure whether to enable it by default. How do most touchpads work out of the box in this respect?

Some of those I know just slowly keep moving the pointer when you reach the end of the touchpad and hold your finger there. I think they call the feature "edge scrolling".

PR #728, opened 2023-11-26
Trackpoint drifts

kareltucek:
Here is some algorithm to reset drifts... 

If you want debug info, then checkout only the first commit - then it will show "RST" on resets on the segment display.

### DIFF
--- a/right/src/debug.c
+++ b/right/src/debug.c
@@ -3,14 +3,12 @@
 
 #ifdef WATCHES
 
-#include "led_display.h"
 #include "timer.h"
 #include "key_states.h"
 #include <limits.h>
 #include "usb_interfaces/usb_interface_basic_keyboard.h"
-#include "macros/core.h"
 #include "macros/status_buffer.h"
-#include "keymap.h"
+#include "segment_display.h"
 
 uint8_t CurrentWatch = 0;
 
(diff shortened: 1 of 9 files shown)


kareltucek:
Sometimes the module seems to stop responding. Not sure why.

mondalaci:
Just started testing this. I'll get back with my findings in a day or two.

mondalaci:
@kareltucek So far, so good, not that I experience drifts often. Usually, once every few days. How about you?

kareltucek:
Well, I can induce them intentionally...

kareltucek:
The more pressing issue is whether the algorithm causes any false positives.

mondalaci:
What happens then? Does your algorithm detect drifts and reset the trackpoint?

kareltucek:
Yes.

mondalaci:
Great news! How often do you usually experience drifts without this fix?

kareltucek:
Rarely, unless I try to induce them on purpose.

(Also this fix has no effect on whether or not I experience drifts. Only it will resolve them after a few seconds of such a drift.)

mondalaci:
I expected a more specific answer than "rarely". :)

I asked to figure out how long should we test your fix before merging it. What do you think?

kareltucek:
I don't think we can test this very thoroughly, so we should probably just release it and wait if anyone complains.

PR #737, opened 2023-12-27
Add new 7.0.0 data model config values.

### LINKED ISSUE #733: Add module and some other settings to user configuration
Instead of `moduleConfigurationCount = ReadCompactLength(buffer);` let's use a `uint8_t` because we'll never exhaust it. Its value has always been zero so far.

Please rewrite `parseModuleConfiguration()` according to the following. (It has never been used so far because Agent serialized an empty module array, so backward compatibility is not an issue.)
- Leave `uint8_t id = ReadUInt8(buffer);` module id as it is.
- Next up, read the 8 LAYERID_BASIC navigation mode values.
- Next up, read `uint8_t modulePropertyCount`.
- Next up, read the module properties, which are key-value pairs. The key is always a `uint8_t`. The length and type of the value is dependent on the key.
  - Make module-independent property keys start from 0 and increment, such as baseSpeed=0, speed=1, ...
  - Make module-dependent property keys start from 255 and decrement, such as pinchZoomDivisor=255, pinchZoomMode=254, ... - Module-dependent properties of different modules can overlap to make the best use of the key id space.

Add the following to the user configuration to positions where they're consistent with already existing properties:
- secondaryRole.*
- mouseKeys.{move|scroll}.axisSkew
- diagonalSpeedCompensation
- doubletapTimeout - Is it the same as doubleTapSwitchLayerTimeout? If so, don't re-add it.
- keystrokeDelay

kareltucek:
Closes #733.

(Untested.)


right/src/config_parser/basic_types.c
```diff
@@ -26,6 +27,16 @@ uint32_t ReadUInt32(config_buffer_t *buffer)
     return uInt32;
 }
 
+float ReadFloat(config_buffer_t *buffer)
```
kareltucek, line 30: Do we want some specific float format here?

right/src/config_parser/basic_types.c
```diff
@@ -26,6 +27,16 @@ uint32_t ReadUInt32(config_buffer_t *buffer)
     return uInt32;
```
kareltucek, line 27: Is the above correct?

My brain says that 0xDDCCBBAA in little endian now gets parsed as 0x0000DDAA.

mondalaci:
Ideally, we shouldn't use floats in the user configuration to begin with. It's fine to use them internally, but it'd make more sense to use `int8_t` or `int16_t` in the user configuration depending on the desired resolution and serialize/deserialize them to/from floats upon writing/reading.

kareltucek:
I am afraid I don't understand. Could you elaborate a bit more?

E.g.:
- Why not floats?
- So am I supposed to create int/uint-encoded logarithmic scales for multiplicative factors, or are you implying that all values have to  be linear (in the sense of having a constant step between any two possible adjecent values)?
- Wouldn't it be a bit confusing to have the same configurations encoded differently in GUI and in macros? (e.g., -127 to 127 uint8 centered around 0 in gui, while in macro, you would be entering a normal float centered around 1 in range `0 <- 1 -> inf`).
- So is Agent supposed to have an interpretation that is semantically similar to the firmware's internal float representation?
- Computing precise exponentials on UHK is darn expensive.

mondalaci:
I mean linear scales.

To give you an example, on the "Mouse key speed" page, one can specify discrete values using the sliders. Setting the initial mouse pointer speed to 1326.234908 allows for unnecessarily fine-grained resolution, and from a usability standpoint, providing a sensible interval and step size per value would make more sense.

I don't know the intricacies of floating-point representation, but I know that some fractional values, such as 0.1 may end up being 0.0999999, which would be wise to avoid by using an integer-based representation. My example may not be valid, but you get the gist of it.

Agent would use per-property intervals and step values to render values.

kareltucek:
Well, sorry, but converting an inherently logarithmic and axis-wise-symmetric scale (like the axis skew) into a linear one, and furthermore making that choice permanent by encoding it into the serialization format sounds like a pretty insane idea to me. 

(If you wonder, the linearity destroys the axis symmetry, making the scale have much finer resolution for one axis than for the other one.)

> I don't know the intricacies of floating-point representation, but I know that some fractional values, such as 0.1 may end up being 0.0999999, which would be wise to avoid by using an integer-based representation. My example may not be valid, but you get the gist of it.

As for this, why not let the GUI handle rounding, stepping, etc? By doing what you propose you are coupling the range and resolution to the serialization format, therefore making them hard to change later as well as confusing to work with.

Anyway, I will not block this further. Are you sure you want what you are asking for?

mondalaci:
Haven't thought of logarithmic settings. On second thought, I agree that we should store them as float. The space-saving would be negligible, conversion would add extra complexity, and Agent can do rounding as suggested.

right/src/config_parser/basic_types.c
mondalaci: I think simply using a float should be fine. Why not use type casting instead of `memcpy()`? I'm unsure about endianness.

kareltucek: `return *(float*)&res;` throws a warning... You are welcome to suggest a better way to cast it.

mondalaci: `return (float)uInt32;` doesn't throw a warning for me.

kareltucek: Yeah, but it also does not do what we want, since it _converts_ the integer into a float, instead of just casting it.

kareltucek: Unless you have another preference, I will change it to:
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstrict-aliasing"
    return *((float*)&uInt32);
#pragma GCC diagnostic pop
```
which disables the warning...

kareltucek:
(Added some fixes.)

right/src/config_parser/basic_types.c
kareltucek: It seems like ReadUInt32 is used nowhere in the project, which explains why this was not noticed earlier...

Fixing it to expect lower endian...

mondalaci:
Please remove `SerializedModuleProperty_CursorAxisLock` because I don't foresee people using it.

Please prefix module-specific properties with module names, such as `SerializedModuleProperty_Touchpad_PinchZoomSpeedDivisor`. It's quite a mouthful, but I'd rather be explicit.

`SerializedModuleProperty_SwapAxes` and `SerializedModuleProperty_InvertScrollDirectionX` only make practical sense for the key cluster, so please modify them accordingly.

Why have `SerializedModuleProperty_InvalidStatePlaceholder`?

kareltucek:
> Why have SerializedModuleProperty_InvalidStatePlaceholder?

Because experience has taught me that it is good to have one value reserved for an undefined state. I am using 255 for this purpose at multiple places in the project - e.g., for macro slot identification, or for layer id in layer switcher... 

Call it a poor man's maybe monad.

mondalaci:
Please elaborate on the usefulness of this pattern because I can't see it.

kareltucek:
Have you ever used NULL? It is basically the same pattern, except for non-pointer type.

mondalaci:
Having NULL at the end of a string or array makes sense to me, but this key-value use case is different, and I can't see its use.

kareltucek:
I don't mean NULL as a zero character, I mean NULL in the sense of an undefined pointer.

kareltucek:
But indeed if need be, it can be allocated somewhere in the middle of the enum... Or as 256, as inmemory representation is not restricted to 8 bits...

mondalaci:
For the life of me, I still can't understand the purpose of this NULLish value. I think it's an unnecessary safeguard, but if you insist on it, an in-memory 256 value is preferred because it doesn't pollute the binary serialization format.

kareltucek:
Nope, just removed it.

right/src/config_parser/parse_config.c
```diff
@@ -194,6 +213,41 @@ parser_error_t ParseConfig(config_buffer_t *buffer)
...
+                default:
+                    return ParserError_InvalidSecondaryRoleActionType;
+            }
+        }
+
+        // Update misc
+
+        if (DataModelMajorVersion >= 7) {
+            DoubleTapSwitchLayerTimeout = doubletapTimeout;
+            DoubletapConditionTimeout = doubletapTimeout;
```
mondalaci: @kareltucek I'm a bit lost. Please explain how the above two variables differ and why they're assigned the same value.

kareltucek: Ah, sorry, forgotten about this.

Well, one controls layer switcher doubletaps. I.e., layer doubletap-to-lock.

The second one controls macro `ifDoubletap` timeout.

Here specifically they are assigned the same value because it is done the same way in set_command.c. 

Generally, as for the original reason, clearly I or we have decided that users will probably not want to control these separately, and therefore simplified the macro api by assigning them from a common configuration variable. 

As for why I keep them separate in the codebase, I don't see a strong reason not to, and probably feel better when each of the separate features has its own abstraction for its parameters.

mondalaci:
If `DoubleTapSwitchLayerTimeout` and `DoubletapConditionTimeout` always contain the same value, I'd love to see one of them removed to minimize redundancy.

We've reviewed this PR with @ert78gb, and he noticed that it's littered with `DataModelMajorVersion >= 7` conditions and suggested that we should rather have a single condition block around the end of the relevant function(s) to simplify the application logic of the firmware and Agent. I agree with him. Going forward, the following structure would be the simplest to use:

```
// We start with generic properties that don't depend on data model versions
...

if (DataModelMajorVersion >= 7) {
    ...
}

if (DataModelMajorVersion >= 8) {
    ...
}

if (DataModelMajorVersion >= 9) {
    ...
}

...
```

If you're fine with it, maybe also the `6` major version equality check could be refactored according to the above structure in the name of long-term consistency.

kareltucek:
> If you're fine with it, maybe also the 6 major version equality check could be refactored according to the above structure in the name of long-term consistency.

I don't see a way to do that while retaining backward compatibility.

If we want to drop backward compatibility, then the check can be just removed.

---------------------

I can easily unify it into two version checks at the expense of mixing order of the properties. That is without voiding the invariant that things are not rewritten during a dry run.

------

As for version 6, it does not respect the dry run property: i.e., rewrites config values even during dry run, which I believe is a bug (at least by definition/specification). Should that be fixed?

mondalaci:
Yes, please fix the dry run bug.

ert78gb:
I saw the new `ReadFloat` method. Does it use little or big edian encoding? Node-js supports both I just have to know which should I use.

thx for the refactoring.

kareltucek:
Little endian, same as ReadUInt16.

kareltucek:
> Yes, please fix the dry run bug.

(Done)

right/src/config_parser/parse_config.c
```diff
@@ -118,6 +104,40 @@ parser_error_t ParseConfig(config_buffer_t *buffer)
...
+    uint16_t secondaryRoles_AdvancedStrategyDoubletapTime = SecondaryRoles_AdvancedStrategyDoubletapTime;
+    uint16_t secondaryRoles_AdvancedStrategyTimeout = SecondaryRoles_AdvancedStrategyTimeout;
+    int16_t secondaryRoles_AdvancedStrategySafetyMargin = SecondaryRoles_AdvancedStrategySafetyMargin;
+    bool secondaryRoles_AdvancedStrategyTriggerByRelease = SecondaryRoles_AdvancedStrategyTriggerByRelease;
+    bool secondaryRoles_AdvancedStrategyDoubletapToPrimary = SecondaryRoles_AdvancedStrategyDoubletapToPrimary;
+    serialized_secondary_role_action_type_t secondaryRoles_AdvancedStrategyTimeoutAction = SerializedSecondaryRoleActionType_Secondary;
+
+    if (DataModelMajorVersion >= 7) {
+        secondaryRoles_Strategy = ReadUInt8(buffer);
+        secondaryRoles_AdvancedStrategyDoubletapTime = ReadUInt16(buffer);
```
ert78gb: @kareltucek  should not it be `secondaryRoles_AdvancedStrategyDoubletapTimeout` ? 
There is timeout on the Agent UI and timeout in the ["documentation"](https://github.com/UltimateHackingKeyboard/agent/issues/1921#issuecomment-1913575443)

kareltucek: Yes.

right/src/config_parser/parse_config.c
```diff
@@ -11,41 +11,22 @@
...
     uint16_t DataModelMinorVersion = 0;
     uint16_t DataModelPatchVersion = 0;
 
     bool PerKeyRgbPresent = false;
 
-static parser_error_t parseModuleConfiguration(config_buffer_t *buffer)
+void readRgbColor(config_buffer_t *buffer, rgb_t* keyActionColors, key_action_color_t keyActionColor)
 {
-    uint8_t id = ReadUInt8(buffer);
-    uint8_t pointerMode = ReadUInt8(buffer);  // move vs scroll
```
ert78gb, line 24: @kareltucek If I read correctly this PR then it drops the old `ModuleConfiguration` logic without any backward compatibility. Agent had not used any data from the old `ModuleConfiguration`. Agent just serialised and deserialised the firmware data.
The default user configuration of the Agent had not contain any default value of `ModuleConfiguration` so if someone reseted the user configuration then it deleted the `ModuleConfiguration`.
Its also mean the factory default does not contains any `ModuleConfiguration`.

Based on the above facts I think I could also drop the old `ModuleConfiguration` code and write the new logic and I don't have to care about the backward compatibility.

ert78gb: I just now talked with Laci and he confirmed I can drop the old code

right/src/config_parser/parse_module_config.c
```diff
@@ -0,0 +1,125 @@
...
+#include "config.h"
+#include "parse_config.h"
+#include "layer.h"
+#include "module.h"
+#include "mouse_controller.h"
+
+static parser_error_t parseNavigationModes(config_buffer_t *buffer, module_configuration_t* moduleConfiguration)
+{
+    // parse navigation modes
+    for (uint8_t layerId = 0; layerId < LayerId_RegularCount; layerId++) {
```
ert78gb, line 14: @kareltucek I am writing the module config parser of the Agent. I am confused because  the `Base` layer id of the Agent is `255`, but the firmware uses `0`
Agent Layer enum definition: https://github.com/UltimateHackingKeyboard/agent/blob/master/packages/uhk-common/src/config-serializer/config-items/layer-name.ts
Firmware Layer enum definition: https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/layer.h#L14

First I did not understand how it had not cause any problem. But I found the [parseLayerId](https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/config_parser/parse_keymap.c#L180) function that has made the transformation. 

The `parseNavigationModes` uses the internal `LayerId` of the firmware. If I implement the above logic then I have to create the same transformation logic which is in the `parseLayerId` function. 

In order to be consistent the firmware should use the transformation logic in the `parseNavigationModes` function too.

What do you think?

kareltucek: I see. I didn't realize this inconsistency.

Well, this will need special handling somewhere anyways, as we clearly don't want to serialize a 256 element array just in order to have base layer at position 255. 

Options I see:
- serialize it as a sequence with base layer first (i.e., to satisfy this code)
- serialize it as a sequence with base layer last (to conform to the order of the agent enum)
- serialize it as a dictionary of key value pairs.

The conceptually cleanest solution seems to be the third one. But from practical standpoint I don't think this is an important decision, and any of the three is equally good. I slightly lean towards leaving it as is - i.e., the first option - but will adapt to any one you should choose.

ert78gb: Before I answer, I would like to ask. Would the `navigationMode` be a structure in the future instead of primitive value?

currently:
```
navigationMode[Layer_Base] = Cursor
```

complex version
```
navigationMode[Layer_Base] = {
  navigationMode: Cursor
  propertyA: Something
}
```

If it could be a complex type then we will serialise it as an array and for flexibility it is worth serialising the `LayerId` as part of the structure. Should we prepare for the complex type version?

```
navigationModes = [
  {
     id: layerId,
     navigationMode: Cursor
     propertyA: Something
  }
]
```


Just now realised does not matter what the structure of the `navigationMode` is/will be, because serialisation, transformation logic should be moved to the Agent side. So the `serialize it as a sequence with base layer first (i.e., to satisfy this code)` version is logical and  the [parseLayerId](https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/config_parser/parse_keymap.c?rgh-link-date=2024-02-11T20%3A44%3A49Z#L180) function needs to be refactored.

mondalaci: I prefer the first option as that'd lead to the cleanest solution from a serialization standpoint.

But I'm curious what may be the reason for the current serialization scheme because maybe I'm missing something.

kareltucek: > Would the navigationMode be a structure in the future instead of primitive value?

I don't see any future need for it to be a structure at this moment.

> Just now realised does not matter what the structure of the navigationMode is/will be, because serialisation, transformation logic should be moved to the Agent side. So the serialize it as a sequence with base layer first (i.e., to satisfy this code) version is logical and the [parseLayerId](https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/config_parser/parse_keymap.c?rgh-link-date=2024-02-11T20%3A44%3A49Z#L180) function needs to be refactored.

I am confused. Do I understand correctly that you propose to make the `LayerId_Base = 0` universal in the serialization format, and therefore basically remove parseLayerId from firmware code? (Or more precisely convert to a backwards-compatible format.)

> But I'm curious what may be the reason for the current serialization scheme because maybe I'm missing something.

I don't understand? 

If you are asking as to the idea behind serialized base layer having id of 255, I am baffled by that too... 

I guess the original reason was that layer ids were used to toggle/hold layers, and thus base layer id was not needed in the enum at all.

ert78gb: I reviewed the PR that introduced the extra layers handling.  https://github.com/UltimateHackingKeyboard/agent/pull/1548/files#diff-2e6e10c7a57591441db51b02684782c9706fe152f7accc8ddac69a78fac02f64R4

The `LayerName` enum already existed without the `base` layer value.
```typescript
export enum LayerName {
     mod,
     fn,
     mouse
 }
```
We added the `base` layer as the last item for backward compatibility.

mondalaci: > I am confused. Do I understand correctly that you propose to make the `LayerId_Base = 0` universal in the serialization format, and therefore basically remove parseLayerId from firmware code?

Yes.

> (Or more precisely convert to a backwards-compatible format.)

Do you mean backwards-incompatible?

> I guess the original reason was that layer ids were used to toggle/hold layers, and thus base layer id was not needed in the enum at all.

I agree that this was probably the reason.

ert78gb: > I am confused. Do I understand correctly that you propose to make the LayerId_Base = 0 universal in the serialization format, and therefore basically remove parseLayerId from firmware code? (Or more precisely convert to a backwards-compatible format.)

Yes

kareltucek: > Do you mean backwards-incompatible? 

Well, the new version does not need explicit conversion at all as the proposed serialized layer id is same as internal firmware layer id. So the only point of leaving there the function at all is for backward-compatibility purposes. So I mean to actually make it handle backward compatibility.

kareltucek: Ok! I will modify the code correspondingly.

kareltucek: Done.

ert78gb:
@kareltucek I am so sorry, I know it was my idea, but I would like to revert the https://github.com/UltimateHackingKeyboard/firmware/pull/737/commits/9832195544ebd15f2b05586eb2ef72cdcca58a35 commit if you don't mind. It is too big change on Agent side and end of the day it is not really necessary for the module configuration.

The Agent side problem is Agent supports json and binary format serialisation. The json format contains the layer name so I need to introduce new enum to support the old 255 base layer Id and the new. I am working on it for a while and the code will really ugly and hard to maintain. The firmware side implementation more elegant

kareltucek:
@ert78gb sure, no problem!

right/src/config_parser/parse_config.c
```diff
@@ -76,19 +57,24 @@ parser_error_t ParseConfig(config_buffer_t *buffer)
...
-        readRgbColor(buffer, KeyActionColor_None);
-        readRgbColor(buffer, KeyActionColor_Scancode);
-        readRgbColor(buffer, KeyActionColor_Modifier);
-        readRgbColor(buffer, KeyActionColor_Shortcut);
-        readRgbColor(buffer, KeyActionColor_SwitchLayer);
-        readRgbColor(buffer, KeyActionColor_SwitchKeymap);
-        readRgbColor(buffer, KeyActionColor_Mouse);
-        readRgbColor(buffer, KeyActionColor_Macro);
+        ledsFadeTimeout = 1000 * ReadUInt16(buffer);
+        perKeyRgbPresent = ReadBool(buffer);
```
ert78gb: @kareltucek I am testing the new user-configuration https://github.com/UltimateHackingKeyboard/agent/pull/2166 of the Agent.
I test with the default user configuration. I can save to the keyboard and read from the keyboard. 
After I switch ON the "Per key backlighting" mode I can't save the user config. The firmware returns with 3 - `ParserError_InvalidSerializedKeyActionType`

Could this `perKeyRgbPresent` renaming cause the issue?

kareltucek: Likely yes. Will look into it on sunday or monday.

Thanks for pointing out and sorry for the complications!

kareltucek: Well, I dont see anything wrong with this piece of code, but the error is likely caused by something not consuming something that should be consumed.

ert78gb: If I read vell the code then 

The appy userfonfiguration process runs first in dry run mode and second in real mode.
The `PerKeyRgbPresent` global variable manages the key actions of the user config contains RGB colors or not.
The default value of the `PerKeyRgbPresent` is false

- the apply user config runs in `dryRun` mode
- here set `perKeyRgbPresent` variable with the userConfig contains RGB colors or not. It is local variable available only in this block
- the `parseKeyActionColor` read the RGB values from the the buffer if `PerKeyRgbPresent` true. In `dryRun` mode the value is always `false` because it set only in real mode. 
- the key action parser reads the first key action from the buffer, but skip the read of RGB data. When the parser reads the 2nd keyaction then the first byte is 255 that is an RGB segment and not a key action type.

I think it is the reason of the error.

kareltucek: Yep, that sounds plausible.

kareltucek: @ert78gb should be fixed!

ert78gb: thx. it works.

mondalaci:
@kareltucek I've started to test the module settings of Agent, and the firmware doesn't respect the selected navigation modes. It always behaves as if "cursor" is set regardless of the actual setting. Please look into it.

kareltucek:
I can't reproduce this, selected navigation modes work fine. 

What I did:

- bump userconfig version to 7.0.0 in package.json in both agent and firmware (Not sure if this step is needed. Firmware will read 7.0.0 config fine, given that agent flashes it. Not sure what conditions are used in Agent.)
- regenerate versions.h with `./generate-versions-h.js --withMd5Sums`
- rebuilt and flashed firmware
- rebuilt agent
- reset configuration
- set mod, fn and mouse layer navigation modes set to caret on trackball module
- tested

mondalaci:
Works well, indeed! I don't know what was the issue on my end.

PR #738, opened 2023-12-27
Fixes #287 Reorder USB Interfaces

### LINKED ISSUE #287: Rearrange USB interfaces for better BIOS compatibility
@xxxajk has just advised that we should change the current USB interface order:

UsbGenericHidCallback
UsbBasicKeyboardCallback
UsbMediaKeyboardCallback
UsbSystemKeyboardCallback
UsbMouseCallback

Let's use the following order instead:

UsbBasicKeyboardCallback
UsbMouseCallback
UsbMediaKeyboardCallback
UsbSystemKeyboardCallback
UsbGenericHidCallback

The assumption is that the USB descriptor parser of the BIOS can claim only the first two interfaces this way instead of having to parse all the interfaces, potentially resulting in better BIOS compatibility.

An unwanted side effect is that Agent is hardcoded to use interface zero as the communication interface which gets renumbered, making Agent unable to communicate with the UHK. To combat this issue, we can change the USB PID, make Agent branch based on the PID, and use the proper interface number. But first, let's implement the above in a PR, and check out if it truly improves BIOS compatibility.

kareltucek:
Closes #287.

Credits of course go to @steamraven .

Note that this also breaks flashing... I suppose that is expected, as I have not applied any patches to the agent submodule.


mondalaci:
@benedekkupper Can you please take a look at the new interface order and confirm that it should result in maximum KVM / BIOS compatibility? Feel free to suggest if you notice any possible improvement.

benedekkupper:
LGTM

kennhardy:
Is this planned to be released in 10.6.0?

mondalaci:
We'll release it in the next firmware version.

mondalaci:
Flashing works fine with the latest Agent.

PR #739, opened 2024-01-03
Implement backlight for v1.

### LINKED ISSUE #569: Backlight macros don't work on UHK 60 v1
Using "set backlight.strategy constantRgb; set backlight.constantRgb.rgb 255 0 0" shows ERR on the lcd screen.
I've an UHK v1 with a cluster module and my intention was to change the cluster lightning depending on the keymap used.

### LINKED ISSUE #655: cannot set rgb backlight on key cluster module with UHK v1
An UHK60v1 with key cluster module.
Agent 3.0.1, firmware 10.1.0.

Backlighting mode configured as 'functional backlighting' in Agent.

I am trying to change the key cluster backlight depending on the layer.

Relevant part of `$onInit:`

```
set backlight.strategy constantRgb
set backlight.constantRgb.rgb 64 64 64
set leds.fadeTimeout 300
```

Default keymap 'CMX'.

Relevant part of `$onKeymapChange CMX`

```
set backlight.keyRgb.base.128 128 128 128
set backlight.keyRgb.base.129 128 128 128
set backlight.keyRgb.base.130 128 128 128
set backlight.keyRgb.mod.128 0 128 0
set backlight.keyRgb.mod.129 0 128 0
set backlight.keyRgb.mod.130 0 128 0
set backlight.keyRgb.mouse.128 0 128 128
set backlight.keyRgb.mouse.129 0 128 128
set backlight.keyRgb.mouse.130 0 128 128
set backlight.keyRgb.fn.128 128 0 0
set backlight.keyRgb.fn.129 128 0 0
set backlight.keyRgb.fn.130 128 0 0
set backlight.keyRgb.fn5.128 128 128 0
set backlight.keyRgb.fn5.129 128 128 0
set backlight.keyRgb.fn5.130 128 128 0
```

I am expecting to see the colour of the key cluster backlight to change when I activate different layers. However, nothing happens. The key cluster just keeps lighting in the same colour all the time.

I also tried to configure per-key backlighting in Agent, and configure different backlight colours for the key cluster keys on each layer. However that also did not achieve anything.

Apparently, rgb backlight control does not work with UHK60 v1, not even on the key cluster (which supports full rgb backlight).

kareltucek:
Closes #569.
Closes #655.

### DIFF
--- a/right/src/ledmap.c
+++ b/right/src/ledmap.c
@@ -291,7 +291,7 @@ static void setPerKeyMonochromatic(const rgb_t* color, uint8_t slotId, uint8_t k
     if (ledMapItem->red == 0 && keyId != 0) {
         return;
     }
-    uint8_t value = ((uint16_t)color->red + color->green + 2*color->blue) / 4;
+    uint8_t value = ((uint16_t)color->red + color->green + color->blue) / 3;
     LedDriverValues[slotId][ledMapItem->red] = value * KeyBacklightBrightness / 255;
 }
 
--- a/right/src/ledmap.c
+++ b/right/src/ledmap.c
@@ -1,3 +1,4 @@
+#include "key_action.h"
 #include "keymap.h"
 #include "layer_switcher.h"
 #include "ledmap.h"
@@ -5,8 +6,10 @@
 #include "device/device.h"
 #include "config_parser/config_globals.h"
 #include "debug.h"
+#include "slot.h"
 
 #define RGB(R, G, B) { .red = (R), .green = (G), .blue = (B)}
+#define MONO(M) { .red = (M) }
 
 rgb_t LedMap_ConstantRGB = RGB(0xFF, 0xFF, 0xFF);
 backlighting_mode_t BacklightingMode = BacklightingMode_Functional;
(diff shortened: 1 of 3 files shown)


right/src/ledmap.c
```diff
@@ -157,70 +285,133 @@ static void setPerKeyRGB(const rgb_t* color, uint8_t slotId, uint8_t keyId)
     LedDriverValues[slotId][ledMapItem->blue] = color->blue * KeyBacklightBrightness / 255;
 }
 
+static void setPerKeyMonochromatic(const rgb_t* color, uint8_t slotId, uint8_t keyId)
+{
+    const rgb_t *ledMapItem = &LedMap[slotId][keyId];
+    if (ledMapItem->red == 0 && keyId != 0) {
+        return;
+    }
+    uint8_t value = ((uint16_t)color->red + color->green + 2*color->blue) / 4;
```
mondalaci: @kareltucek Why double the blue?

kareltucek: It is cheaper to compute, as this can be computed just using binary shifts without any divisions...

mondalaci: I see, but the blue gets disproportionally weighted, isn't it?

kareltucek: Yes, but that does not seem as a big deal, as in practice, I can hardly distinguish lighter keys from darker, let alone trying to guess the original color...

mondalaci: I'm all for optimizing the code, but not at the expense of accurate backlight, even if it's barely visible.

kareltucek: All right...

PR #742, opened 2024-01-10
Initialize .m_data_2 RAM segment.

### LINKED ISSUE #673: Momentary random colors upon booting
When powering up the UHK, random key colors can be seen for a moment. This is slightly disturbing and should be avoided.

kareltucek:
Closes #673 

Steps to reproduce:
- Press the button on the back side to jump into factory keymap.
- Observe that the keymap does not work, including random lighting, as it is initialized randomly.


mondalaci:
Excellent job figuring this out!

PR #749, opened 2024-01-31
Key timing hunt rebased

### LINKED ISSUE #746: False secondary role activations
https://forum.ultimatehackingkeyboard.com/t/double-or-no-tapkey-triggered-randomly-by-ifprimary-ifsecondary/660/31

kareltucek:
Changelog:
- advanced strategy revised: safety-margin-related behavior might have changed
- added `secondaryRole.advanced.triggerByMouse` config value. Advanced secondary roles are now not triggered by mouse movement by default.
- added `secondaryRole.advanced.triggerByPress` config value.

Closes #746 .
Implements https://github.com/UltimateHackingKeyboard/agent/issues/792 .

### DIFF
--- a/right/src/secondary_role_driver.c
+++ b/right/src/secondary_role_driver.c
@@ -111,6 +111,8 @@ uint16_t SecondaryRoles_AdvancedStrategyDoubletapTime = 200;
 uint16_t SecondaryRoles_AdvancedStrategyTimeout = 350;
 int16_t SecondaryRoles_AdvancedStrategySafetyMargin = 50;
 bool SecondaryRoles_AdvancedStrategyTriggerByRelease = true;
+bool SecondaryRoles_AdvancedStrategyTriggerByPress = false;
+bool SecondaryRoles_AdvancedStrategyTriggerByMouse = false;
 bool SecondaryRoles_AdvancedStrategyDoubletapToPrimary = true;
 secondary_role_state_t SecondaryRoles_AdvancedStrategyTimeoutAction = SecondaryRoleState_Secondary;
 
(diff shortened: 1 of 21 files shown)


right/src/utils.c
```diff
@@ -103,3 +112,39 @@ void Utils_SafeStrCopy(char* target, const char* src, uint8_t max) {
     memcpy(target, src, stringlength);
     target[stringlength-1] = '\0';
 }
+
+const char* Utils_KeyAbbreviation(key_state_t* keyState)
+{
+    if (true) {
```
mondalaci: What's the point of this condition if only this branch runs?

right/src/secondary_role_driver.c
```diff
@@ -132,18 +135,26 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
         && resolutionKey == previousResolutionKey
         && resolutionStartTime - previousResolutionTime < SecondaryRoles_AdvancedStrategyDoubletapTime
         ) {
+        KEY_TIMING(KeyTiming_RecordComment(resolutionKey, "PA"));
```
mondalaci, line 138: This function is becoming quite heavy and difficult to comprehend. Is there any way to refactor it to multiple functions?

scripts/timingGraph
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
```
mondalaci, line 1: Please let's use [zx](https://github.com/google/zx) instead of bash for scripts going forward. In my experience, even few-liner bash scripts are a maintenance burden.

kareltucek: Could you elaborate on reasons? 

At a glance, it looks super inconvenient...

right/src/secondary_role_driver.c
kareltucek: I would much rather not do that.

The function is long, but straightforward. Chopping it into multiple function would just make it much harder to navigate.

right/src/utils.c
kareltucek: Clearly to preserve the other branch in case it is needed later.

On second thought, I will remove the other branch.

right/src/secondary_role_driver.c
mondalaci: Fair enough.

mondalaci:
Can I merge this, or do you want to change anything else?

kareltucek:
You can merge it!

mondalaci:
Regarding #746, please summarize the issue and provide the easiest steps to reproduce it. There are a lot of forum comments in the related thread, I get lost, and I have to deal with a lot.

kareltucek:
- Map a lot of secondary roles on alphanumeric keys on base layer.
- Connect your trackball module. 
- Set advanced secondary roles as follows:
```
set secondaryRole.defaultStrategy advanced
set secondaryRole.advanced.timeout 1000
set secondaryRole.advanced.timeoutAction secondary
set secondaryRole.advanced.safetyMargin 0
set secondaryRole.advanced.triggerByRelease 0
```
- This way, no secondary roles should activate unless you press the key for at least a second.
- Now, give it 15 minutes of hard typing. I mean, bang the keys really hard.
- If you are lucky, some secondary role activates for no apparent reason. (The reason is small movement of the ball caused by your typing.)


Alternatively:
- Map some secondary role on your base layer
- Set advanced secondary roles as follows:
```
set secondaryRole.defaultStrategy advanced
set secondaryRole.advanced.timeout 1000
set secondaryRole.advanced.timeoutAction secondary
set secondaryRole.advanced.safetyMargin 0
set secondaryRole.advanced.triggerByRelease 0
```
- Tap the key and simultaneously move the cursor using your module.
- Observe secondary role getting activated. In this case, it might be desired, or not. We operate with the assumption that it is *not* desired. If it is, then you have the option to allow it using the `set secondaryRole.advanced.triggerByMouse 1` option.

PR #752, opened 2024-02-11
Add ifLayerToggled command.

kareltucek:


mondalaci:
What's the use case for this command?

Shouldn't you add a description in the documentation?

kareltucek:
As to description, I think it is quite self descriptive.

As to usecase, I don't think we should discriminate commands based on usecases. It is totally reasonable to want to know if current layer is held or toggled, and it is an information that can't be obtained directly, and that's all that should be sufficient to merge a command that is this general and non-exotic. (I feel quite irritated by this question to be honest as I don't feel like I should need to justify this usecase.)

But to satisfy your curiosity (which *is* totally legit) I am not using doubletap-to-lock on base layer, because it often leads to undesired activations (as I often press a holdLayer, then change my mind, and then change my mind again...)(also I have the case buttons modified to be quiet, which has the unfortunate side-effect of them "bouncing"). 

So the idea is to lock when I doubletap such a hold layer button while *inside* the layer. I.e., hold left-case-button, doubletap right-case-button (which in my case both lead to mouse layer).

```
ifLayer mouse ifDoubletap ifNotLayerToggled final { 
    toggleLayer mouse
    setLedTxt 500 "TGL"
}
ifLayer mouse ifLayerToggled final {
     untoggleLayer
     setLedTxt 500 "UGL"
}
holdLayer mouse
```

mondalaci:
Thanks for the explanation. Usually, I'm interested in the use cases behind new commands.

PR #753, opened 2024-02-12
Try fix warning.

kareltucek:
Compiler throws:

```
/opt/firmware/right/src/buspal/bm_usb/usb_descriptor.c:60:37: warning: expression does not compute the number of elements in this array; element type is 'usb_device_interface_struct_t' {aka 'struct _usb_device_interface_struct'}, not 'usb_device_interfaces_struct_t' {aka 'struct _usb_device_interfaces_struct'}
[-Wsizeof-array-div]
   60 |     sizeof(g_hid_generic_interface) / sizeof(usb_device_interfaces_struct_t),
      |                                     ^
/opt/firmware/right/src/buspal/bm_usb/usb_descriptor.c:60:37: note: add parentheses around the second 'sizeof' to silence this warning
/opt/firmware/right/src/buspal/bm_usb/usb_descriptor.c:45:31: note: array 'g_hid_generic_interface' declared here
   45 | usb_device_interface_struct_t g_hid_generic_interface[] = { {
      |                               ^~~~~~~~~~~~~~~~~~~~~~~
```

Following fix seems to makes sense. UHK seems to work fine like this, but to tell the truth I am not entirely sure that the fix is correct, as I don't see much into the usb machinery.

### DIFF
--- a/right/src/buspal/bm_usb/usb_descriptor.c
+++ b/right/src/buspal/bm_usb/usb_descriptor.c
@@ -57,7 +57,7 @@ usb_device_interfaces_struct_t g_hid_generic_interfaces[USB_HID_GENERIC_INTERFAC
     USB_HID_GENERIC_PROTOCOL,        /* HID generic protocol code */
     USB_HID_GENERIC_INTERFACE_INDEX, /* The interface number of the HID generic */
     g_hid_generic_interface,         /* Interfaces handle */
-    sizeof(g_hid_generic_interface) / sizeof(usb_device_interfaces_struct_t),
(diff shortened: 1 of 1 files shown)


mondalaci:
@benedekkupper Please take a look and let us know if this fix is correct.

benedekkupper:
Noone needs to look into the USB machinery, it's sufficient to look at the code. This line fixes the `count` field of `usb_device_interfaces_struct_t`, which should be set by the array size of what's set at the `interface` field. The interface has `usb_device_interface_struct_t*` type, so does `g_hid_generic_interface`. The fix is correct. (Whether this bug was patched somewhere else to make the whole system operational at the end is another question, one that I cannot answer so easily.)

PR #754, opened 2024-02-12
Add dev docs into firmware release for agent autocomplete.

kareltucek:
reference-manual.md is needed for autocompletion, so adding just as doc/dist is included in case user doesn't have access to the internet.


mondalaci:
How is it needed for autocompletion?

kareltucek:
The Agent uses the grammar from reference-manual.md

PR #774, opened 2024-05-02
Parse keyid abbreviaitons in set keymapAction.

kareltucek:
Reported in https://forum.ultimatehackingkeyboard.com/t/temporarily-change-d-to-delete-with-alt-tab-sequence-without-affecting-alt-d/942/3

Steps to reproduce:
- use `set keymapAction.base.d keystroke delete`
- observe error

### DIFF
--- a/right/src/macros/set_command.c
+++ b/right/src/macros/set_command.c
@@ -648,7 +648,12 @@ static macro_variable_t keymapAction(parser_context_t* ctx, set_command_action_t
 
     ConsumeUntilDot(ctx);
 
-    uint16_t keyId = Macros_ConsumeInt(ctx);
+    uint16_t keyId = Macros_TryConsumeKeyId(ctx);
+
+    if (keyId == 255) {
+        Macros_ReportError("Failed to decode keyid!", ctx->at, ctx->at);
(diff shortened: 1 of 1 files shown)


mondalaci:
Previously, I recall replacing numeric key ids with symbols, but it must have been context-dependent. Would you elaborate on it?

For me, `set keymapAction.base.d keystroke d` is far more intuitive than `set keymapAction.base.81 keystroke d`, and I'm unsurprised someone makes such an error.

kareltucek:
> Previously, I recall replacing numeric key ids with symbols, but it must have been context-dependent. Would you elaborate on it?

I don't understand?

The trouble clearly is just that I had missed this site when replacing parsing code from a generic integer parsing to key ids. (If that's what you are confused about anyways.)

> For me, set keymapAction.base.d keystroke d is far more intuitive than set keymapAction.base.81 keystroke d, and I'm unsurprised someone makes such an error.

It was me who made that "error". And it was not an error - it is correct according to specification. This is not an implementation of a feature request. It is a fix of a bug.

mondalaci:
Gotcha! Thanks for explaining.

PR #776, opened 2024-05-23
Implement config manager.

kareltucek:
Moves all configuration to one central place, and thus allows global reset.

This does not reset component states.


mondalaci:
How should I test this?

What do you mean by "component states"?

kareltucek:
I mean macro engine state (scheduler state, macro states, possible static variables), mouse key state arrays, and other similar things.

mondalaci:
How should I test this PR? I'm unsure.

kareltucek:
Choose a settings that is not a part of the userconfig and does not edit keymap/ledmap in ram. Then write it from a macro (not a macro event). Now save configuration from Agent and test that the setting was resetted.

mondalaci:
Please provide me with a couple of such example settings.

kareltucek:
I am not at computer right now. I guess autoRepeatDelay, autoShiftDelay, chordingDelay, secondaryRole.advanced.triggerByPress, stickyModifiers, modifierLayerTrigger, oneShotTimeout may be of that kind, but please check that against Agent.

mondalaci:
My trackpoint speed settings are disrespected regardless of the module section settings.

kareltucek:
> My trackpoint speed settings are disrespected regardless of the module section settings.

Ouch.

Fixed.

PR #790, opened 2024-07-04
Keyid parser fixes.

kareltucek:
Try:

```
set backlight.keyRgb.base.leftModule.key1 0 0 0
set backlight.keyRgb.base.leftModule.key2 0 0 0
set backlight.keyRgb.base.leftModule.key3 0 0 0
```

Reported at https://forum.ultimatehackingkeyboard.com/t/right-click-killing-the-board/704/18

### DIFF
--- a/right/src/macros/commands.c
+++ b/right/src/macros/commands.c
@@ -1392,17 +1392,18 @@ static bool processIfKeyPendingAtCommand(parser_context_t* ctx, bool negate)
 static bool processIfKeyActiveCommand(parser_context_t* ctx, bool negate)
 {
     uint16_t keyid = Macros_TryConsumeKeyId(ctx);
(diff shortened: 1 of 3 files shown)


mondalaci:
```
/home/laci/projects/firmware/right/src/macros/commands.c: In function 'processIfKeyActiveCommand':
/home/laci/projects/firmware/right/src/macros/commands.c:1397:13: warning: comparison between pointer and integer
 1397 |     if (key == 255) {
```

Mousing is not working anymore (neither with mouse keys nor with modules), and the LEDs are not lit.

kareltucek:
> Mousing is not working anymore (neither with mouse keys nor with modules)

This is strange.

Try now and if problems persist, then please share your UserConfig.

mondalaci:
Works well, thanks!

As for the unexpected issues, the culprit was that I was running a newer user config version.

PR #794, opened 2024-07-25
High-res scrolling support for trackball

### LINKED ISSUE #741: High-resolution scrolling
This is a feature request for high-resolution scrolling via the trackball module.

Because the trackball module is (relatively) high-DPI, it should theoretically be possible for it to provide high-resolution scroll events. Changing `scrollSpeedDivisor` from the default of 8 to 1 sends 8 times the events, and of course the default interpretation by the OS is to scroll 8 times as much for a given physical distance.

Scroll speed can't always be adjusted sufficiently (if it can at all) to account for this; plus simply reducing scroll speed in the OS (when it's an option, that is) might not truly indicate to the OS that a fine-grained level of scrolling is possible; I'm not 100% sure on this but it even at lower scroll speeds it might still default to scrolling in larger discrete steps (e.g. never going lower than one line, even if a view can support scrolling fractions of lines).

AFAICT the "right" way to do this is to use the resolution multiplier control in the USB HID protocol (see, e.g., <https://usb.org/sites/default/files/hut1_13.pdf>). This functionality was developed in part by Microsoft, so it's how Windows does it, and AIUI the Linux support for high-resolution scrolling follows it as well, including some software-side conventions adopted in Windows (specifically, the 120 value assigned to a "standard" scroll detent). FWICT if a resolution multiplier is specified, the Linux kernel will just use that as a divisor for the scroll "amount", and clients that support the corresponding libinput events will handle the scroll accordingly (and there is a legacy event for those that don't).

(The only way I could find to change this value other than through the kernel—which does it via reading the resolution multiplier from the device—is using [libinput-config](https://gitlab.com/warningnonpotablewater/libinput-config), but that works globally and so affects all scroll devices.)

rightaditya:
This PR is for HID-compliant high-resolution scrolling support for the trackball. Closes #741. See [this comment](https://github.com/UltimateHackingKeyboard/firmware/issues/741#issuecomment-2247465574) for notes on testing this on Linux.

## TODO

- [x] Actually handle values in `SetReport` request (will require separating X and Y high-res modes)
- [x] Add exception(s) for Linux's broken (?) `SetReport` requests
- [x] Test on Windows
- [x] Use a single multiplier for both axes (try just making the multiplier global in the report, but if that doesn't work, the logical collections will have to be removed)
- [x] Fix [mouse scroll keys](https://github.com/UltimateHackingKeyboard/firmware/issues/741#issuecomment-2267976306) not working in high res mode

### DIFF
--- a/right/src/mouse_controller.c
+++ b/right/src/mouse_controller.c
@@ -24,6 +24,8 @@ typedef struct {
     float x;
     float y;
     float speed;
+    float scaleX;
+    float scaleY;
     int16_t yInversion;
     int16_t xInversion;
     float speedDivisor;
@@ -362,6 +364,8 @@ static void processAxisLocking(
     float x = args.x;
     float y = args.y;
     float speed = args.speed;
+    float scaleX = args.scaleX;
+    float scaleY = args.scaleY;
     int16_t yInversion = args.yInversion;
     int16_t xInversion = args.xInversion;
     float speedDivisor = args.speedDivisor;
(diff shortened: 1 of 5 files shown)


rightaditya:
I've added support for the high-res mode to be selectively disabled per-axis via a USB SetReport feature request, which enables support for this kind of setting via registry values in Windows. It doesn't play too well with axis locking, though. I can't think of a use case for having high-res scrolling enabled on one axis and not another, and there's really just one physical scroll device here, so IMO it'd make more sense to just have a single multiplier that applies to both wheels. This may require changing the structure of the USB descriptor.

rightaditya:
@mondalaci I figured there was no need to sign the CLA until this PR is considered ready for merging, but I can do so sooner if you need it for the CI (or just to silence the bot, lol).

mondalaci:
@rightaditya Yes, it'd make sense to sign the CLA now. A single multiplier for both axes is sufficient.

rightaditya:
Just made the change and tested it. Now on Windows, disabling high-res scrolling for a single axis via the relevant registry key only disables the OS's handling for high-res values on that axis, so scrolling on that axis becomes very sensitive (but moves in larger chunks). Disabling both axes works as expected. I can't figure who would possibly be fiddling with the registry keys to disable high-res scrolling on a single axis only, so this seems a reasonable compromise. (The two-multiplier case worked better for that case, though was still a bit janky and unnecessarily complicated the axis-locking code.)

This behaviour suggests that the firmware is handling the SetReport and GetReport requests as it should, but unfortunately I can't be 100% sure without inspecting USB packet capture logs and I can't get these on Windows since I only have it installed on a USB drive and packet capture with USBPcap doesn't work when this is the case.

Otherwise I think that the minimal stuff needed for this is all done. I could take a look at USB packet capture logs for macOS if desired to see if perhaps the OS is trying to activate high-res mode in some other way than expected.

mondalaci:
@rightaditya Thanks for the excellent work! We'll test this thoroughly, and @kareltucek may fine-tune the firmware in this respect. We'll follow up.

PR #803, opened 2024-09-06
Get proper device number when number of digits is mixed

csnyder616:
The hidraw numbers on my machine span 7-11. With the alphabetical sorting in the old docs, hidraw9 was being used instead of 11. This commit changes the sorting to be numerical.


kareltucek:
Thanks for the fix!

PR #809, opened 2024-10-03
Backport a numlock state fix.

kareltucek:
This is a backport of the issue 186 from the private repository.

It deals with a num-lock conditions, although I am not sure how, or if at all this is reproducible (after all, compilation is an unstable process), but some users on the forum reported the problem being present with the old firmware.

### DIFF
--- a/right/src/usb_interfaces/usb_interface_basic_keyboard.c
+++ b/right/src/usb_interfaces/usb_interface_basic_keyboard.c
@@ -134,7 +134,7 @@ usb_status_t UsbBasicKeyboardCallback(class_handle_t handle, uint32_t event, voi
 
         case kUSB_DeviceHidEventSetReport: {
             usb_device_hid_report_struct_t *report = (usb_device_hid_report_struct_t*)param;
-            if (report->reportType == USB_DEVICE_HID_REQUEST_GET_REPORT_TYPE_OUPUT && report->reportId == 0 && report->reportLength == sizeof(usbBasicKeyboardOutBuffer)) {
+            if (report->reportType == USB_DEVICE_HID_REQUEST_GET_REPORT_TYPE_OUPUT && report->reportId == 0 && report->reportLength <= sizeof(usbBasicKeyboardOutBuffer)) {
                 LedDisplay_SetIcon(LedDisplayIcon_CapsLock, report->reportBuffer[0] & HID_KEYBOARD_LED_CAPSLOCK);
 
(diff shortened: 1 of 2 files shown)


mondalaci:
I must know how to reproduce this issue. I have a Num Lock key and a Numpad 1 key. When keep pressing these keys alternately, I get 1 and End, as expected.

kareltucek:
The problem depends on memory layout which (as far as I know) may not be stable between different compiler versions.

I didn't manage to reproduce the problem with master branch of this repository. I had encountered it some two months ago when it exhibited in the other repository, and fixed it there, believing that official firmware is not affected.

But then this turned up: https://forum.ultimatehackingkeyboard.com/t/macro-with-alt-codes-behaving-strangely-and-inconsistent-find-out-state-of-numlock/1535/5?u=kareltucek Long story short, rpnfan reports exactly the same problem as the one that I had fixed in the other repository. 

For more info see the forum and the ticket 186 in the repository that must not be named.

mondalaci:
@rpnfan Can you reproduce this issue on your UHK 60? If so, I'd appreciate the reproduction steps and your UHK 60 configuration.

rpnfan:
I am not sure what you mean with "this issue", but currently (Agent 4.2, Firmware 11.1.1) the NumLock state can not be read out, as it should. I am on Win 11, set to use US International. 

My config is attached and here is a place where I use the relevant command:

image

[UHK60-2024-09-29-UserConfiguration.json](https://github.com/user-attachments/files/17262857/UHK60-2024-09-29-UserConfiguration.json)

kareltucek:
Aaah, sorry, I didn't bother to read that command carefully on the forum. 

What you mean and want is:
```
ifNotNumLockOn tapKey numLock
```
Or
```
ifNotNumLockOn {
    set backlight.constantRgb.rgb 0 255 0
    tapKey numLock
}
```

So yes, uhk60 is probably fine.

Still, the request to release new firmware in order to allow `ifHold` stands.

rpnfan:
Yes, what I want is correctly summed up. But the NumLock state is not recognized with the current firmware. So, basically the `ifNotNumLockOn `command not working at all for me, which is critical to use Alt-codes to output characters reliably, without just assuming a specific NumLock-state.

kareltucek:
Have you tried those macros that I posted above? Because that's not what is in your User config. 

There you have:
```
ifNotNumLockOn 
tapKey numLock
```

Which is two separate commands, so the tapKey gets always performed irrespectively of num lock state.

PR #1053, opened 2024-12-22
Fix toggle layer bug.

kareltucek:


kareltucek:
(Steps to reproduce: tap a "toggle layer" native action. Nothing happens.)

PR #1054, opened 2024-12-22
Log rolling fixes.

kareltucek:
Closes #819 .

This ports some logging code from #819.


kareltucek:
Merging since this is a minor thing and there is not much to test...

PR #1062, opened 2024-12-26
Light entire matrix at the beginning of switch test mode.

### LINKED ISSUE #1058: All LEDs must be lit in switch testing mode
```
erase-hardware-config.ts
factory-update-uhk60.ts firmwareDir uhk60v2
write-hardware-config.ts uhk60v2 ansi
set-variable.ts testSwitches 1
```

After executing the above but before pressing any key, the ANSI LShift LED should be enabled, but it's not on the UHK 60.

image

kareltucek:
Closes #1058

### DIFF
--- a/device/src/keyboard/key_scanner.c
+++ b/device/src/keyboard/key_scanner.c
@@ -22,6 +22,7 @@
 #include "attributes.h"
 #include "layouts/key_layout.h"
 #include "layouts/key_layout_80_to_universal.h"
+#include "test_switches.h"
 
 // Thread definitions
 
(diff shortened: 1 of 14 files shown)


mondalaci:
The issue persists. Have you tested it?

kareltucek:
That last commit fixes only one minor issue that I have found while testing. As for the cause of the invalid key count, I am still in the dark.

mondalaci:
Ok, thanks! As soon as this gets fixed, I'll release a new firmware.

kareltucek:
Long story short, try now.

mondalaci:
On my ANSI UHK 60 v2, in switch test mode, when I press the left Shift, the left ISO Shift LED lights up instead of the ANSI LED.

kareltucek:
I can't reproduce it. Have you powercycled the board after writing that hardware config? Hardware config is still reloaded only after power cycling.

mondalaci:
This PR works when power-cycling my UHK after `write-hardware-config.ts uhk60v2 ansi` but before `set-variable.ts testSwitches 1`

Can you reload the hardware config with reasonable effort after writing it? Alternatively, we can provide an ansi/iso parameter to testSwitches, but that would be redundant. The bottom line is that I prefer to make factory flashing fast and avoid unnecessary steps, such as a firmware reboot.

kareltucek:
Well, this reloads the led settings when the config is written. I have not tried to update the other settings, such as the serial number though.

PR #1067, opened 2024-12-27
Fade uhk80 leds gradually.

### LINKED ISSUE #1001: Gradual LED fadeout
Instead of suddenly disabling the key backlight and LED display upon timeout, make them gradually fade out over a second.

The global brightness control feature of the LED drivers and OLED display should be used instead of individual LED control. If they can't be used, let me know so we can discuss the implementation.

kareltucek:
Closes #1001

### DIFF
--- a/device/src/keyboard/leds.c
+++ b/device/src/keyboard/leds.c
@@ -33,6 +33,10 @@ void setLedsCs(bool state)
 
 static volatile bool ledsNeedUpdate = false;
 
+static uint8_t currentScaling = 0;
+static uint8_t scalingFadeStep = 5;
+static uint8_t scalingFadeDelay = 10;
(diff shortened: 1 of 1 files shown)


mondalaci:
Very smooth.

Should we create a separate issue for fading modules? (Which would be blocked by the UHK 80 module firmware update issue.)

kareltucek:
Yes please.

Since leds are controlled over i2c from master. I actually dont think tht module update will be needed

mondalaci:
You're right about I2C.

Also, should I create a fade issue specifically for the UHK 60 as it's not supported by this PR?

kareltucek:
No. uhk60 and modules share the same code I believe.

PR #1071, opened 2025-01-01
Reduce audible noise, and expose related settings for experimentation…

### LINKED ISSUE #976: Minimize the audible noise of the UHK 80
First, let me emphasize that we should wait with this issue because the final product's sound signature will probably differ from the prototypes.

In `IS31FL3743B_DS.pdf` of the datasheets repo, see the "phase delay" and "spread spectrum" options. There are 2 (PHC) * (4 (RNG) * 4 (CLT) + 1 (SSP=0) ) = 34 options according to the datasheet. I tried these settings a while back, and they have made a big difference, and the settings I found to be best are currently implemented in `ledUpdater()` but they should be re-tested for the final product.

`AN-154-EN IS31FL3743B PWM Frequency Options Rev.A.pdf` features the PWM Frequency setting, which has 7 options. I haven't tried this setting yet.

The optimal settings may differ per keyboard half.

mondalaci:
… via the device protocol.

Closes #976.

### DIFF
--- a/device/src/keyboard/leds.c
+++ b/device/src/keyboard/leds.c
@@ -78,13 +78,51 @@ static void recalculateScaling() {
     }
 }
 
+void UpdateLedAudioRegisters(uint8_t phaseDelay, uint8_t spreadSpectrum, uint8_t pwmFrequency) {
+    k_mutex_lock(&SpiMutex, K_FOREVER);
+
+    // Set phase delay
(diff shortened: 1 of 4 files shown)


mondalaci:
@kareltucek Please help make this build.

mondalaci:
Thanks, nice catch! Totally missed it's a dongle-related build error.

Issue #1051, opened 2024-12-20
Simplify BLE pairing


mondalaci:
Consider the current pairing popup on Android:

![image](https://github.com/user-attachments/assets/df52d759-8afb-423f-9cbe-4d2b2cbae070)

After typing the pairing code on the UHK, I must tap the Pair button.

On some other keyboards, this pairing popup doesn't contain a Pair button but only a Cancel button. When the pairing code is entered, the popup disappears, and the pairing succeeds. The UHK should follow suit.

mondalaci:
@benedekkupper Can you look into this? Users keep contacting us, and we keep telling them they must hit the Pair/Connect button on various OSes, which shouldn't be necessary if this issue was implemented.

mhantsch:
I can't even hit a pair button because none appears. I can't pair the UHK80 to my Linux laptop unless I resort to some commandline foo, which also only works sporadically.

benedekkupper:
Fixed via #1102

PR #1102, opened 2025-01-25
bt: use correct zephyr callbacks for passkey entry (#1051)

benedekkupper:
This is a draft so far, I couldn't even test it as for some reason the right half didn't advertise itself. The point here is using the `passkey_entry` callback instead of `passkey_display` and `passkey_confirm`, so the BT IO capabilities are reported to the other side differently. Pairing will simply complete by successfully entering the challenge passkey (that is actually not known to the keyboard), therefore the Pair/Connect confirmation on the other end is removed. @kareltucek please review the code, especially the interactions between the BT side and the screen.

### DIFF
--- a/device/src/bt_conn.c
+++ b/device/src/bt_conn.c
@@ -28,6 +28,7 @@
 #include "config_manager.h"
 #include "zephyr/kernel.h"
 #include <zephyr/bluetooth/gatt.h>
+#include "stubs.h"
 
 bool Bt_NewPairedDevice = false;
 
--- a/device/src/bt_conn.c
+++ b/device/src/bt_conn.c
@@ -613,12 +613,7 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
 
 // Auth callbacks
 
-static void auth_passkey_display(struct bt_conn *conn, unsigned int passkey)
-{
-    printk("Passkey for %s: %06u\n", GetPeerStringByConn(conn), passkey);
-}
-
-static void auth_passkey_confirm(struct bt_conn *conn, unsigned int passkey) {
+static void auth_passkey_entry(struct bt_conn *conn) {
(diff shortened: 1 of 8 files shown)


mondalaci:
I've tested this PR, and it works great.

Maybe it's worth mentioning that it takes about 2-3 seconds for the pairing screen to disappear on my phone after entering the pairing code on the UHK, which feels a bit sluggish. Is this a BLE limitation, or can it be reduced, @benedekkupper?

benedekkupper:
> Maybe it's worth mentioning that it takes about 2-3 seconds for the pairing screen to disappear on my phone after entering the pairing code on the UHK, which feels a bit sluggish. Is this a BLE limitation, or can it be reduced, @benedekkupper?

I don't have an obvious answer to this, we send the BT passkey as soon as the 6 digits are entered (which in itself is a bit strange, not having to press Enter, meaning that the last digit cannot be corrected, unlike the rest). What's your experience with other similar BLE keyboards?

mondalaci:
@benedekkupper I vaguely remember other keyboards not expecting an explicit enter at the end of the pairing code. In practice, I think it's not a problem because the user can retry pairing in the unlikely case the last digit is misspelled.

@kareltucek Does this PR look good to you? If so, I'd merge it.

benedekkupper:
Do not merge yet, not all targets are building correctly.

mondalaci:
@benedekkupper Do you plan to fix the build issues or expect us to fix them?

kareltucek:
I will fix that one

mhantsch:
Looking forward to trying this change on my Linux system (which currently refuses to show a passkey entry screen at all with the UHK80).

kareltucek:
Yep, looks good!

kareltucek (approved):

PR #1122, opened 2025-02-05
BLE advertising simplification

benedekkupper:
left uses directed advertising, right advertises HID only


device/src/bt_advertise.c
```diff
@@ -25,107 +24,54 @@
...
     if (DEVICE_IS_UHK80_RIGHT) {
+        bt_le_filter_accept_list_clear();
         if (BtConn_UnusedPeripheralConnectionCount() <= 1 && SelectedHostConnectionId != ConnectionId_Invalid) {
             bt_le_filter_accept_list_add(&HostConnection(SelectedHostConnectionId)->bleAddress);
-            advertisementParams->options = BT_LE_ADV_OPT_FILTER_CONN;
-        } else {
-            advertisementParams->options = BT_LE_ADV_OPT_NONE;
         }
-    }
-}
```
benedekkupper: @kareltucek what is the intention of this `setFilters` method? What I take away from the documentation is that `bt_le_filter_accept_list` API is for scanning rather than advertising. Also, `advertisementParams` variable is modified, but never used.

kareltucek: Ah 🤦.

Assume the user wants to connect to a specific device (dongle or ble hid) that is not yet connected.

The filter was supposed to make sure that only that one device (identified by `&HostConnection(SelectedHostConnectionId)->bleAddress` address) can connect.

benedekkupper: In the advertising role this can be accomplished with directed advertisement (see below for the left half), but according to the zephyr documentation this is not allowed in the direction of Apple BLE central devices, so we could only do it towards dongles. Do you think that would make sense, and if yes, how would you do it?

Did you ever test this `bt_le_filter_accept_list_` calls, if they have an effect?

kareltucek: > Do you think that would make sense, and if yes

Yes. 

When a device tries to connect anyways, I generally refuse their connection attempts in the connected callback. Not sure if this applies for ble hid devices though, as they like to change their ble address depending on security level.

> , how would you do it?

Well, I guess this means different adavertisement depending on whether  `Connections_Type(SelectedHostConnectionId) == ConnectionType_BtHid`

If yes then advertise normally, if no, then advertise direct?

> Did you ever test this bt_le_filter_accept_list_ calls, if they have an effect?

I have tested the scenario. Indeed the accept_list didn't work as good as desired. I assumed that the advertisement acted correctly, but paired centrals tried to connect even without advertisement. (I should have dug deeper, but didn't.)

mondalaci:
@kareltucek I've just tested this PR, and it removes the extra bottom entry that was exposed by the left half:

image

Can I merge this PR, or is there anything else that I should test?

device/src/bt_advertise.c
benedekkupper: > If yes then advertise normally, if no, then advertise direct?

The direct advertisement would go towards the dongle then, right? How do I get the dongle's address?

kareltucek: `HostConnection(SelectedHostConnectionId)->bleAddress`.

(And yes.)

You can check that it is indeed a dongle via the Connections_Type mentioned above. Not sure if it is needed in this context.

kareltucek:
> @kareltucek I've just tested this PR, and it removes the extra bottom entry that was exposed by the left half:

Please wait till Benedek says that everything is ready.

You should test the case of switching to a device via "switchHost name" when the host is not connected (e.g., dongle is violet) and all peripheral connection slots are taken (e.g., there are already three blue/green dongles) - feel free to reduce the peripheral connection count for testing, and independently for production too (it has its cons as well as pros).

device/src/bt_advertise.c
benedekkupper: Thanks, I have updated the code, please review.

mondalaci:
When testing this branch, the UHK 80 BLE device no longer appears on my phone in the available devices list.

mondalaci:
Now, I can pair my UHK 80 to my Android phone, and the connection is established. However, the USB host connection remains, and instead of automatically switching, I have to explicitly switch to the BLE host connection.

@kareltucek Assuming it's due to the firmware logic, can you look into this?

device/src/bt_advertise.c
benedekkupper: I reverted this due to mondalaci's feedback, for now the right is always general advertising

kareltucek:
@mondalaci I can't reproduce this.

Does this happen wih a "new bluetooth device", or a host-connections-list-present device? In the latter case, is your switchover checked?

mondalaci:
I could reproduce the issue again, then switched to master, and the issue was still there, so this branch is probably not the culprit.

This happens (or I'd rather say not happens, that is, the automatic switchover) directly after pairing a "new bluetooth device"). I didn't save the host connections.

Does this log help?

```
Starting advertising, try 0!
NUS and HID advertising continued
Received passkey pairing inquiry.
Type `uhk passkey xxxxxx` to pair, or `uhk passkey -1` to reject
Sending passkey to conn 0x2000bce0
Bt connection secured: unknown (ac:d6:18:1b:3d:b5), level 4. It is auth_conn, so ignoring.
Pairing completed: unknown (ac:d6:18:1b:3d:b5), bonded 1
Con state: Host2(My Phone, Connected), Peer host1
Bt connected to host1 (ac:d6:18:1b:3d:b5)
Established HID connection with host1 (ac:d6:18:1b:3d:b5)
Con state: Host2(My Phone, Ready), Peer host1
Starting advertising, try 0!
NUS advertising continued
host1 (ac:d6:18:1b:3d:b5) conn params: interval=7 ms, latency=0, timeout=5000 ms
host1 (ac:d6:18:1b:3d:b5) conn params: interval=11 ms, latency=0, timeout=1000 ms
I: starting HID app 1: 1
```

kareltucek:
Well, the log looks like UHK knows your "My Phone" already. In that case it behaves according to the switchover checkbox.

mondalaci:
You're right, my oversight! This PR looks good to me, and I'd merge it. Any objections?

kareltucek:
> Any objections?

Yes.

device/src/bt_advertise.c
```diff
@@ -25,107 +24,51 @@
...
-    uint8_t connectedHids = 0;
-    for (uint8_t peerId = PeerIdFirstHost; peerId <= PeerIdLastHost; peerId++) {
-        if (Peers[peerId].conn && Connections_Type(Peers[peerId].connectionId) == ConnectionType_BtHid) {
-            connectedHids++;
-        }
-    }
-    return connectedHids;
-}
-
-uint8_t BtAdvertise_Type() {
```
kareltucek, line 109: This doesn't look like something that should have been removed.

- You are always trying to advertise, leading to errors when the peripheral connection count is reached. My code in turn tries to remedy the situation by increasingle aggressive strategies, including disconnecting current connections. The code was written so that uhk works even with `CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT=1`. With this settings and this code uhk will not work at all over bluetooth (dongle or hid).
- Switchover scenario might need to be taken care of when the limit is reached.

How exactly does the current "simplified" advertising work on right half with respect to dongles? It broadcasts itself as a HID device, and dongles see this and try to connect despite there not being any advertisement for NUS?

I assume I should restore the necessary functionality? Will look into it tomorrow...

benedekkupper: 1. Only the content of the advertising is changed, the call sites of these methods control when advertising is started or stopped, I didn't change the logic there.
2. Although I didn't test the changes with a dongle, I did check the scanning logic, and there I saw no filtering based on service UUIDs, only an address based filtering, that's why I was brave enough to remove it. My original idea of directed advertising when intending to connect to a dongle didn't work - according to mondalaci's feedback - so you are very welcome to take a look at this yourself, or to revert the changes for the right half and continue using HID / NUS advertising content.

kareltucek: > 1. Only the content of the advertising is changed, the call sites of these methods control when advertising is started or stopped, I didn't change the logic there.

The trouble is the BtAdvertise_Type() also decides whether or not to advertise at all. Anyways no worries, I will fix that shortly.

> 2. I did check the scanning logic, and there I saw no filtering based on service UUIDs, only an address based filtering, that's why I was brave enough to remove it. 

I see. I like the change, but this really needs to be mentioned in a comment somewhere ;-). (Will add that.)

kareltucek: Actually, the compatibility is single directional.

When one ble HID is already connected we want to advertise NUS only - as we can't handle multiple ble HIDs at the same time. Unless you or @mondalaci have a better idea how to handle the situation that is.

mondalaci: @kareltucek If it's possible to reuse the single available BLE HID connection upon connecting a new BLE HID device, ongoing BLE HID advertising would be preferred.

device/src/bt_advertise.c
```diff
@@ -25,107 +24,51 @@
...
-}
+        adv_param = *BT_LE_ADV_CONN_ONE_TIME;
 
-uint8_t BtAdvertise_Start(uint8_t adv_type)
-{
-    setFilters();
+        err = bt_le_adv_start(&adv_param, adHid, ARRAY_SIZE(adHid), sdHid, ARRAY_SIZE(sdHid));
+
+    } else if (DEVICE_IS_UHK80_LEFT) {
+        adv_param = *BT_LE_ADV_CONN_DIR_LOW_DUTY(&HostConnection(ConnectionId_NusClientRight)->bleAddress);
```
kareltucek: Btw., here should be `&Peers[PeerIdRight].addr` - this couldn't have worked (@mondalaci i.e., I believe the left half does never work over nus with this branch). 

What's worse is that the code doesn't work when given correct addresses (they start to scan and advertise, but never actually connect). I take the above as an evidence that it wasn't ever tested.

-> I will restore it to the original NUS advertisement, but leave the subsystem in place with a comment. Feel free to try to fix it @benedekkupper .

kareltucek (commented):
Argh, sorry, wanted to "add a single comment".

kareltucek:
~~@benedekkupper given your prj.conf refactor, where should following snippet be placed? (It is relevant for right half only. They are now in nrf_shared.conf)~~

Nevermind, I have added new right.conf

```
# Todo: place these where they belong
# config for right half host switching
CONFIG_BT_MAX_CONN=4
CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT=3
```

-------

@mondalaci when testing this branch, please note that macro `switchHost "name"` scenarios should be (also) tested with CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT set to 1.

device/src/bt_advertise.c
```diff
@@ -106,26 +105,51 @@ static uint8_t connectedHidCount() {
     return connectedHids;
 }
 
-uint8_t BtAdvertise_Type() {
+#define ADVERTISEMENT(TYPE) ((adv_config_t) { .advType = TYPE })
+#define ADVERTISEMENT_DIRECT_NUS(ADDR) ((adv_config_t) { .advType = ADVERTISE_DIRECTED_NUS, .addr = ADDR })
+
+adv_config_t BtAdvertise_Config() {
```
kareltucek, line 129: (An additional reason to leave this in place is that it documents the scenarios that can happen, and makes logs more explicit - nevermind what advertisement params the backend actually ends up using.)

kareltucek:
Btw, 363328db067741cfc940519745054bab8e76ae30 broke dongle leds. 

051e7ea507100089e78dce9b674ef343e20e812c fixes it. I assume it has to do with weak linking and its order, and I guess implies that it is not as safe and reliable as I though.

device/src/bt_advertise.c
benedekkupper: Sorry for all the mess I made. I guess the missing line is `advParam.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;` before calling the `bt_le_adv_start`

kareltucek: Unfortunately still doesn't work. Left advertises, right scans, but they never connect.

---

Also tried this, based on comments in the code:
```
            advParam = *BT_LE_ADV_CONN_ONE_TIME;
            advParam.peer = advConfig.addr;
```
This results in error 60 (timeout).

(Pushing a minor refactor to make the code a bit more bearable.)

benedekkupper:
> ~@benedekkupper given your prj.conf refactor, where should following snippet be placed? (It is relevant for right half only. They are now in nrf_shared.conf)~
> 
> Nevermind, I have added new right.conf
> 
> ```
> # Todo: place these where they belong
> # config for right half host switching
> CONFIG_BT_MAX_CONN=4
> CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT=3
> ```

You could put this to ble_hid.conf, since only the right side is a BLE HID device.

Issue #974, opened 2024-01-07
Charging status and percentage


mondalaci:
The charging status and percentage of both halves must be displayed on the OLED display.

### Charging status

There are 3 charging statuses: charging, not charging, and battery unavailable (as the UHK 80 will be available without batteries via the webshop).

`chargerStatCallback()` should be used to update the charging state. if `gpio_pin_get_dt(&chargerStatDt)` is 0, charging is in progress; otherwise, the keyboard half is not being charged.

If the battery is absent, `chargerStatCallback()` is called twice per second with `gpio_pin_get_dt(&chargerStatDt)` alternating between 0 and 1.

~The right half has a [hardware bug I'm trying to fix](https://github.com/UltimateHackingKeyboard/firmware/issues/852) that causes the `chargerStatCallback()` to get called gazillions of times per second with a 0 stat value. Let's be aware of this bug, but don't work around it, as I'll fix it on the hardware level.~

Let's keep charging (chargerEnDt) always enabled for now. Later, we can implement a battery conservation mode by disabling charging above 70 or 80 percent.

### Charging percentage

See `cmd_uhk_charger()` for the relevant code.

1. This is the raw ADC value of battery voltage:

```
        err = adc_read(adc_channel.dev, &sequence);
        int32_t val_mv = (int32_t)buf;
```

Measure and monitor this value in a thread at a reasonable interval, such as 5 or 10 seconds. Whenever it changes, make the left half send it to the right half. The right half should process the raw ADC battery measurements as follows.

2. Convert the raw ADC measurement to mV using adc_raw_to_millivolts_dt()

3. The mV value must be converted to percentage, which is a logarithmic-shaped curve. I've made several measurements with the `scripts/log-battery.mjs` script while [dis]charging the battery in `scripts/battery-stats`.

Some wise words from Martin, our electronics consultant:

> Log it out your USB port, and let it run from fully charged to discharged.
> In addition to measuring voltage, it is a good idea to periodically check the current also.
> If current is steady over the full time duration, the # of mA times # of hours should be close to the mAhr capacity of your battery.

[...]

PR #1142, opened 2025-02-20
Implement basic battery manager.

### LINKED ISSUE #1011: setLedTxt errors if variables are used
As a follow-up from UltimateHackingKeyboard/firmware#933 ...

Error at countdown 1/4/17: Unrecognized command: count
> 4 | [setLedTxt 1000 $count](https://github.com/UltimateHackingKeyboard/firmware-uhk80/issues/new#/macro/7?actionIndex=0&lineNr=4&columnNr=17&inlineEdit=true)
>   |

(totally not urgent, but I don't want this to be forgotten)

### LINKED ISSUE #1206: Zephyr logs API specification
UHK will expose two new usb variables: 
```
        UsbVariable_ShellEnabled,
        UsbVariable_ShellBuffer,
```

- UsbVariable_ShellEnabled is false by default. When true, firmware gathers zephyr shell logs in a new buffer.
- UsbVariable_ShellEnabled is set by Agent when Agent/user decide they want to see logs.
- when there is a new log, the 8 bit is set in the byte 2 of get device state command.
- when UsbVariable_ShellBuffer is read, firmware consumes a part of the buffer and returns it to Agent as a null delimited string. (Null delimiter is present only if sending less than maximum payload length.)  
- when Agent is disconnected, UsbVariable_ShellEnabled is automatically set to false again. 

-------------------------------
@ert78gb @mondalaci does this sound sane?

@ert78gb would you like some additional metadata in the UsbVariable_ShellBuffer response?

### LINKED ISSUE #1210: Insead of "charge me", blink a status icon
On the main OLED screen, blink a battery empty icon in the top status bar, such as https://fontawesome.com/icons/battery-slash?f=classic&s=solid

### LINKED ISSUE #1214: Weird button release behavior after updating to 13.0.2
I've recently updated from an older firmware version, 11.2.0, to 13.0.2, and am now experiencing various oddities around keys triggering actions they shouldn't on my UHKv1, usually in combination with the mouse layer.

I can reproduce as follows:

1. **Hold down** the key to trigger the mouse layer (left special key under the spacebar).
2. Keep moving the mouse to the left or right by **holding** the default keys on the mouse layer (e.g. what is visually on the keyboard the 'L' key to move right).
3. Whilst holding down the mouse layer and mouse movement keys (so the mouse is now moving), **press (but don't release)** the Meta/Windows key.
4. Observe how the Meta key press immediately registers as a key up/release as well, despite the key not being released yet. This is not the case when the mouse layer is not active. If you release the key, it triggers the Meta action bound to it a second time.

In GNOME this is especially noticeable since this results in the overview immediately being opened or closed, despite me not having released the super key. In GNOME a way to move windows is also to hold down the super key and then use the mouse to click on the window and move it afterwards, which this sudden change has broken or at least made unreliable.

**If you switch around 2 and 3 and press the Meta key first, this behaviour doesn't happen and everything works as intended.**

Another oddity is that 3 often results in the mouse movement speeding up drastically, as if the Meta key is some sort of accelerator all of a sudden. This may be a different issue though (perhaps #1186).

kareltucek:
Solves the most pressing issues of #974.

Closes #1206.
Closes #1210.
Closes #1011.
Closes #1214.

What it does:
- implement controls that allow easily setting battery limits
- implement a state machine that enables and disables the charger as needed
- implement a charging mode for stationary use that targets 3.8V, and its macro controls `set battery.chargeLimit { full | optimizeHealth }`
- eliminate the step of battery indicators between powered and unpowered states and various brightness states.
- blink battery-low icon on the oled when battery is low
- implement setLedTxt for uhk80
- fix key chatter on uhk60 when letters are typed while mouse-keys are being used

Furthermore:
- switch zephyr to deferred logging.
- implement a new zephyr logging backend and forward logs to Agent via a new buffer.
- move status buffer to wormhole region and log crashlogs there.

What it does not do:
- adaptively detect 100% voltage
- we do not calibrate according to actual battery health (yet)


kareltucek:
This isn't tested very well yet.

kareltucek:
Known issues:
- ~~battery presence detection is broken.~~ fixed
- there is still the step caused by whether battery is being drained or not during measurement. (I don't plan to fix this at the moment.)
- ~~i yet plan to implement shutdown mode when battery drops bellow "0". (It looks really bad when battery indicator says 0, but the keyboard still runs happily.~~ implemented, although in a very basic manner.

mondalaci:
The code looks very good and well abstracted.

If the charger is disconnected, what policy will you implement for `ChargeRegion_Empty` vs. `ChargeRegion_AlmostEmpty`? It might be worth displaying `Charge me` on the OLED for either or both and not enabling per-key backlight in the first place.

kareltucek:
For empty, uhk shuts down (as deep as I can make it).

Almost empty is just an auxiliarry region that makes sure that the keyboard doesnt oscillate between power on and off.

I am not displaying anything at the moment.

mondalaci:
Sounds reasonable.

`Charge me` may make sense for `ChargeRegion_AlmostEmpty`, but we'll see what works in practice.

kareltucek:
I am not strictly against introducing the warning, but I would prefer to split into a separate issue (I am already writing another code on top of this, am not sure where to render it, and am not even sure we need it)...

kareltucek:
I think this is ready for merge.

kareltucek:
Actually, not yet. Half will not wake up if restarted 🤦.

mhantsch:
Instead of a "charge me" message (and not knowing where to place it), I would flash the battery icon slowly (e.g. 1/sec).

mondalaci:
I'd center-align "charge me". A center-aligned battery icon should also work, probably even better.

When running `powerMode sleep`, both halves go to sleep, but the right half displays the colors of the Fn2 layer actions with minimum brightness, and the left half has S and D on with minimum brightness.

S+F or J+L need very little time to activate. I think it'd be useful to request that you press these keys (and no other key) for at least half a second to activate.

I suggest using the `battery.lifeExtenderMode` name instead of `battery.stationaryMode`.

I need explicit instructions on how to test whether backlight oscillation is noticeable during battery voltage measurement. I use a USB and a bridge cable. How long should I observe it? I also believe that the charge level is an important variable, and the backlight oscillation may be more easily noticeable at low battery levels.

mondalaci:
Also, red-blinking S+F and J+L would consume less power than white-blinking them, so I'd prefer red.

mondalaci:
I may be wrong about the requested half-second activation time. It may not be necessary in practice. Feel free to ignore it; real-world usage will reveal whether it's justified.

mondalaci:
`shutDown` should be named `shutdown`, as it's a single word. But more importantly, what's its purpose, as I can't wake up my UHK after shutting it down, except by resetting it? The dim LED issue applies, as with `sleep`.

kareltucek:
> When running powerMode sleep, both halves go to sleep, but the right half displays the colors of the Fn2 layer actions with minimum brightness, and the left half has S and D on with minimum brightness.

Noted.

> I suggest using the battery.lifeExtenderMode name instead of battery.stationaryMode.

Stationary mode is I think a common term for this, used for instance with laptop batteries. @mhantsch @pcooke9 (anyone else) opinions?

2025-04-03-112149_1293x654_scrot

> shutDown should be named shutdown

Sure!

> S+F or J+L need very little time to activate. I think it'd be useful to request that you press these keys (and no other key) for at least half a second to activate.
> ...
> I may be wrong about the requested half-second activation time. It may not be necessary in practice. Feel free to ignore it; real-world usage will reveal whether it's justified

I don't think there is a risk of accidental wake with current sf/jl, but it means I can increase the polling time 🤔. 

> But more importantly, what's its purpose, as I can't wake up my UHK after shutting it down, except by resetting it? The dim LED issue applies, as with sleep.

It is a mode that UHK enters when the battery runs out. Uhk wakes from it when you connect it to usb (power).  Users are not meant to use it directly.

> I need explicit instructions on how to test whether backlight oscillation is noticeable during battery voltage measurement. I use a USB and a bridge cable. How long should I observe it? I also believe that the charge level is an important variable, and the backlight oscillation may be more easily noticeable at low battery levels.

Well, in `charger.c`, you can drop
```
#define CHARGER_UPDATE_PERIOD 60000
```
to something like 2000.

But in your original report you didn't need to do that and yet your UHK oscillated with a ~1 second-ish period. How you did it I don't know.

I expect the backlight oscillation to be noticeable when the battery is actively charging (i.e., powered, but not full).

mondalaci:
I don't insist on the suggested naming, but "stationary mode" doesn't ring a bell to me, so I devised the alternative using Claude. We should use what is most intuitive for most users.

Would you do two consecutive checks for the polling time regarding SF/JL? It's probably the best approach.

The shutdown mode should be a development feature and may be exposed separately as such. However, the sleep and lock modes will be directly exposed as device actions via Agent for easy use, so it may not matter. Please expose the relevant device actions for Robi.

My batteries are at 100%, and I can't notice the oscillation. Let me deplete them and I'll retest.

kareltucek:
> The shutdown mode should be a development feature and may be exposed separately as such. However, the sleep and lock modes will be directly exposed as device actions via Agent for easy use, so it may not matter. Please expose the relevant device actions for Robi.

The sleep mode is exposed. 

I am not sure the lock mode should be exposed actually, as it doesn't reduce power consumption.

kareltucek:
> My batteries are at 100%, and I can't notice the oscillation. Let me deplete them and I'll retest.

(I think it should be proportional to the current that charges the battery actually, so the lower the battery, the more pronounced it should be.)

mondalaci:
I'd also expose the lock mode as a device action. Imagine someone carrying his UHK around without getting disconnected via wireless.

kareltucek:
All suggestions should be addressed except for "charge me" text and the battery mode name. I am not at all convinced by `battery.lifeExtenderMode`. Colleagues suggest "apple mode" and "optimized charging mode". Will leave it as is until more feedback is gathered.

mhantsch:
> Also, red-blinking S+F and J+L would consume less power than white-blinking them, so I'd prefer red.

5% of men have red-green colour deficiency. I am one of them. I can hardly see red, or it looks much darker to me than other colours. I don't think red is a good choice.

mondalaci:
Thanks, I wouldn't have ever thought about that! I recommend blue then.

pcooke9:
As for the battery mode; I kinda like "Stationary mode" because it makes you think of the use cases. But I think something like "battery protect" is probably more intuitive.

~~The description text for each mode is great the way they are in the image though.~~ 🤦🏻Sorry, hadn't had coffee yet.😂 

EDIT: Maybe better wording would be `battery.preserveHealthMode` or `battery.protectHealthMode`.
(I think Laszlo's "lifeExtender" wording is probably closer in spirit to what the function actually does, but the word "extend" generally refers to adding more to something, or to make something larger/longer, which isn't _technically_ possible in terms of battery life.)

mondalaci:
Thanks for your suggestions! I like `battery.preserveHealthMode`.

mhantsch:
I'm going to throw my €0.02 in:

```
set battery.chargeLimit 75    // maximum percentage to charge to
set battery.chargeLimit full   // charge to 100%
set battery.chargeLimit optimizeHealth // charge to 80%
set battery.chargeLimit stationaryUse   // charge to 50%
```

kareltucek:
@mhantsch thanks, but the feature is set to maintain the lipol storage optimum which is (according to my sources) 3800 mV, which corresponds to roughly 83% according to current percentage computation. It is intentionally anchored to voltage and not to percentage, and intentionally not exposed to the user, as they would tend to leave the battery undercharged, incorrectly believing they are doing a good thing for their battery.

pcooke9:
MAX Concise strikes again!😆
That does look more logical, for sure.

~~Is 50% actually considered optimal for stationary use? I don't think I've ever seen any info on that use case.~~ Karel's comment answered my question before I could ask.🏃‍♂️

kareltucek:
Although, I must admit I like the proposed syntax because it is more descriptive as well as more future proof.

I suggest using this syntax. Internally, it would still map to a simple boolean.

```
set battery.chargeLimit full
set battery.chargeLimit optimizeHealth
```

~~@mondalaci ?~~ Edit: implemented this variant. Feel free to protest.

mhantsch:
> 3800 mV, which corresponds to roughly 83%

I always thought for long-term storage, it's best to keep LiPos around half filled, i.e. somewhere between 40%-60%. And up to ~80% you still get very little degradation, but benefit from more energy being available in the battery in case you have to go off the charger.

In fact, when I get fresh replacement batteries e.g. for my phone, or my laptop, they arrive about half charged when I take them out of the box. Clearly they come at "best storage level", and instructions are to fully charge them once at the beginning (so the BMS can learn their full capacity).

I understand LiPos age most when kept at less than 20% or higher than 80% for prolonged periods of time.

I can't translate that to voltage; I don't have data for that.

mhantsch:
Lipo-battery-guide-Voltage-vs-capacity-used-percentage

https://oscarliang.com/wp-content/uploads/2017/02/Lipo-battery-guide-Voltage-vs-capacity-used-percentage.jpg

kareltucek:
There are other issues, like our charging voltage and the percentage calculation. The charger cannot provide more than ~4.1V, which means that for us, 4000mV is 100%.

Rest assured that we charge to 3.8V, whatever the actual percentage is.

kareltucek:
As far as user experience goes, I feel like 500ms scanning rate and blue blinking is bad. I suggest dropping to 75ms and white blinking.

kareltucek:
Another chapter of battery management should be finished. I would not start another until this one is tested and merged in master. 

Current state of affairs:
- Charger should be returning consistent battery voltages / percentages. E.g., when plugging the charger in and out, or when turning backlight on or off when running on battery.
- Battery percentage is now computed based on actual charging curves.
- I have raised the cutoff limit to 3.4V after seeing the battery curves, and think we should consider raising it to 3.5V.

We do not adapt to actual battery health at this point - instead we use a pre-measured statistics. Switching to some adaptive or manual calibration will be another step at some point in the future.

At this point I suggest to test and merge this and see if any issues pop up.

kareltucek:
Oh, and you can set `#define DEBUG_BATTERY_TESTING true` to test stuff in `debug.h`.

It adds some logs, disables voltage averaging, and shortens battery measurement intervals, so it is possible to observe any quirks much more easily.

PR #1157, opened 2025-03-03
Disable gamepad by default.

### LINKED ISSUE #1153: Add gamepad enable flag to user configuration
On UHK80 the gamepad interface can be disabled by a shell command already, this should be extended / changed to be done via agent and stored in the user configuration. The API to call when the value changes is `HID_SetGamepadActive`.

kareltucek:
Closes #1153 .

### DIFF
--- a/device/src/main.c
+++ b/device/src/main.c
@@ -171,6 +171,7 @@ int main(void) {
         }
     }
 
+    HID_SetGamepadActive(false);
     USB_Enable(); // has to be after USB_SetSerialNumber
 
     // has to be after InitSettings


mondalaci:
Given [this feedback](https://github.com/UltimateHackingKeyboard/firmware/issues/1143#issuecomment-2695077060), do you plan to work more on this?

If no, should we merge it, in which case what improvement does it provide?

If yes, what additional developments do you plan?

kareltucek:
I don't think there is anything more to do, so no.

I think yes, we should merge it:
- consistency. Gamepad interface is disabled on uhk60, so it ahould be on uhk80 too.
- gamepad interface caused some issues to people with uhk60, so it is reasonable to expect that it may cause similar issues for uhk80
- we don't support gamepad atm, so there is no point in exposing its hid interface

mondalaci:
Ok, meged. What other steps to take to increase KVM compatibility?

kareltucek:
No idea. That is a question for Benedek.

PR #1162, opened 2025-03-06
Parse device actions.

### LINKED ISSUE #1099: Device actions
This is the firmware-side issue of https://github.com/UltimateHackingKeyboard/agent/issues/2502

A new `key_action_type_t` value has to be added for special actions. For clarity, it'd make sense to add one id for connections and another for the other special actions, which are expected to be extended later, but if we're short on key ids (assuming some bits of this byte are used in the RAM for other purposes), feel free to use a single, shared `key_action_type_t` value.

It'd be nice to visualize the empty/nonempty connection slots by showing the empty ones with half the brightness of their color when functional backlighting is used. (I'll extend the default configuration to map the "switch to connection 1", "switch to connection 2", ... actions to neighboring keys of the Fn2 layers.)

The new functional key color for special key ids must also be added.

kareltucek:
Closes #1099. Implements support for agent#2502 .

@ert78gb next connection should be coded as `36 0`, previous as `36 1`. Switch to connection 3 would be coded as `36 2 3`, sleep as `37 0`.

(Not tested.)


ert78gb:
I will continue the Agent implementation today.

ert78gb:
The https://github.com/UltimateHackingKeyboard/agent/issues/2502 issue contains 

> Connections: Display the connections in a list similar to the one on the Connections page, but only display a checkbox and the connection name per row. Start with "Next connection", "Previous connection", "Last connection", then list all the 20 connections starting with their array index

The `Last connection` directly has been removed?

kareltucek:
I see. I didn't notice that. I will implement that. Please use the obvious index for the action.

ert78gb:
Obvious is subjective :) The `Last connection` will 0 because the enum is in alphabetical order.

kareltucek:
Well, yes. Last 0, next 1, previous 2, switchbyid 3 😇.

right/src/ledmap.c
```diff
@@ -505,6 +505,10 @@ static rgb_t* determineFunctionalRgb(key_action_t* keyAction) {
         case KeyActionType_PlayMacro:
             keyActionColor = KeyActionColor_Macro;
             break;
+        case KeyActionType_Connections:
+        case KeyActionType_Other:
+            keyActionColor = KeyActionColor_Special;
```
ert78gb: @kareltucek  Would we like to keep it as Special or we rename it to Device too? It is up to you I just follow the same namings in Agent

kareltucek: Good catch. Feel free to change it to Device.

mondalaci: Name it device.

kareltucek:
Closing in favor of #1174 . (I have accidentaly rebased the commits somewhere along the way.)

PR #1163, opened 2025-03-06
Add reboot macro command.

### LINKED ISSUE #1154: Implement a `reboot` macro command

kareltucek:
Closes #1154

### DIFF
--- a/device/src/messenger.c
+++ b/device/src/messenger.c
@@ -24,6 +24,7 @@
 #include "resend.h"
 #include "debug.h"
 #include "trace.h"
+#include "usb_commands/usb_command_reenumerate.h"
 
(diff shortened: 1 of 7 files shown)


mondalaci:
When I add a reboot macro, map it to a key on the Fn2 layer, and save my configuration, the keyboard seemingly reboots even though I didn't press the relevant key. The configuration doesn't get saved, and on one occasion, my configuration was corrupted in the process.

Why restart both halves for the UHK 80 but only the right half for the UHK 60?

The documentation should be updated.

kareltucek:
Ah, of course. 🤦

> Why restart both halves for the UHK 80 but only the right half for the UHK 60?

Because uhk60 left half is just a module.

PR #1166, opened 2025-03-06
Always look up and log host names.

kareltucek:


mondalaci:
It's still not good.

image

Upon pairing:

```
Supplied connection (123) doesn't correspond to a host connection!
connected n/a (n/a, a8:41:f4:11:a4:08), 3 3
Received passkey pairing inquiry.
Type `uhk passkey xxxxxx` to pair, or `uhk passkey -1` to reject
```

kareltucek:
Oops. 

Should be better now, but not tested.

PR #1168, opened 2025-03-07
Try to prevent trackpoint jumps.

kareltucek:
@mondalaci please give a try to this branch. I have realized that unlike uhk60, uhk80 is probably allowed to preempt either way.


mondalaci:
I figured that trackpoint jumps on the UHK 80 don't actually occur often, maybe not at all.

It's just that there's a bug that sometimes kicks in. It makes mouse movement faster and also makes the mouse pointer very jumpy. But it's not trackpoint-related and affects mouse keys, too. I'll keep testing master to see whether trackpoint jumps are an issue at all for the UHK 80.

kareltucek:

PR #1182, opened 2025-03-15
Message pool fixes backported

kareltucek:
(This is mainly to generate and test a build. We can't work with master anymore as it requries new agent.)


mondalaci:
This PR makes mouse movement choppy and erratic both via modules and mouse keys.

PR #1184, opened 2025-03-17
Fix nus refusal crashes.

kareltucek:
Steps to reproduce:
- disconnect dongle
- unpair it
- connect dongle
- try to pair

Observe:
- right half crashes.
- apply first commit of this PR and retry
- now it no longer crashes, but doesn't pair either

### DIFF
--- a/device/src/bt_advertise.c
+++ b/device/src/bt_advertise.c
@@ -3,6 +3,7 @@
 #include <zephyr/bluetooth/gatt.h>
 #include "attributes.h"
 #include "bt_conn.h"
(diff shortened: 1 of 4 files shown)


kareltucek:
No point in merging this now.

mondalaci:
Please explain why.

kareltucek:
I will open a PR that will contain the two closed PR and the discussed pairing modes.

All three sets of changes are closely related. Testing them separately would be duplication of effort, and even if any problems were found they may easily not be relevant anymore because of the changes on the other branches.

PR #1216, opened 2025-04-22
UHK-60: fix mouse high-res scrolling for Linux

benedekkupper:
There is a Linux bug that [I reported](https://bugzilla.kernel.org/show_bug.cgi?id=220144), the kernel incorrectly calculates the size of the set report payload when there is no report ID, ending up with a 0 length payload. This is what @rightaditya worked around with by accepting set report with no data. The reason no such workaround is necessary on UHK80 is because there we use report IDs. I changed the UHK60 implementation to have a report ID, so Linux sends correct payloads, and we can eliminate the guessing logic.


mondalaci:
On Linux, scrolling in vscode was already smooth, but not in Chrome. This PR also makes scrolling smooth in Chrome.

On Windows 11, however, scrolling was smooth in Chrome, but now it's choppy. This must be fixed.

on macOS, scrolling is unchanged.

benedekkupper:
Please check the new build

mondalaci:
Now, the choppiness persists on Windows, and the smoothness is gone on Linux-Chrome. So, the current state is a step backward compared to your earlier commit, and it reflects master. macOS is unchanged.

mondalaci:
@benedekkupper Should I test this again?

benedekkupper:
No, I just rebased so I could flash it with agent, but I observe the same behavior. I failed the disassembly to solder the debug connection, so I’m kind of blocked on UHK-60 for the time being.

benedekkupper:
With the debugging infrastructure set up, I was able to trace the remaining issue in this PR, and now it should work just as it does on UHK80. @mondalaci @rightaditya please try this out.

kareltucek:
That bug - https://github.com/UltimateHackingKeyboard/firmware/issues/1261 - is present on uhk80 too. (Unless this pr fixes it.)

mondalaci:
@benedekkupper What's the rationale for fixing MCUXpresso SDK USB descriptors if the UHK 60 firmware will be ported to c2usb anyway?

benedekkupper:
It's an interim fix, since it was in progress anyway I thought I will clear this up, as getting to c2usb on kinetis might take a longer time.

For the suspend issue, I cannot reproduce myself, because both my Linux systems (old Thinkpad laptop, RPi 400) don't have working suspend mode.

kareltucek:
I can reproduce it with my home laptop. Works fine with the other one though. Both linuxes.

kareltucek:
I.e., I can provide logs. I will be away from friday for a week though.

rightaditya:
> With the debugging infrastructure set up, I was able to trace the remaining issue in this PR, and now it should work just as it does on UHK80. @mondalaci @rightaditya please try this out.

@benedekkupper I'm out of town at the moment but can give it a go over the weekend.

rightaditya:
> There is a Linux bug that [I reported](https://bugzilla.kernel.org/show_bug.cgi?id=220144)

@benedekkupper Thanks for reporting it! I considered it but didn't really want to interact with LKML, especially since I wasn't 100% confident about what was going on. Glad to see the issue officially documented... hopefully something will come of it.

mondalaci:
> I can reproduce it with my home laptop. Works fine with the other one though. Both linuxes.

Please double-check this, as high-res scrolling now works consistently across OSes for me.

benedekkupper:
If this branch in its current state still produces the scrolling problem after a suspend-wakeup cycle, a brute-force way to solve #1261 is to do a USB soft disconnect-reconnect after wake. We can already identify when the keyboard is plugged into a Linux host (mouse scroll multiplier set, no MSOS descriptor fetched), and limit this functionality to it.

kareltucek:
I gave this branch a quick test with UHK80 first, not realizing that this PR tires to fix only the uhk60 variant of the bug. It makes cursor shoot at roughly 200x its normal speed.

As for UHK60, mouse speed after a clean boot is fine. After suspending the host, the leds go out at first, then briefly flash twice, then the board reboots.

PR #1218, opened 2025-04-25
Extend user config with battery settings.

kareltucek:
Implements firmware side of https://github.com/UltimateHackingKeyboard/agent/issues/2619

I have set the version to 9.99.0, so we can merge it into master and yet add more things before 10.0.0 release.

### DIFF
--- a/right/src/config_parser/parse_config.c
+++ b/right/src/config_parser/parse_config.c
@@ -32,6 +32,10 @@
 #include "keyboard/oled/widgets/widgets.h"
 #endif
 
+#ifdef __ZEPHYR__
+#include "state_sync.h"
+#endif
+
 version_t DataModelVersion = {0, 0, 0};
 
     bool PerKeyRgbPresent = false;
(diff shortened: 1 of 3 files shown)


mondalaci:
Unlike the "charging brightness" setting, the "extend battery life" setting doesn't update after saving the configuration.

pcooke9:
> Unlike the "charging brightness" setting, the "extend battery life" setting doesn't update after saving the configuration.

The "Extend battery life: limit maximum battery charge level" wasn't persisting for me either, at first. It was because I had `set battery.chargeLimit optimizeHealth` in my `onInit`.

I deleted the macro from `onInit`, but after a couple of failed attempts to set the "Extend battery life" toggle, I rebooted the UHK80, and I saw that the `set battery.chargeLimit optimizeHealth` entry had returned to my `onInit`. After a couple of failed attempts to delete it, and rebooting the UHK80 and restarting Agent, the removed `onInit` entry finally saved properly.

The "Extend battery life: limit maximum battery charge level" toggle works properly after making sure `set battery.chargeLimit optimizeHealth` is actually removed from `onIint`.

mondalaci:
I can't believe I've made the same mistake! :D

PR #1220, opened 2025-04-29
Log traces on freezes.

kareltucek:


kareltucek:
Deprecated by https://github.com/UltimateHackingKeyboard/firmware/pull/1229

PR #1223, opened 2025-05-06
Add off backlight mode.

kareltucek:
(Added for personal convenience.)


mondalaci:
This macro only disabled the LEDs of the right half of my UHK 80.

kareltucek:
That's an unrelated bug that I haven't been able to reliably reproduce so far, but has there been for quite a while :-/

PR #1238, opened 2025-05-20
WIP: Implement argument parser.

kareltucek:
Relates to:
- https://forum.ultimatehackingkeyboard.com/t/is-there-a-way-to-parametrize-macros/2386/2

### DIFF
--- a/right/src/config_parser/parse_keymap.c
+++ b/right/src/config_parser/parse_keymap.c
@@ -226,7 +226,21 @@ static parser_error_t parseZeroBlock(key_action_t *keyAction, config_buffer_t *b
     return ParserError_Success;
 }
 
-static parser_error_t parseKeyAction(key_action_t *keyAction, config_buffer_t *buffer, parse_mode_t parseMode, uint8_t *actionsToZero)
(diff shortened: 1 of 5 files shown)


mondalaci:
How to test this?

kareltucek:
For the moment just flash it and see that uhk doesn't explode.

Otherwise it is just an empty parser so that Robi can comfortably implement the UI part.

kareltucek:
Anyway, given that Agent still doesn't have version handling sorted out, we should not merge this for the moment.

mondalaci:
I expect argument lengths to be almost always smaller than 255 bytes, so a `ReadCompactLength()`-like serialization would be more space efficient.

kareltucek:
Good point!

mondalaci:
Also, instead of `SerializedArgumentType_Label` and `SerializedArgumentType_MacroArg`, I'd rather have `SerializedKeyActionType_Label` and `SerializedKeyActionType_MacroArg` as it'd also be more space-efficient and there are plenty of `SerializedKeyActionType_*` values to use. Correct me if I'm wrong.

kareltucek:

PR #1243, opened 2025-05-23
Implement zephyr shell usb command.

### LINKED ISSUE #995: Expose the Zephyr shell via the device protocol
Then, Agent would make the shell accessible via xterm.js

https://devzone.nordicsemi.com/f/nordic-q-a/86501/zephyr-shell-with-custom-interface-transport-layer/361292

The link protocol would transfer the shell of the left half.

We could use a new USB protocol command for reading and writing the shell. When there is shell content to read, GetDeviceState would mark the relevant bits as dirty.

kareltucek:
Closes https://github.com/UltimateHackingKeyboard/firmware/issues/995

E.g., this lists `uhk connections` command:

```
agent/packages/usb/send-command.ts --vid=0x37a8 --pid=9 0x1e 117 104 107 32 99 111 110 110
101 99 116 105 111 110 115 0
```

### DIFF
--- a/device/src/proxy_log_backend.c
+++ b/device/src/proxy_log_backend.c
@@ -50,15 +50,19 @@ void ProxyLog_SetAttached(bool attached) {
     ProxyLog_IsAttached = attached;
 }
 
+static void addChar(char c) {
+    if (bufferLength < PROXY_BACKEND_BUFFER_SIZE) {
+        buffer[POS(bufferLength)] = c;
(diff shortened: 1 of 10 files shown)


mondalaci:
`uhk connections` successfully runs this way, and its output shows up via hardware UART, but I'd also expect it to show up via the Zephyr shell of Agent to make this feature useful for users.

kareltucek:

PR #1244, opened 2025-05-23
Don't parse incompatible config versions.

### LINKED ISSUE #791: Don't parse newer, unsupported user configs
The config parser is backward compatible, which is great. But I've just tested a firmware issue, and my UHK has misbehaved, presumably because my UHK has a newer user config than what is unsupported by the flashed firmware.

When the firmware encounters a user config with a major or major.minor user config version number larger than the one supported, blink "DOW" on the display hinting that the user should *dow*ngrade the user config.

kareltucek:
Closes https://github.com/UltimateHackingKeyboard/firmware/issues/791

Not tested. Please check by reading it carefully.

### DIFF
--- a/right/src/config_parser/basic_types.h
+++ b/right/src/config_parser/basic_types.h
@@ -9,6 +9,7 @@
 // Typedefs:
 
     typedef struct {
+        bool isValid;
         uint8_t *buffer;
         uint16_t offset;
     } config_buffer_t;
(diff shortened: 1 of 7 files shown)


mondalaci:
I'm not sure how to test this. If you backported this PR to 13.0.2 in a separate branch, I could easily test it.

kareltucek:
Just drop the user config version in firmware s package json, generate versions, compile and flash.

mondalaci:
Dropped the user config version to 10.0.0, generated the versions, flashed the firmware, but DOW doesn't appear on the display.

kareltucek:
Talking about uhk60 or uhk80?

mondalaci:
Both.

PR #1250, opened 2025-05-29
Fix flash write race.

### LINKED ISSUE #1249: UHK 80: some config changes are not persistent across reboots
I'm on:
Firmware: 14.0.1
Agent: 7.0.1

When I set:
LED settings / LED display timeout or Key backlight timeout to 10s in the agent for USB-powered it resets to 5minutes (both) after the I restart the keyboard by plugging it into another machine.

kareltucek:
Closes https://github.com/UltimateHackingKeyboard/firmware/issues/1249 . 

Original writing sequence:
- load new config into staging area via usb
- apply the config:
  - parse the config in dry run mode
  - if this succeeds, swap the buffers
  - parse it again, but this time actually apply it
  - execute effects 
- launch storage transfers

Issues with the above:
- c2usb has small stack. This used to cause stack overflows on uhk80.
- we don't want to trigger all the application effects from an interrupt.

As a result I had decided to make the application asynchronous and execute it from the main thread. 

So the new uhk 80 sequence up until now was: 
- load new config into staging area via usb
- apply the config usb command:
  - validate config in dry run mode
  - if it succeeds, return success
  - schedule application for the main thread
- from the main thread: 
  - validate the config again
  - if this succeeds, swap the buffers
  - parse it again, but this time actually apply it
  - execute effects 
- launch storage transfers

This now leads to a race on these two events: 
- write validated buffer to flash
- swap the buffers

### DIFF
--- a/device/src/main.c
+++ b/device/src/main.c
@@ -166,7 +166,7 @@ void mainRuntime(void) {
             Flash_ReadAreaSync(userConfigArea, 0, StagingUserConfigBuffer.buffer, USER_CONFIG_SIZE);
             printk("Applying user config\n");
             bool factoryMode = false;
-            if (factoryMode || UsbCommand_ApplyConfig(NULL, NULL) != UsbStatusCode_Success) {
+            if (factoryMode || UsbCommand_ValidateAndApplyConfigSync(NULL, NULL) != UsbStatusCode_Success) {
                 UsbCommand_ApplyFactory(NULL, NULL);
             }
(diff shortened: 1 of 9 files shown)


kareltucek:
Steps to reproduce: 
- import vpal's user config
- change led timeout
- save the config
- press reset button
- notice that usb isn't working now
- reconnect the usb or powercycle uhk if needed
- when agent loads, see that the old timeout is still there

PR #1253, opened 2025-05-30
Staging

kareltucek:
...


kareltucek:
[uhk-firmware-14.0.3-debug-v1.8.tar.gz](https://github.com/user-attachments/files/20525346/uhk-firmware-14.0.3-debug-v1.8.tar.gz)

Issue #999, opened 2024-09-14
Cannot pair UHK80 via Bluetooth to my Linux laptop via settings UI


mhantsch:
Framework Laptop 13 running Pop_OS! 22.04 (based on Ubuntu 22.04). 

I am trying to connect & pair the UHK80. It shows up in the Bluetooth settings as "Not set up" but when I click on it, nothing happens. I don't get prompted with pairing code, nor does the UHK show anything.

Pairing works fine from my Android phone.

If I try manually from commandline, I see this:

```
(base) max@max-framework:~/src/lampster$ bluetoothctl pair D2:89:0A:CA:61:05
Attempting to pair with D2:89:0A:CA:61:05
[CHG] Device D2:89:0A:CA:61:05 Connected: yes
[CHG] Device D2:89:0A:CA:61:05 Connected: no
Failed to pair: org.bluez.Error.AuthenticationCanceled
```

However, if I manually just issue a "connect", it then starts pairing:

```
(base) max@max-framework:~/src/lampster$ bluetoothctl connect D2:89:0A:CA:61:05
Attempting to connect to D2:89:0A:CA:61:05
[CHG] Device D2:89:0A:CA:61:05 RSSI: -53
[CHG] Device D2:89:0A:CA:61:05 Connected: yes
Connection successful
```

So I finally got it to pair and it works fine now, but I couldn't have done that through the UI.
Seems like there is still room for improvement.

Let me know how I can provide you with further traces, or anything else you need.

mondalaci:
For me, pairing works from blueman-manager 2.4, but it's not intuitive.

When I first tried it, I entered the pairing code on my UHK. But by then, the pairing notification disappeared on Linux Mint, making me forget to pair it on my PC, so pairing didn't succeed, seemingly for no reason.

Later, I retried pairing and realized that I should open my notifications. When confirming the pairing on my PC, it immediately worked.

Which GUI do you use exactly, and have you confirmed pairing on your laptop?

mhantsch:
I am using the Pop_OS! system configuration UI. It looks like this:

![image of Bluetooth settings](https://github.com/user-attachments/assets/25ed2584-5bf6-435f-acd4-faec3fffe179)

[...]

PR #1262, opened 2025-06-07
Fix linux pairing.

kareltucek:
Edit, explanation: https://github.com/UltimateHackingKeyboard/firmware/issues/999#issuecomment-2952811985


kareltucek:
Alright, saying that it is fixed is is a huge exaggeration, I have managed to pair at least 5 times with this branch this evening 😑.

benedekkupper:
@kareltucek perhaps this together with #1335 will be a good candidate for user tests, what do you think?

kareltucek:
@benedekkupper I think so. 

(To be precise, I guess we mean that someone should try it first, and if things go well then we should give it to users.)

mhantsch:
If I should test this, let me know soon (and which build). I'll be away (AFUHK) from Sunday on a business trip. So either in the next few days, or it will have to wait until the second half of September.

benedekkupper:
The commit was picked up in #1335 and got merged, closing this PR.

PR #1265, opened 2025-06-08
move reboot after powerMode, instead of being in the middle

rdpate:
When reboot was added to the documentation, it was just before powerMode.  Later, it was resorted into its current location that splits powerMode documentation, which was itself missing a hyphen.

Commit moving reboot: 323b1a690b2f2d1c343db3814bd62bc3b4329ccf


mondalaci:
Thanks for your contribution!

PR #1268, opened 2025-06-09
Refactor bt logging.

kareltucek:


mondalaci:
Should I test this? If so, how?

kareltucek:
If you see some logs in Agent / status buffer that seem like they are not wanted there, then let me know. Apart from that, I think there is not much to test.

PR #1269, opened 2025-06-10
Fix update of last report when macro ends.

kareltucek:
Changelog: 
- Fix a macro engine bug where usb reports weren't updated at the end of a macro. This would exhibit as keys activated from the macro would get stuck.

Steps to reproduce: 
- run a macro with a single action: `pressKey a`
- Expected: brief a tap
- Actual: a gets stuck until another scancode key is pressed.

Reported in https://forum.ultimatehackingkeyboard.com/t/gaming-config-tips-tricks/2447/16

### DIFF
--- a/right/src/macros/core.c
+++ b/right/src/macros/core.c
@@ -265,6 +265,12 @@ static void freeLocalScopes()
 
 static macro_result_t endMacro(void)
 {
+    /* If macro has finished and some keys are pressed, we need to merge the reports once more in order to release them */
+    if (S->ms.reportsUsed) {
+        S->ms.reportsUsed = false;
+        EventVector_Set(EventVector_SendUsbReports);
+    }
+
     freeLocalScopes();
 
     S->ms.macroSleeping = false;


mondalaci:
Should I merge this even though the forum use case persists?

kareltucek:
Please share your UserConfig, the forum usecase works fine for me.

mondalaci:
[UserConfiguration.json](https://github.com/user-attachments/files/20738322/UserConfiguration.json)

Hit Fn2+d to switch to the Doom Eternal keymap, then hit the Caps Lock key to run the macro.

kareltucek:
Sill seems to work just fine 🤨.

kareltucek:
What exactly is it that doesn't work? 

Can you reproduce it with https://keyboardchecker.com/ ?

mondalaci:
When tapping Caps Lock, `5` gets stuck forever on my UHK 80. I can reproduce it with keyboardchecker.com

pcooke9:
~I can reproduce it on v14.2.0. The `5` key gets stuck. If you tap a different key while it's stuck, it will seem to stop, but then the real `5` key doesn't work anymore. Keyboardchecker.com shows the `5` is indeed being held, even when it seems to have stopped outputting to a text editor. Double-tap and hold the macro a couple of times resolves it.~

Derp.🤦‍♂️

kareltucek:
The fix is not part of 14.2.0!

If you want to decide it for us, pkease the relevant ci aetifact ;-).

pcooke9:
OMG, sorry. Long day!😅

pcooke9:
Ok, using the right one this time...🤦‍♂️

Quick tapping the macro to produce `5` will keep it repeating until tapping it again.

If the `5` is stuck and you use the macro to trigger the `1` key, the `1` will be output as it should, but the `5` keeps repeating until quick tapped again. Not sure if that is expected behavior or not.

kareltucek:
> If the 5 is stuck and you use the macro to trigger the 1 key, the 1 will be output as it should, but the 5 keeps repeating until quick tapped again. Not sure if that is expected behavior or not.

That's what the macro says 😊.

mondalaci:
`pressKey a` doesn't get stuck with firmware 14.0.1, either.

Although it may not be closely related, the suggested macro doesn't achieve what is intended. It's meant to tap `5` on tap, tap `1` on double-tap, and not interfere with other keys, such as `w`, according to the original forum post.

kareltucek:
> pressKey a doesn't get stuck with firmware 14.0.1, either.

Interesting. 

I stand corrected, this does:

```
pressKey a
noOp
```

>  It's meant to tap 5 on tap, tap 1 on double-tap

Ah, sorry. I was replying to Phil in that thread, not to you.

Regarding your usecase, what about this?

```
ifDoubletap final tapKey 1
tapKey 5
```

>  and not interfere with other keys, such as w, according to the original forum post.

Does it interfere with other keys? How?

pcooke9:
> > and not interfere with other keys, such as w, according to the original forum post.
> 
> Does it interfere with other keys? How?

I think he means this macro:

```
tapKey 5
ifGesture $thisKeyId tapKey 1
```

When holding the `W` key, and then single tapping the above macro to trigger `5`, the `W` key is stuck for a second. Even if you let go of the `W` key immediately after tapping the macro. However, doing the same with double-tap to trigger `1` doesn't produce a stuck `W`.
___


> Regarding your usecase, what about this?
> 
> ```
> ifDoubletap final tapKey 1
> tapKey 5
> ```

That macro you just provided works fine for me with no stuck `W`.😊

One potential quirk for both of the above macros though...
Double-tapping them outputs both a `5` and a `1` for every double-tap. That might not be the desired behavior if your usecase is to select an item in game. It might be fine for switching shotguns in DOOM, but not so much if you're trying to select instant-use items (think grenades and health potions).

I just checked the macro I'm using to double-tap to lock LShift (sprint):
```
ifDoubletap final pressKey persistent LS
releaseKey persistent LS
holdKey LS
```

And it outputs twice on double-tap as well. Since my current usecase is only for locking `LShift` or `LCtrl`, it's not really a big deal, but could be if used differently.

kareltucek:
But what does that have to do with this PR 🤦. @mondalaci , please either post your usecase concerns in a new forum topic, or create a new feature request or bug report ticket on github. This is ridiculous ;-).

mondalaci:
```
ifDoubletap final tapKey 1
tapKey 5
```

This works perfectly. Thank you!

PR #1299, opened 2025-07-04
Fix charger stopped detection.

### LINKED ISSUE #1298: UHK80 wrong battery level on firmware 14.2.1
After updating my UHK80 from 14.0.1 to 14.2.1 I noticed the battery level showing 100% for both sides for a couple of days.

Doing some investigation It seem that during the firmware update process the maximum charge value for the battery was altered.

Before updating, the battery level was around 86% left side and 92% right side, and using the `statsBattery` macro showed a maximum charge of ~4100mV.

After the update to 14.2.1 it is showing ~3800mV with the current charge ~being 4000mV which should explains why it is stuck at 100% for a long time.

kareltucek:
Closes #1298

Steps to reproduce:
- charge battery to 100% (say 4050 or so mV).
- now disconnect power and let it discharge. 
- observe that battery stays on 100% all the way down to 3800mV.
- also observe `Charger stopped bellow %dmV. This is suspicious!` logs. 

Still testing -> draft.

### DIFF
--- a/device/src/keyboard/charger.c
+++ b/device/src/keyboard/charger.c
@@ -316,10 +316,11 @@ void Charger_UpdateBatteryState() {
             setChargerEnPin(false);
 
             if (actuallyCharging) {
+                chargerStopped = false;
                 previousVoltage = getVoltage();
                 previousCharging = true;
             } else {
-                chargerStopped = actuallyEnabled && !actuallyCharging;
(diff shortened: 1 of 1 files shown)


kareltucek:
@mondalaci lets merge this? It almost certainly is an improvement...

PR #1300, opened 2025-07-05
Alter zero block action serialization to work with per-key-rgb.

kareltucek:
Here is the modification as discussed in https://github.com/UltimateHackingKeyboard/agent/issues/2421, 

Bumping the version to 11.1.0


mondalaci:
This works well according to my testing. My configuration used to occupy 17,939 bytes, including the per-key RGB lights, which are specifically used for testing purposes. Now, it occupies 12,441 bytes, a major improvement.

PR #1305, opened 2025-07-08
Add uhk60 support to build.sh

kareltucek:
I have added separate branches for uhk60 into the build.sh script...

Please, Benedek, merge it into your branch as you see fit.


benedekkupper:
What you might have missed is that I have added a new west extension, `west agent`, which is equivalent to `flashUsb`, but operates with the same parameters as `west flash`. So for flashing via debugger is done with `west flash` (for mcuxsdk as well), flashing via USB bootloader is done via `west agent`. Note that `west agent` is prepared to work with all devices already, but I haven't tested it yet (it just invokes the ts scripts with the correct arguments, which are printed on the command line for verification). So I recommend that with this in mind you can simplify how flashUsb works in this script.

build.sh
```diff
@@ -292,21 +299,74 @@ END
...
+            git submodule update --init --recursive
+            cd "$ROOT/.."
+            west init -l "$ROOT"
+            west config --local build.cmake-args -- "-Wno-dev"
+            west update
+            west patch
+            ;;
+        switchMcux)
+            west config manifest.file west_mcuxsdk.yml
+            # is west update && west patch needed?
```
benedekkupper: west update && west patch is needed once after checking out a new branch, but for both nRF SDK and MCUX SDK independently

build.sh
```diff
@@ -236,42 +228,57 @@ function createCentralCompileCommands() {
...
-            # update following according to README
-            git submodule init
-            git submodule update --init --recursive
-            cd "$ROOT/.."
-            west init -l "$ROOT"
-            west config --local build.cmake-args -- "-Wno-dev"
-            west update
-            west patch
+        flash)
+            echo "no idea how to flash mcux devices, please use flashUsb command"
```
benedekkupper: Flashing command is same for zephyr and mcux devices, `west flash --build-dir $BUILD_DIR`

benedekkupper:
I have cherry-picked the commits to the branch, closing PR

PR #1317, opened 2025-07-30
Increase uart thread priority.

kareltucek:
Phil has reported this to me. 

The issue: when using snap tap macros, the macros would keep the main thread busy constantly. Since macros are processed in the main thread whose priority is higher (0), the uart thread would not get a chance to run at all, leading to bridge cable timeouts, and consequently lagging and possibly stuck keys. 

A strange thing is that Phil says that it used to work fine. I don't see what is different now.

### DIFF
--- a/device/src/keyboard/uart.c
+++ b/device/src/keyboard/uart.c
@@ -18,7 +18,7 @@
 // Thread definitions
 
 #define THREAD_STACK_SIZE 1000
-#define THREAD_PRIORITY 5
+#define THREAD_PRIORITY -5
 
 #define UART_FOREVER_TIMEOUT 10000
 #define UART_RESEND_DELAY 100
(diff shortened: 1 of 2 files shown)


mondalaci:
Perhaps the main thread is busier than it was earlier due to some code changes, which do not leave the UART thread with sufficient performance.

How to test this one? Feel free to merge it if you've already tested it.

kareltucek:
@mondalaci that sure is possible.

Still UART thread should have higher priority.

kareltucek:
Regarding testing, using Phil's configuration is the easiest. Yes, I have tested it.

pcooke9:
Just for reference here, I should probably mention that up until the other day, I hadn't played any games that need the snap-tap macros for a couple of months or so, so I'm unsure which FW the issue began with. Back when Max first created the macro, I don't recall ever having any issues with it.

I also wanted to confirm the increased UART priority is _much_ better! Using the snap-tap macros still occasionally induces lag and stuck keys, but a _lot_ less frequently. I no longer see a bunch of link reset messages in the Zephyr logs, but I did see:
```
2025-07-30 18:11:15 | UHK 80 right | Uart: didn't receive ack 308615, resending
```
which I think corresponded with a heavy lag (several seconds) and a stuck key.

PR #1318, opened 2025-07-30
Add bluetooth reconnect command.

### LINKED ISSUE #1314: Feature Request: Macro command to trigger a reconnect
Rather than requiring a dedicated hotkey to force a connection to a specific device/dongle/ble, a command that will trigger a reconnect to the currently selected device would be useful.

kareltucek:
Closes https://github.com/UltimateHackingKeyboard/firmware/issues/1314 .

### DIFF
--- a/device/src/bt_conn.c
+++ b/device/src/bt_conn.c
@@ -813,6 +813,18 @@ void BtConn_DisconnectAll() {
     bt_conn_foreach(BT_CONN_TYPE_LE, bt_foreach_conn_cb, NULL);
 }
 
+void BtConn_DisconnectOne(connection_id_t connectionId) {
+    uint8_t peerId = Connections[connectionId].peerId;
+
+    if (peerId == PeerIdUnknown) { return; }
+
+    struct bt_conn *conn = Peers[peerId].conn;
+
+    if (!conn) { return; }
+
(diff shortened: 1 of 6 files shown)


mondalaci:
How does this differ from the [lastSelected switchHost target](https://github.com/UltimateHackingKeyboard/firmware/pull/1315) exactly?

kareltucek:
This:
- acts on a current connection
- disconnects the connection before reconnecting. 

lastSelected scenario:
- user taps switchHost A
- UHK connects to A
- UHK looses connection and disconnects
- UHK falls back to another connection, probably USB
- maybe another bluetooth device, say B, tries to grab the now free UHK, and UHK switches over to it B
- user comes back to the computer, and sees the UHK connected to B
- now:
  - the user can tap `switchHost lastSelected`, which makes UHK try to connect to A again. 
  - the user can tap `reconnect`, which disconnects B and then tries to connect to B again.

reconnect scenario:
- user taps switchHost A
- UHK connects to A
- UHK looses connection, but doesn't detect it
- UHK still thinks it is connected to A, but scancodes don't reach A
- now:
  - the user can tap `switchHost lastSelected`, which makes UHK try to connect to A. UHK finds out it is already connected and does nothing
  - the user can tap `reconnect`, which forces zephyr to terminate the connection, then tries to establish connection with A anew.

Issue #1228, opened 2025-05-12
Unable to pair BLE devices on f/w 14.0.0


tunney:
I removed all my BLE devices on my keyboard, then when trying to readd I cannot pair them. Never get the prompt on the keyboard for the pairing key.

Rolling back to 13.0.2 resolves this.

mondalaci:
I'm on firmware 14.0.0, I don't have any BLE devices or dongles connected to my UHK 80, I can pair my Android phone, and both the keyboard and mouse functionality work well.

Do you have any dongles connected?

What's the operating system of the host you're trying to connect to?

Please provide the Zephyr log while trying to connect. Click on the Agent logo on the About page 7 times, then go to "Advanced settings", click on "Zephyr logging", then check "Right half".

tunney:
MacOs Sequoia 15.4.1
2023 MacBook nPro with M3 Max

No Dongles connected.
No devices paired.

I would provide the logs and all that but https://github.com/UltimateHackingKeyboard/firmware/issues/1226

Last week I managed after a lot of attempts to flash 14.0.0 to the left side, yesterday no joy.

Could not pair with iPhone 15 Pro or the Mac yesterday on 14.0.0. Paired when downgraded to 13.0.2

tunney:
now on MacOs 15.5

I used the workaround in the "Bricked" thread to get 14.0.0 back on the uhk80

Removed everything that was paired with the keyboard (this does not work cleanly)
Then tried to pair the MacBook

Saw UHK 80 and clicked "Connect@

<img width="478" alt="Image" src="https://github.com/user-attachments/assets/ae1a7487-6066-4f23-b755-adb734b3d9b2" />

Never ending spinners with no dialog on the uHK80

Turn off BT on the mac and back on and now see "UHK 80 BLE" with no option to pair

<img width="480" alt="Image" src="https://github.com/user-attachments/assets/66e1bc27-2a95-4226-91ad-473359a1c911" />

ldriscoll:
I run my uhk 80 with two dongles, one for linux, and one for a mac on 15.4.1. When running with the 14.0.0 firmware I'm unable to connect to the dongle on the mac, however going back to 13.0.2 everything works fine. 

On 14.0.0: Initially it connected to the mac dongle ok, but after switching to the linux dongle it was unable to reattach to the mac dongle. I removed the dongles from the connection, and readded them, a number of times, to no avail.

Sorry, I don't have any logs.

[...]

PR #1328, opened 2025-08-12
migrate nRF build to sysbuild

benedekkupper:
All behavior should be the same as before. The project is (re)organized in a way that currently the legacy and sysbuild are both functional.


mondalaci:
Please let me know when we can test this.

benedekkupper:
If you are brave/curious/impatient, you can try already, I only managed to put it together just before a week long break, so from my side there won’t be a chance to move this forward until I’m back.

mondalaci:
This build has issues: https://github.com/UltimateHackingKeyboard/firmware/issues/1228#issuecomment-3183206650

benedekkupper:
Other changes are split from this PR, this branch now has only build related changes (and removing "USB" from the BT model name of UHK80 right).

benedekkupper:
@kareltucek please review 6d9272d

kareltucek:
@benedekkupper yep it is fine. 

(I am a bit confused by the code as I would expect the debug_eventloop_timing.c code to be ifdeffed the way the thread_stack.c is, as those features are mutually exclusive anyways, but that is my problem and I will refactor that when the code is actually needed.)

Issue #1331, opened 2025-08-22
Smart Macros: retroactive function of various commands.


firngrod:
With the following macro assigned to key `A`
```
while (1) {
 ifReleased final tapKey a
 ifNotKeyActive $thisKeyId final tapKey b
 ifPlaytime 1500 final tapKey c
}
```
and another key, just to demonstrate
```
postponeKeys delayUntilRelease
```
the following oddities can be observed:

- When `A` is pressed alone and on it's own, it will output `a` or `c` as expected
- When `A` is pressed during postponing, playTime is counted from key press, not from dequeue time.  This may me as intended, but it's unexpected to me.  Otherwise works as expected and outputs `a` or `c` as expected
- When `A` is pressed and released while postponed, `ifReleased` and `ifNotKeyActive $thisKeyId` will never evaluate true and the macro will only ever output `c`

Running UHK60 V1 with firmware 15.1.0

firngrod:
In an attempt to find a work-around for this using `delayUntilReleaseMax`, I have discovered that while it seems to work as a way of detecting release in that it exits immediately when the key is released in all cases that I have tested, the time-out appears to be from macro start, not from command start, every time, including the same issue with `ifPlaytime` here, such that the macro below exits after one second, not 3 as one might expect

```
delayUntil 1000
delayUntilReleaseMax 1000
delayUntilReleaseMax 1000
write 'done'
```

I had hoped to use it to detect releases by delaying max 5ms and comparing time before to time after.

firngrod:
Man, I found such a good work-around for this for my use-case that I'm almost going to be sad when you fix it.  :)

For anyone reading and wondering, in your macros, just before yielding to the queue, if the next pending key has been released, save it's ID, and then test $thisKeyId against that stored value as well when testing ifReleased in your macros.

kareltucek:
> When A is pressed during postponing, playTime is counted from key press, not from dequeue time. This may me as intended, but it's unexpected to me. Otherwise works as expected and outputs a or c as expected

I think this is expected. Macro engine / postponer replay the events as if we weren't messing with time. I.e., from the point of view of the macro the macro was started when the key was pressed. 

A less clear question is what time a postponed macro should see as _now_ when performing those checks. 

> When A is pressed and released while postponed, ifReleased and ifNotKeyActive $thisKeyId will never evaluate true and the macro will only ever output c

I see. The bug is that the while doesn't yield at the end of the executed block. 

Fixing.

[...]

PR #1345, opened 2025-09-16
Make while macro command yield.

kareltucek:
Fixes a bug reported in https://github.com/UltimateHackingKeyboard/firmware/issues/1331 

Changelog: 
- Make macro "while" command yield after every iteration.


kareltucek:
This increases my benchmarks from ~3100ms to ~4000ms. The good news is that this is more a bug of a benchmark than new parsing issues.

Issue #1039, opened 2024-12-17
Weird scrolling behaviour in firmware 12.0.0


abonhote:
Hi all

Since I updated to 12.0.0 today, I'm experiencing a weird mouse wheel behaviour, at least while using Windows 10. Not all applications are affected. The developer console of the UHK agent seems to work fine, while my SQL tool of choice (allroundautomations.com => PL/SQL developer) refuses to scroll. Also notepad++ has problems scrolling, it needs two keystrokes to scroll down.

I'd love to give you more information, but I'm not sure how. Please tell me what you need to investigate on it.

Cheers

André

kareltucek:
My guess this is related to the new smooth scrolling feature @mondalaci ?

mondalaci:
I agree that hi-res scrolling is probably the culprit since 12.0.0 is the first firmware release with #794 merged.

Related issue: #741

I can reproduce the issue in Notepad++ on Windows 11.

@rightaditya @mhantsch Any insights?

abonhote:
Is there a way to turn this off?

mondalaci:
There's no way to turn this off. Downgrading Agent and the firmware is the only way.

mhantsch:
@abonhote how are you triggering the mouse wheel? Using a module (Touchpad, trackball, trackpoint, mini-trackball) or from key presses?

What are your mouse-relevant settings in Agent?

abonhote:
Keys only, normally. The trackball and the touch pad seem to work fine.

 Can't send any config before tomorrow.

rightaditya:
Is the scrolling smooth in Notepad++ with the trackball? Or is it the same as before the upgrade? Unfortunately I can't test myself as my Windows install is currently not working properly.

I'd have to check the struct details but I think the fastest fix would be to disable the scroll multiplier entirely for mouse events from keys. As it stands the check is for `isScroll` on the kinetic state to decide whether or not to use it.

@abonhote The trackball and touchpad are fine for *both* Notepad++ and the SQL tool, right?

rightaditya:
Hmm, OK, it won't be so simple to revert *just* for the mouse keys since they still need to send values scaled according to the OS's (new) expectation.

[...]

PR #1352, opened 2025-09-18
Add configuration of mouse keys first tick behavior.

kareltucek:
Implements some of the behavior discussed in https://github.com/UltimateHackingKeyboard/firmware/issues/1039#issuecomment-3302241930 .


kareltucek:
We probably don't want this, do we?

PR #1354, opened 2025-09-18
Add GitPython as development prerequisite in readme

rightaditya:
It's required by `scripts/west_patch.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,12 @@ this firmware uses the [west workspace](https://docs.zephyrproject.org/latest/de
 create a wrapping directory, which will store the firmware git repository, and the *west workspace*
 with the third-party SW components.
 
-Here is the initial checkout:
+Here is the initial checkout and installation of required Python packages:
 ```bash
 mkdir uhk-workspace
 cd uhk-workspace
 git clone --recurse-submodules git@github.com:UltimateHackingKeyboard/firmware.git
+pip3 install -r firmware/scripts/requirements.txt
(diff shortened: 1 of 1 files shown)


kareltucek:
They are in scripts/requirements.txt

kareltucek:
But yes, it should be mentioned in the readme

kareltucek:
This is the sequence that I use from build.sh, if it's of any help:

```
            pip3 install west
            pip3 install -r scripts/requirements.txt
            nrfutil install toolchain-manager
            nrfutil toolchain-manager install --ncs-version $NCS_VERSION
            git submodule init
            git submodule update --init --recursive
            cd "$ROOT/.."
            west init -l "$ROOT"
            west config --local build.cmake-args -- "-Wno-dev"
            west update
            west patch
```

mondalaci:
requirements.txt should be used and featured in the readme.

rightaditya:
~~Mentioning the `git submodule` commands in the readme would be nice too, since it includes the `git clone` command. I realized that was necessary when the compile failed and I found the bootloader directory empty 😅~~

EDIT: derp, missed the `--recurse-submodules` in the `git clone` command :melting_face:

A note that building Agent is required to flash might be helpful also. To be clear, the error messages for both of these made it clear what was wrong (more for Agent than the bootloader, though)—it's not like I had to do any googling to figure out what was wrong.

rightaditya:
I just modified this to include an explicit `requirements.txt` installation step.

PR #1361, opened 2025-09-24
DOCS: improve build instructions.

kareltucek:
For [Firngrod](https://forum.ultimatehackingkeyboard.com/u/Firngrod)'s review ;-).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,15 @@ If you're one of the brave few who wants to hack the firmware then read on.
     - `scripts`
       - `make-release.mjs` - script to build a full release tarball
 
-### Prerequisities (for both UHK60 and UHK80):
+### Dependencies (for both UHK60 and UHK80):
 
 - git
 - pip3
-- nodejs (and optionally nvm for version management, then (e.g.) `nvm install 20; nvm use 20`)
-- west and some other python packages (`pip3 install -r scripts/requirements.txt`)
-- (UHK60 and modules) gcc-arm-none-eabi toolchain:
+- ninja
+- cmake
+- nodejs (and optionally nvm for version management, then (e.g.) `nvm install 22; nvm use 22`)
+- west and some other python packages (`pip3 install -r scripts/requirements.txt` after you have cloned the repository)
+- (UHK60 and modules) gcc-arm-none-eabi (sometimes also called arm-none-eabi-gcc) toolchain:
   after it is installed, set an environment variable in the default shell, e.g.
   - `export ARM_GCC_DIR="/usr"` for Linux or WSL in ~/.bashrc
   - `export ARM_GCC_DIR="/opt/homebrew"` for macOS in ~/.zshrc
@@ -99,7 +101,7 @@ Basic actions (see help for more):
 - `build` - full pristine build
 - `make` - incremental build
 - `flash` - flash via debug probe, consider setting up `.devices` file. See `./build.sh help`.
-- `flashUsb` - flash via USB (UHK80 only)
+- `flashUsb` - flash via USB
 - `release` - build full release tarball
 
 Release:
@@ -109,6 +111,8 @@ Release:
 ```
 
 ### Manual workspace setup
+
+_Note: this and following sections are redundant If you have successfully completed above build.sh procedure._
 
 Unlike most common workflows, where the git repository is the top level directory,
 this firmware uses the [west workspace](https://docs.zephyrproject.org/latest/develop/west/workspaces.html#t2-star-topology-application-is-the-manifest-repository) structure. This means that you should first
--- a/README.md
+++ b/README.md
@@ -16,12 +16,23 @@ If you want to use the latest firmware version for your UHK, then instead of goi
 
 If you're one of the brave few who wants to hack the firmware then read on.
 
-### Global prerequisities (for both UHK60 and UHK80):
+### Directory structure:
+
+- `uhk-workspace` - west workspace that manages both uhk firmware repository and third party components
+  - `zephyr, c2usb, bootloader, nrf` - third party components
+  - `firmware` - the actual uhk repository
+    - `lib/agent` - Agent as a git submodule
+      - `.nvmrc` - expected Node.js version
+      - `packages/usb` various development scripts to communicate with the UHK over USB
+    - `scripts`
+      - `make-release.mjs` - script to build a full release tarball
+
+### Prerequisities (for both UHK60 and UHK80):
 
 - git
 - pip3
 - nodejs (and optionally nvm for version management, then (e.g.) `nvm install 20; nvm use 20`)
-- west (`pip3 install west`)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
-### Setup workspace
+Valid targets:
+- UHK80: `right`, `left`, `dongle`
+- UHK60: `rightv1`, `rightv2`, `left`, `keycluster`, `trackball`, `trackpoint`
+
+Basic actions (see help for more):
+- `build` - full pristine build
+- `make` - incremental build
+- `flash` - flash via debug probe, consider setting up `.devices` file. See `./build.sh help`.
+- `flashUsb` - flash via USB (UHK80 only)
```
firngrod: I use `flashUsb` to flash to my UHK 60v1

scripts/requirements.txt
```diff
@@ -1,3 +1,4 @@
+west
```
firngrod, line 1:
```suggestion
west
ninja
cmake
```

README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
+cd firmware
+./build.sh setup
+```
+
+(Optional) Update and build Agent (we don't update the submodule reference as often as we should):
+```
+cd lib/agent
+git fetch origin && git checkout origin/master
+nvm use 22 && npm ci && npm run build
+cd ../..
```
firngrod, line 69: Thank you for putting me back the right folder!

README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
+    - `scripts`
+      - `make-release.mjs` - script to build a full release tarball
+
+### Prerequisities (for both UHK60 and UHK80):
 
 - git
 - pip3
 - nodejs (and optionally nvm for version management, then (e.g.) `nvm install 20; nvm use 20`)
-- west (`pip3 install west`)
+- west and some other python packages (`pip3 install -r scripts/requirements.txt`)
```
firngrod: It feels a little anachronistic referencing files in the repo as a pre-requisite.

README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
 - (UHK60 and modules) gcc-arm-none-eabi toolchain:
```
firngrod: This package is called arm-none-eabi-gcc under Arch Linux.  Maybe mention alternative name?  Or not, you might not want to start down the road of catering to all distros.  Also, if you use Arch, you're asking for it.

build.sh
```diff
@@ -229,22 +229,34 @@ function createCentralCompileCommands() {
...
+    if [ "$ARM_GCC_DIR" == "" ]
+    then
+        if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+            export ARM_GCC_DIR="/usr"
+        elif [[ "$OSTYPE" == "darwin"* ]]; then
+            export ARM_GCC_DIR="/opt/homebrew"
+        fi
+    fi
+}
+
```
firngrod, line 262: Nice.

README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
+      - `.nvmrc` - expected Node.js version
+      - `packages/usb` various development scripts to communicate with the UHK over USB
+    - `scripts`
+      - `make-release.mjs` - script to build a full release tarball
+
+### Prerequisities (for both UHK60 and UHK80):
 
 - git
 - pip3
 - nodejs (and optionally nvm for version management, then (e.g.) `nvm install 20; nvm use 20`)
```
firngrod: Is 20 used/needed?  I'm pretty sure I ran it with 22 without any issue.

README.md
```diff
@@ -16,21 +16,99 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
+- `flashUsb` - flash via USB (UHK80 only)
+- `release` - build full release tarball
+
+Release:
+
+```
+./build.sh release
+```
+
+### Manual workspace setup
```
firngrod, line 113: I suggest very clearly signalling that this section is redundant if the user successfully completed the previous section.

firngrod (changes requested):
Thank you for taking the time to bring these instructions up to date!

I can only talk about the UHK60 stuff, I don't own a UHK80.

I have only addressed the new instructions.

With the addition of `ninja` and `cmake` in `requirements.txt`, I was able to build and flash.  `ninja` was just missing.  `cmake` could be installed globally, but in my first dev env, that did not work out for me for some reason, so apparently not all `cmake`s are built the same.

Beyond that, I had bit of a learning journey with `pip3` under Arch.  It will not install packages outside of an active python virtual environment without a stern talking to.  I guess this will be the same for all Arch derived distros (Manjaro, EnveavorOS, SteamOS etc), and possibly other distros as well.  Furthermore, I think it's just good practice to use virtual environments.
However, VSCode does not play nice with virtual environments out of the box.  It needs to be launced with the venv already activated, which is not going to be the workflow for many, if anyone at all.
I will see if I can find a reasonable solution to that, and if I can, I will make a PR or something.  Until then, we leave that as an exercise for the reader.

scripts/requirements.txt
kareltucek: ninja and cmake are not python native, and should probably be installed from the native system package manager, shouldn't they? 

(I am listing them above in the normal dependencies section.)

kareltucek:
I think all should be addressed?

scripts/requirements.txt
firngrod: Installing them from the package manager is fine as long as we don't depend on a specific version.  The strengths of installing them with `pip3` are 1)  It's easy, and the command is platform independent and 2)  We could use a version different from system version, or from other projects on the computer when using python venvs.
We need neither, and we don't use venvs, so just have them be prerequisites, yes.

README.md
```diff
@@ -16,21 +16,103 @@ If you want to use the latest firmware version for your UHK, then instead of goi
...
+### Dependencies (for both UHK60 and UHK80):
 
 - git
 - pip3
-- nodejs (and optionally nvm for version management, then (e.g.) `nvm install 20; nvm use 20`)
-- west (`pip3 install west`)
-- (UHK60 and modules) gcc-arm-none-eabi toolchain:
+- ninja
+- cmake
+- nodejs (and optionally nvm for version management, then (e.g.) `nvm install 22; nvm use 22`)
```
firngrod, line 36: Actually, I think the requirement should be nvm with nodejs 22 installed.  The scripts depend on it.

firngrod (commented):
Two more things:
The nvm comment, and a nitpick:  `build.sh setup` installs west, and then installs `requirements.txt` which includes west.  Harmless redundancy.

kareltucek:
Let's merge this, I am already missing some bug fixes ;-).

PR #1362, opened 2025-09-24
Fix UHK60 backlight sleep mode.

### LINKED ISSUE #1176: UHK60v2: LED Segment Display Remains On After Computer Turned Off
A similar issue was first described in https://github.com/UltimateHackingKeyboard/firmware/issues/1123, however that issue has since been closed.

In my setup I have the keyboard connected to a powered USB hub, which is connected to my computer. When turning off the computer, the LED segment display (which by default shows that current keymap) **remains on**, despite key backlight LEDs being correctly powered off.

This appears to be related to v12.x firware versions (as I did not observe it until flashing v12.3.3 recently and was previously on v11.2.0). Note that I am unable to downgrade to a v11.x firmware, as the Agent v4 cannot detect the keyboard and Agent v5 cannot flash v11 firmware.

As a workaround, I currently set the LED display brightness to 0, which is not ideal. Note that setting a LED display timeout does not seem to have an effect in this state (the LED segment display remains on even after the timeout).

kareltucek:
Closes #1176 

Reproduction steps:
- set backlight fade timeout on a UHK60 to say 4 seconds
- connect uhk60 into a dumb usb charger (the older, the better)
- observe that after four seconds it doesn't turn its backlight of
- in case it does, tap some key (a one that has a scancode mapped)
- again observe that after four seconds it doesn't turn its backlight of

### DIFF
--- a/right/src/usb_interfaces/usb_interface_basic_keyboard.c
+++ b/right/src/usb_interfaces/usb_interface_basic_keyboard.c
@@ -27,6 +27,7 @@ bool UsbBasicKeyboard_NumLockOn = false;
 bool UsbBasicKeyboard_ScrollLockOn = false;
 
 static bool needsResending = false;
+static uint8_t retries = 0;
 
 usb_hid_protocol_t usbBasicKeyboardProtocol;
 
@@ -205,7 +206,7 @@ void UsbBasicKeyboardSendActiveReport(void)
     UsbReportUpdateSemaphore |= 1 << USB_BASIC_KEYBOARD_INTERFACE_INDEX;
     usb_status_t status = UsbBasicKeyboardAction();
     //The semaphore has to be set before the call. Assume what happens if a bus reset happens asynchronously here. (Deadlock.)
-    if (status != kStatus_USB_Success) {
+    if (ShouldResendReport(status == kStatus_USB_Success, &retries)) {
         //This is *not* asynchronously safe as long as multiple reports of different type can be sent at the same time.
         //TODO: consider either making it atomic, or lowering semaphore reset delay
         needsResending = true;
--- a/right/src/main.c
+++ b/right/src/main.c
@@ -83,7 +83,8 @@ static void initConfig()
 static void sendFirstReport()
 {
     bool success = false;
-    while (!success) {
+    // Wait until sending a report is successful, but don't block longer than 5 seconds.
+    while (!success && Timer_GetCurrentTime() < 5000) {
         UsbReportUpdateSemaphore |= 1 << USB_BASIC_KEYBOARD_INTERFACE_INDEX;
         usb_status_t status = UsbBasicKeyboardAction();
         if (status != kStatus_USB_Success) {
@@ -94,7 +95,7 @@ static void sendFirstReport()
         }
     }
     while (UsbReportUpdateSemaphore) {
-            __WFI();
+        __WFI();
     }
 }
 
--- a/right/src/usb_interfaces/usb_interface_media_keyboard.c
+++ b/right/src/usb_interfaces/usb_interface_media_keyboard.c
@@ -4,12 +4,12 @@
 #include "utils.h"
 #include "event_scheduler.h"
 
-
 uint32_t UsbMediaKeyboardActionCounter;
 static usb_media_keyboard_report_t usbMediaKeyboardReports[2];
 usb_media_keyboard_report_t* ActiveUsbMediaKeyboardReport = usbMediaKeyboardReports;
 
 static bool needsResending = false;
+static uint8_t retries = 0;
 
 static usb_media_keyboard_report_t* GetInactiveUsbMediaKeyboardReport(void)
 {
@@ -46,7 +46,7 @@ usb_status_t UsbMediaKeyboardAction(void)
 void UsbMediaKeyboardSendActiveReport(void) {
     UsbReportUpdateSemaphore |= 1 << USB_MEDIA_KEYBOARD_INTERFACE_INDEX;
     usb_status_t status = UsbMediaKeyboardAction();
-    if (status != kStatus_USB_Success) {
+    if (ShouldResendReport(status == kStatus_USB_Success, &retries)) {
         UsbReportUpdateSemaphore &= ~(1 << USB_MEDIA_KEYBOARD_INTERFACE_INDEX);
         EventVector_Set(EventVector_ResendUsbReports);
         needsResending = true;
--- a/right/src/usb_interfaces/usb_interface_mouse.c
+++ b/right/src/usb_interfaces/usb_interface_mouse.c
@@ -4,12 +4,14 @@
 #include "event_scheduler.h"
 #include "attributes.h"
 #include "macros/status_buffer.h"
+#include "utils.h"
 
 #ifdef __ZEPHYR__
 #include "usb/usb_compatibility.h"
 #endif
 
 static bool needsResending = false;
+static uint8_t retries = 0;
 
 #include "usb_descriptors/usb_descriptor_mouse_report.h"
 
@@ -178,7 +180,7 @@ void UsbMouseSendActiveReport(void)
 #else
     UsbReportUpdateSemaphore |= 1 << USB_MOUSE_INTERFACE_INDEX;
     usb_status_t status = UsbMouseAction();
-    if (status != kStatus_USB_Success) {
+    if (ShouldResendReport(status == kStatus_USB_Success, &retries)) {
         UsbReportUpdateSemaphore &= ~(1 << USB_MOUSE_INTERFACE_INDEX);
         EventVector_Set(EventVector_ResendUsbReports);
         needsResending = true;
--- a/right/src/usb_interfaces/usb_interface_system_keyboard.c
+++ b/right/src/usb_interfaces/usb_interface_system_keyboard.c
@@ -2,12 +2,14 @@
 #include "usb_composite_device.h"
 #include "usb_report_updater.h"
 #include "event_scheduler.h"
+#include "utils.h"
 
 #ifndef UTILS_ARRAY_SIZE
 #define UTILS_ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
 #endif
 
 static bool needsResending = false;
+static uint8_t retries = 0;
 
 uint32_t UsbSystemKeyboardActionCounter;
 static usb_system_keyboard_report_t usbSystemKeyboardReports[2];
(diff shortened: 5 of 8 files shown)


jpyllman:
I have not tested this yet. But to me this code does not look good/safe style. I have not looked at the full code for the firmware. So I can not say if it follows the style of not.

Question: what does Timer_GetCurrentTime() really return??

Also, I comment that this happens when there is nobody that responds to the USB. So to me it should more be that it turns off even if there is an error on the USB. I guess that the thing that is sent on the USB is telling backlight turned off?

kareltucek:
> I have not tested this yet. But to me this code does not look good/safe style.

Can you elaborate? 

> Question: what does Timer_GetCurrentTime() really return??

Uptime in milliseconds.

>  I guess that the thing that is sent on the USB is telling backlight turned off?

No. It is a normal usb report, supposedly containing a report about user having tapped some key.

-----------------

The actual problems is that sending usb report also trips the "user has done something" logic, and therefore extends sleep timer. 

The idea of the actual change above is that if a report fails to deliver for 250ms straight, then it is probably no longer relevant. 

It is true that when the usb fails for other reasons than a timeout, it may be much less than 250ms. 

So my conclusion is that the resending logic should trigger based on time rather than number of attempts.

Does that sound better?

jpyllman:
> > I have not tested this yet. But to me this code does not look good/safe style.
> 
> Can you elaborate?

Maybe it is just that I am more used to read/write c++. But having static variables and anyway sending pointers to them in a function call feels redundant. And also, updating static is not supposed to be done??

> > Question: what does Timer_GetCurrentTime() really return??
> 
> Uptime in milliseconds.

So the inclusion of that in the if is to not perform any check before the first 5s after getting power to the keyboard??

> > I guess that the thing that is sent on the USB is telling backlight turned off?
> 
> No. It is a normal usb report, supposedly containing a report about user having tapped some key.
> 
> The actual problems is that sending usb report also trips the "user has done something" logic, and therefore extends sleep timer.
> 
> The idea of the actual change above is that if a report fails to deliver for 250ms straight, then it is probably no longer relevant.
> 
> It is true that when the usb fails for other reasons than a timeout, it may be much less than 250ms.
> 
> So my conclusion is that the resending logic should trigger based on time rather than number of attempts.
> 
> Does that sound better?

OK. So why is this usb report thing only happen when keyboard is not connected?? Since it is only then that the keyboard stay awake. Maybe that is where to start instead of kind of skip errors. Maybe making the usb report transparent and not count to keep alive.

Sorry if my feedback sounded negative, this patch is kind of the only code in UHK I have looked at. So I have no feel of the overall look/feel of this code.

kareltucek:
> Maybe it is just that I am more used to read/write c++. But having static variables and anyway sending pointers to them in a function call feels redundant. 

They are distinct static variables from multiple compilation units, so the pointer is actually necessary.

The static basically says "Don't pollute global namespace by this variable.".

> And also, updating static is not supposed to be done??

I have never heard of that.

Can you elaborate on the reasons?

> > > Question: what does Timer_GetCurrentTime() really return??
> > 
> > Uptime in milliseconds.
> 
> So the inclusion of that in the if is to not perform any check before the first 5s after getting power to the keyboard??

This code is part of the startup sequence, before the normal eventloop processing begins. 

It's purpose is to wait until usb connection is established, and only then allow the eventloop to start running. It prevents a bunch of bugs that are minor, but still annoying.

One of the troubles had been that if there was no host, this while loop would spin indefinitely.

In order to recalculate sleeping logic, backlight states, etc., we need to get into the standard eventloop.

So this code now says: "When starting the keyboard, wait until one empty usb report is successfully delivered, but don't block for more than 5 seconds."

> 
> > > I guess that the thing that is sent on the USB is telling backlight turned off?
> > 
> > No. It is a normal usb report, supposedly containing a report about user having tapped some key.
> > The actual problems is that sending usb report also trips the "user has done something" logic, and therefore extends sleep timer.
> > The idea of the actual change above is that if a report fails to deliver for 250ms straight, then it is probably no longer relevant.
> > It is true that when the usb fails for other reasons than a timeout, it may be much less than 250ms.
> > So my conclusion is that the resending logic should trigger based on time rather than number of attempts.
> > Does that sound better?
> 
> OK. So why is this usb report thing only happen when keyboard is not connected?? Since it is only then that the keyboard stay awake. Maybe that is where to start instead of kind of skip errors.
 
Because when the keyboard is not connected, then the usb report never reaches the host and the keyboard keeps resending it forever.

> Maybe making the usb report transparent and not count to keep alive.

What exactly do you mean by the word "transparent" here? 

Transparent to whom? And in the sense of being invisible, or in the sense of not hiding whatever is behind it?

> Sorry if my feedback sounded negative, this patch is kind of the only code in UHK I have looked at. So I have no feel of the overall look/feel of this code.

Nonono, it is actually very appreciated. This codebase suffers a lot by not receiving proper code reviews :-).

jpyllman:
> > Maybe it is just that I am more used to read/write c++. But having static variables and anyway sending pointers to them in a function call feels redundant.
> 
> They are distinct static variables from multiple compilation units, so the pointer is actually necessary.
> 
> The static basically says "Don't pollute global namespace by this variable.".
>
> > And also, updating static is not supposed to be done??
> 
> I have never heard of that.
> 
> Can you elaborate on the reasons?

OK, I seem to get confused because now when I use static in C++ it is together with constexpr to create data structure with const data. But I remember using it for 'top-level' of storage of state and stuff.
  
> > > > Question: what does Timer_GetCurrentTime() really return??
> > > 
> > > 
> > > Uptime in milliseconds.
> > 
> > 
> > So the inclusion of that in the if is to not perform any check before the first 5s after getting power to the keyboard??
> 
> This code is part of the startup sequence, before the normal eventloop processing begins.
> 
> It's purpose is to wait until usb connection is established, and only then allow the eventloop to start running. It prevents a bunch of bugs that are minor, but still annoying.
> 
> One of the troubles had been that if there was no host, this while loop would spin indefinitely.
> 
> In order to recalculate sleeping logic, backlight states, etc., we need to get into the standard eventloop.
> 
> So this code now says: "When starting the keyboard, wait until one empty usb report is successfully delivered, but don't block for more than 5 seconds."
> 

OK, it is basically because of everything living in the event loop. I get that.

> > > > I guess that the thing that is sent on the USB is telling backlight turned off?
> > > 
> > > 
> > > No. It is a normal usb report, supposedly containing a report about user having tapped some key.
> > > The actual problems is that sending usb report also trips the "user has done something" logic, and therefore extends sleep timer.
> > > The idea of the actual change above is that if a report fails to deliver for 250ms straight, then it is probably no longer relevant.
> > > It is true that when the usb fails for other reasons than a timeout, it may be much less than 250ms.
> > > So my conclusion is that the resending logic should trigger based on time rather than number of attempts.
> > > Does that sound better?
> > 
> > 
> > OK. So why is this usb report thing only happen when keyboard is not connected?? Since it is only then that the keyboard stay awake. Maybe that is where to start instead of kind of skip errors.
> 
> Because when the keyboard is not connected, then the usb report never reaches the host and the keyboard keeps resending it forever.
> 
> > Maybe making the usb report transparent and not count to keep alive.
> 
> What exactly do you mean by the word "transparent" here?
> 
> Transparent to whom? And in the sense of being invisible, or in the sense of not hiding whatever is behind it?

What I meant was that some USB communication would not effect the state of being used or not. That the event loop could still spin around without keeping the state alive.

Together with  the 5s startup thing, it feels to me that the current way this works will not handle the USB connection moving around. Being connected to a docking station and other solutions where you can move between computer and other 'devices' might need other solutions.

To me seeing the USB communication would be nice. With the docking station I have, things generally seem to work with windows machines. But not 100% with Linux. But the backlight staying on is always the same problem, since no connected HID device to talk to.

> > Sorry if my feedback sounded negative, this patch is kind of the only code in UHK I have looked at. So I have no feel of the overall look/feel of this code.
> 
> Nonono, it is actually very appreciated. This codebase suffers a lot by not receiving proper code reviews :-).

I am trying to become a better reviewer. I generally 'hate' it. But at the same time, we need to work together to make a better place.

jpyllman:
So, now I had time to test the new firmware. And it seemed that I moved from 15.0.1 to firmware with this patch. And it looks like keeping the backlight off when not connected to computer. So actually, now when disconnecting computer from dock the backlight turns off and I can not get it on until I connect a computer. This is fine.

Going back to the 15.2.0 firmware kind of works the same. But I actually somehow managed to get the backlight turned on again with the 15.2.0 version, and then it does not turn off. I am talking about when computer disconnected form dock.

So, 15.2.0 seem to already solve of the dock problems I have seen. But this patch seem to really keep it off if not connected to a computer.

right/src/usb_interfaces/usb_interface_basic_keyboard.c
```diff
@@ -205,7 +206,7 @@ void UsbBasicKeyboardSendActiveReport(void)
     UsbReportUpdateSemaphore |= 1 << USB_BASIC_KEYBOARD_INTERFACE_INDEX;
     usb_status_t status = UsbBasicKeyboardAction();
     //The semaphore has to be set before the call. Assume what happens if a bus reset happens asynchronously here. (Deadlock.)
-    if (status != kStatus_USB_Success) {
+    if (ShouldResendReport(status == kStatus_USB_Success, &retries)) {
```
nlordell, line 209: AFAICT, the `retries` counter never gets cleared. I'm not familiar enough with how the firmware works, and when it gets restarted to know whether or not this is an issue.

kareltucek:
> What I meant was that some USB communication would not effect the state of being used or not. 

Yes, I can add that. I didn't do that originally to not clutter the code.

> That the event loop could still spin around without keeping the state alive.

We do not want to spin it for no reason, just as we don't want to send a scancode for a key that was briefly tapped yesterday afternoon ;-) .

> Together with the 5s startup thing, it feels to me that the current way this works will not handle the USB connection moving around.

Can you elaborate?

> AFAICT, the retries counter never gets cleared.

Great catch!

kareltucek:
Some fixes pushed. What about now? 😃

jpyllman:
> > Together with the 5s startup thing, it feels to me that the current way this works will not handle the USB connection moving around.
> 
> Can you elaborate?

What I think is that there is difference in starting from no power and getting a new connection, but not new power. And I just think that this 5s delay is only when new powerer. Not when you have new connection. And maybe that is what is needed. But, it feels like there might be a 'bigger reset' when new device is connected.

jpyllman:
I downloaded the latest build from the branch. There still seem to be something working. After disconnection the computer and the keyboard turns off backlight I can press the keyboard and get the backlight back on. And it turns off after the timeout. But doing this a couple of times it seems to get stuck in backlight being on.

kareltucek:
> But doing this a couple of times it seems to get stuck in backlight being on.

Can you provide a bit more details? E.g., how many times is "couple of times"?, what is the approximate uptime of the keyboard when this happens (seconds, minutes or hours)? Have you tried waiting another 10 or so seconds? Provide a testing UserConfig...

Just tested 20 or so times without any issue.

jpyllman:
I am not sure how to describe it. Now I have put the build after your last updates. And now we are back at, disconnecting the computer from the dock will turn off background light direct and I can not trigger it on by pressing any key. So, generally that is better than it was when it stayed on all the time.

kareltucek:
That's strange. 

Anyways does it turn back on when you connect it again?

kareltucek:
Let's merge this, as it clearly solves at least two causes for the problem.

PR #1367, opened 2025-09-25
Fix secondary role unblock mouse bug.

kareltucek:


right/src/secondary_role_driver.c
```diff
@@ -308,6 +308,10 @@ secondary_role_result_t SecondaryRoles_ResolveState(key_state_t* keyState, secon
         resolutionCallerIsMacroEngine = isMacroResolution;
         resolutionState = startResolution(keyState, strategy);
         resolutionState = resolveCurrentKey(strategy);
+        bool resolvedNow = resolutionState != SecondaryRoleState_DontKnowYet;
+        if (resolvedNow) {
+            PostponerExtended_UnblockMouse();
+        }
```
firngrod, line 314: Pretty much the same as I came up with.

firngrod (approved):
And the rest can't hurt either.  Used it for a day now without issue.

PR #1375, opened 2025-10-21
Implement dynamic pinctl configuration.

kareltucek:


kareltucek:
Tested using macros with args 0 through 3. (3 = current state of affairs)
```
zephyr settings write uhk/uartDebug/mode 3
reboot
```

PR #1393, opened 2025-11-28
Firngrod/primary from same half

firngrod:
I did try my best to keep to the repository guidelines.  Let me know if anything is out of place.

I elected not to make it exclusive to advanced strategy as it could also be useful for simple strategy.


right/src/secondary_role_driver.h
```diff
@@ -72,11 +72,17 @@
...
+        SecondaryRole_DefaultFromSameHalf,
+        SecondaryRole_PrimaryFromSameHalf,
+        SecondaryRole_SecondaryFromSameHalf,
+    } secondary_role_same_half_t;
+
 // Variables:
 
 // Functions:
 
-    secondary_role_result_t SecondaryRoles_ResolveState(key_state_t* keyState, secondary_role_t rolePreview, secondary_role_strategy_t strategy, bool isNewResolution, bool isMacroResolution);
```
firngrod, line 79: Removed an unused parameter.

right/src/utils.c
```diff
@@ -176,3 +176,13 @@ bool ShouldResendReport(bool statusOk, uint8_t* counter) {
...
+
+uint8_t IsRightHalfOrModule(uint16_t id)
+{
+    return id < 64 || id >= 192;
+}
+
+bool Utils_AreKeysOnTheSameHalf(uint16_t oneKey, uint16_t anotherKey)
+{
+    return IsRightHalfOrModule(oneKey) == IsRightHalfOrModule(anotherKey);
+}
```
firngrod, line 188: Might not have to be util function.  There are two schools of thought on this I think:  1)  It should not be moved to utils until it's needed elsewhere and 2) If it's not, it will get implemented multiple places.

kareltucek: Both variants are fine in case of `Utils_AreKeysOnTheSameHalf`. 

`IsRightHalfOrModule` should definitely remain in utils.c so that those magic constants remain centralized.

right/src/utils.c
```diff
@@ -176,3 +176,13 @@ bool ShouldResendReport(bool statusOk, uint8_t* counter) {
         return false;
     }
 }
+
+uint8_t IsRightHalfOrModule(uint16_t id)
```
kareltucek:
```suggestion
static uint8_t isRightHalfOrModule(uint16_t id)
```

If a function is not used from another compilation unit, we prefer it to be static and dromedaryCase. (Allows compiler to inline it, and makes clear what is actually used project-wide and what not.)

doc-dev/reference-manual.md
kareltucek:
```suggestion
CONDITION = {ifPrimary | ifSecondary} [ simpleStrategy | advancedStrategy | primaryFromSameHalf | secondaryFromSameHalf ]*
```

Let's allow any order?

(Also, the grammar edits are important. That's where Agent autocomplete comes from, and generally I see it as the key point for navigation.)

doc-dev/reference-manual.md
```diff
@@ -473,7 +474,7 @@ Commands:
 - `holdLayer LAYERID` mostly corresponds to the sequence `toggleLayer <layer>; delayUntilRelease; untoggleLayer`, except for more elaborate conflict resolution (releasing holds in incorrect order).
 - `holdKeymapLayer KEYMAPID LAYERID` just as holdLayer, but allows referring to layer of different keymap. This reloads the entire keymap, so it may be very inefficient.
 - `holdLayerMax/holdKeymapLayerMax` will timeout after <timeout> ms if no action is performed in that time.
-- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
+- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] [ primaryFromSameHalf | secondaryFromSameHalf ]... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```
kareltucek:
```suggestion
- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy | primaryFromSameHalf | secondaryFromSameHalf ]*... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```

doc-dev/reference-manual.md
```diff
@@ -497,7 +498,7 @@ We allow postponing key activations in order to allow deciding between some scen
 - `ifKeyPendingAt/ifNotKeyPendingAt <idx> <keyId>` looks into postponing queue at `idx`th waiting key and compares it to the `keyId`.
 - `consumePending <n>` will remove n records from the queue.
 - `activateKeyPostponed KEYID` will add tap of KEYID at the end of queue. If `atLayer LAYERID` is specified, action will be taken from that layer rather than current one. If `prepend` option is specified, event will be place at the beginning of the queue.
-- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
+- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] [ primaryFromSameHalf | secondaryFromSameHalf ] ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```
kareltucek:
```suggestion
- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy | primaryFromSameHalf | secondaryFromSameHalf ]* ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```

doc-dev/reference-manual.md
```diff
@@ -537,7 +538,7 @@ Conditions are checked before processing the rest of the command. If the conditi
 - `ifRecording/ifNotRecording` and `ifRecordingId/ifNotRecordingId MACROID` test if the runtime macro recorder is in the recording state.
 - `ifShortcut/ifNotShortcut [IFSHORTCUT_OPTIONS]* [KEYID]*` will wait for future keypresses and compare them to the argument. See the postponer mechanism section.
 - `ifGesture/ifNotGesture [IFSHORTCUT_OPTIONS]* [KEYID]*` just as `ifShortcut`, but breaks after 1000ms instead of when the key is released. See the postponer mechanism section.
-- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
+- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy ] [ primaryFromSameHalf | secondaryFromSameHalf ] ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```
kareltucek:
```suggestion
- `ifPrimary/ifSecondary [ simpleStrategy | advancedStrategy | primaryFromSameHalf | secondaryFromSameHalf ]* ... COMMAND` will wait until the firmware can distinguish whether primary or secondary action should be activated and then either execute `COMMAND` or skip it.
```

doc-dev/reference-manual.md
```diff
@@ -145,11 +145,12 @@ COMMAND = set module.touchpad.pinchZoomDivisor <1-100 (FLOAT)>
 COMMAND = set module.touchpad.pinchZoomMode NAVIGATION_MODE
 COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
+COMMAND = set secondaryRole.primaryFromSameHalf <prevents secondary role keys from triggering as secondary when the triggering key is from the same half of the keyboard (BOOL)>
```
kareltucek: > I elected not to make it exclusive to advanced strategy as it could also be useful for simple strategy.

I think the global configurations should remain separate for simple and advanced strategy, so that there are two presets that the user can choose between depending on a context - typically modifier/layer keys vs alphanumeric keys. 

So unless there is a deeper reasoning behind it that I am missing, I would prefer to make it an option of the advanced strategy only. (Its triggerByPress option makes it possible to configure the advanced strategy to act as the simple one, so it isn't limiting the configurability almost at all.)

------

`secondaryFromSameHalf` makes me think that it makes same half resolution *always* produce secondary role, but that is not the case, is it? 

Would it make sense to leave only the  `primaryFromSameHalf`? If not, then maybe `primaryFromSameHalf` and `eitherRoleFromSameHalf` would work?

right/src/secondary_role_driver.c
```diff
@@ -132,7 +134,10 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
     postponer_buffer_record_type_t *actionRelease;
 
     PostponerQuery_InfoByKeystate(resolutionKey, &dummy, &dualRoleRelease);
-    PostponerQuery_InfoByQueueIdx(0, &actionPress, &actionRelease);
+    uint8_t lastKeyQueuePos = PostponerQuery_PendingKeypressCount() - 1;
+    PostponerQuery_InfoByQueueIdx(lastKeyQueuePos, &actionPress, &actionRelease);
+    uint16_t actionKeyId = PostponerExtended_PendingId(lastKeyQueuePos);
+    uint16_t dualRoleId = Utils_KeyStateToKeyId(resolutionKey);
```
kareltucek: _last_ key in the queue, or the _action_ key keyid? You may get scenarios where there are more keys after the action key, in which case this is a major change to the logic of the entire driver, isn't it?

If you are just trying to get the action key keyid, then I would expect this to do the trick:

Edit: remove the suggestion.

right/src/secondary_role_driver.c
```diff
@@ -58,6 +59,7 @@
  */
 
 static bool resolutionCallerIsMacroEngine = false;
+static bool isPrimaryFromSameHalf = false;
```
kareltucek: This name (the leading `is` part) is hard to digest for me. Wouldn't `primaryFromSameHalf`, or `resolutionPrimaryFromSameHalfEnabled` work better?

right/src/macros/commands.c
```diff
@@ -1002,6 +1003,13 @@ static macro_result_t processIfSecondaryCommand(parser_context_t* ctx, bool nega
         strategy = SecondaryRoleStrategy_Advanced;
     }
 
+    if (ConsumeToken(ctx, "primaryFromSameHalf")) {
+        fromSameHalf = SecondaryRole_PrimaryFromSameHalf;
+    }
+    else if (ConsumeToken(ctx, "secondaryFromSameHalf")) {
+        fromSameHalf = SecondaryRole_SecondaryFromSameHalf;
+    }
```
kareltucek:
```suggestion
while(true) {
    if (ConsumeToken(ctx, "simpleStrategy")) {
        strategy = SecondaryRoleStrategy_Simple;
    }
    else if (ConsumeToken(ctx, "advancedStrategy")) {
        strategy = SecondaryRoleStrategy_Advanced;
    }
    else if (ConsumeToken(ctx, "primaryFromSameHalf")) {
        fromSameHalf = SecondaryRole_PrimaryFromSameHalf;
    }
    else if (ConsumeToken(ctx, "secondaryFromSameHalf")) {
        fromSameHalf = SecondaryRole_SecondaryFromSameHalf;
    }
    else {
        break;
    }
}
```

Let's allow any order?

right/src/macros/set_command.c
```diff
@@ -324,6 +324,9 @@ static macro_variable_t secondaryRoles(parser_context_t* ctx, set_command_action
         DEFINE_NONE_LIMITS();
         ASSIGN_CUSTOM(int32_t, intVar, Cfg.SecondaryRoles_Strategy, ConsumeSecondaryRoleStrategy(ctx));
     }
+    if (ConsumeToken(ctx, "primaryFromSameHalf")) {
```
kareltucek:
```suggestion
    else if (ConsumeToken(ctx, "primaryFromSameHalf")) {
```

right/src/secondary_role_driver.c
```diff
@@ -173,6 +178,12 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
 
     //handle trigger by press
     if (Cfg.SecondaryRoles_AdvancedStrategyTriggerByPress) {
+        bool actionKeyWasPressedAndFromTheSameHalf = isPrimaryFromSameHalf && actionPress != NULL && Utils_AreKeysOnTheSameHalf(actionKeyId, dualRoleId);
+
+        if(actionKeyWasPressedAndFromTheSameHalf) {
```
kareltucek:
```suggestion
        if (actionKeyWasPressedAndFromTheSameHalf) {
```

An opinion-dependent nitpick, but laszlo is very keen on these.

kareltucek (changes requested):
Nice job overall, just a couple of questions and suggestions :-).

right/src/utils.c
firngrod: I remember reading that, actually.  Woopsie

doc-dev/reference-manual.md
firngrod: I was wondering where exactly the autocomplete came from, and I was wondering why it didn't pick these up.  Well, now I know.  :)
Also, I thought mutually exclusive parameters would be grouped, but yeah, nothing will cause errors, the last ones will take effect.

doc-dev/reference-manual.md
firngrod: The timeout makes it not possible to entirely match simple strategy, and that is actually something I have given a bit of thought.  I'm not too happy about the timeout functionality, but I have yet to come up with a better solution.
I will remove it from Simple and move the config under Advanced, however.  I can see your point.

---

Both keywords are needed, depending on what your global setup is (whether primary from same half is enabled or not).  I have primary enabled in configuration, use standard secondary role config on my 10 HRM keys on each layer, and then a macro with `secondaryFromSameHalf` for my two thumb mods.  Alternatively, one could have one parameter to inverse the behavior from the global config, but I strongly dislike that idea.
Naming is always hard.  I feel that the primary/secondary split is proper.  The concept it that if a secondary-role key is resolved by an action from another key (press or release), either secondary or primary action is taken, depending on whether that other key was on the same half or not.  The rest of the logic is identical (timeout triggers secondary, which then allows secondary effect to apply to keys from the same half)
`primary/secondaryActionFromSameHalfTrigger` might be a better description, but it's awfully long.

right/src/secondary_role_driver.c
firngrod: `PostponerQuery_InfoByQueueIdx(0, &actionPress, &actionRelease);` gives me the first key pressed after the secondary-role key under evaluation, not necessarily the one which triggered the evaluation.  For trigger on press, that's fine, but it breaks with Trigger on release especially when we care which key triggered the resolution.
If I press `Ctrl/O+Alt/S` on my right half and `Delete` on my left, `Ctrl/O`, `PostponerQuery_InfoByQueueIdx(0, &actionPress, &actionRelease);` gives me the `Alt/S` key every time the `Ctrl/O` evaluates, never the `Delete`, and it doesn't work.
Maybe I should pass the action key down all the way from `SecondaryRoles_ResolveState`?  I'm going to try that.

right/src/secondary_role_driver.c
firngrod: Sure, I'll go with `primaryFromSameHalf`

right/src/secondary_role_driver.c
firngrod: Nope, that was the resolution key.

kareltucek: Please elaborate on the above scenario. 

What is the exact scenario, what do you expect to happen and why? 

So far the way I see it, when you press O+S, with primaryFromTheSameHalf enabled, it should resolve as primary action for O (by definition of primaryFromTheSameHalf), and perform the O primary action. Next, you get S press applied, which triggers a new dual role resolution on S+Delete.

What am I getting wrong?

doc-dev/reference-manual.md
kareltucek: > Naming is always hard. I feel that the primary/secondary split is proper.

Indeed. 

I mean, linguistically, secondary role makes sense as an opposite of a primary role, but semantically this no longer holds, as there are three possibly sensible behaviors: 

1) same half always triggers primary role.
2) same half always triggers secondary role.
3) same half has no effect. 

This PR implements behaviors 1 and 3, which is fine, and I am not suggesting implementing the 2 behavior as well.

But the naming strongly suggests that the options force behaviors 1 and 2 rather than 1 and 3.

right/src/secondary_role_driver.c
firngrod: It's trigger on release and O and S are being held.  I didn't make that clear.  Neither O or S should trigger before timeout or a key has been released.  And since Delete is released first, O should trigger secondary because it was from the other half.  That is why I need information on the last postponed key, the Delete key.  Realistically, I should loop over the queue to find the first released key.  I think I'm going to try that to see if it makes any difference.

doc-dev/reference-manual.md
firngrod: Is there a scenario where an alternate role key is resolved by an action from another key (press or release) and the resulting action is not either primary or secondary?

right/src/secondary_role_driver.c
kareltucek: (The queue lookups are sorted by press times, right?)

So the scenario is: 

- press O
- press S
- press Delete
- release Delete
- we are now examining O secondary resolution caused by release of the delete key.
- your intention is to activate a secondary O+Delete role

In what real life scenario are you triggering a secondary role with another intermediate tap in between of the secondary key and the action key?

I mean, I would expect that the S tap is an evidence that you are in the middle of writing and so that the O+whatever would be an undesired O secondary role activation.

-------------------

>  Realistically, I should loop over the queue to find the first released key.

Well, you have no guarantee that secondary evaluation takes place on every wake event, so you will indeed have to loop over the queue. 

But let's not get ahead of ourselves. I still don't understand the scenario.

doc-dev/reference-manual.md
kareltucek: From a point of view of a finished resolution? No.

right/src/secondary_role_driver.c
firngrod: My conceptual idea of trigger by release is that when an alt role key is being held, the release of ANY key pressed after it should trigger resolution, not just the next key.  This allows the very important possibility of combining alt role keys in a multi-mod combo, like the above mentioned `Ctrl+Alt+Del`. Press O, press N, press Delete, release Delete, release O and N any order.
On release of Del, I want O to trigger Secondary for Ctrl because of that Delete release trigger, then N to trigger secondary to Alt because of that Delete trigger, then Delete with those mods, then the mods release.

What happens in the current main branch logic is that once I release the combos, I either release O or N first.  If I release N first, O triggers Ctrl because of N release, then N triggers Alt because of the previous Delete release, then Delete happens and mods are released in the proper order.  If I release O first, first O triggers primary because N is still held, then N triggers secondary because Delete was previously released, and then things unroll from there.  This is largely mitigated by the Trigger safety margin, and generally it works as I want it to anyway, and it's all acceptable.

However, that logic will never work properly once we start caring which key caused the resolution to happen, in this case whether N or Delete caused the O resolution by releasing first.  Without looking through the queue for the first released button, multi-mod combos can only be made by utilizing the timeout to trigger the first mod as secondary.

In general, taking the last key seems to work fine, but I do sometimes see what I feel are false negatives, although I cannot be sure.  It may be that I press a secondary Shift, then press the one I want to Capitalize, then press the next key immediately before I release the to-be-capitalized key.  I want to see if that is the case.

kareltucek: > Ctrl+Alt+Del

Ah, right, I didn't realize this was a multi-mod scenario.

I see now the reasoning and fully agree.

> It may be that I press a secondary Shift, then press the one I want to Capitalize, then press the next key immediately before I release the to-be-capitalized key. I want to see if that is the case.

Correct. That's why you can't blindly take the last key. 

With a proper implementation things still should work if you press a `postponeKeys delayUntilRelease` macro, then tap your secondary scenario, then tap another key or two, then release the postponeKeys macro.

So to sum up, my conclusion is that:

- For release triggers, you need to take the first-by-time released key in the queue, that also has the corresponding press in the queue. (This should be implemented in the postponer as a new function that does all the work of finding such a key.) 

- For the press trigger, you still need to take the first-by-order pressed key in the queue, as there may be no released key in the queue at all. (I.e., according to the old algorithm.)

(Ofc., let me know if there is anything suspicious in my reasoning whatsoever.)

doc-dev/reference-manual.md
kareltucek: Still, I think the point stands, as this is not a point of view of a finished resolution, but a point of view of one trigger condition, that (as far as a random reader knows) can pick out of three possible restrictions of the future (primary / secondary / let someone else decide).

kareltucek: Ideas:
- `primaryFromSameHalf` and `ignoreSameHalf`
- `primaryFromSameHalf` and `dontKnowFromSameHalf`
- `primaryFromSameHalf` and `disablePrimaryFromSameHalf`

right/src/secondary_role_driver.c
firngrod: Sounds good.  I will get on that when I have the time.  Hopefully tonight or tomorrow night.

doc-dev/reference-manual.md
firngrod: I still don't get how it can be `ignoreSameHalf`.  This PR changes nothing about _when_ an alt role key is resolved.  When a trigger key is pressed/released, depending on trigger choice, the alt role key pending resolution is always resolved.  There is no option to not resolve when a key the trigger comes.   If the trigger key is from the other half, the alt role key always triggers secondary.  If the trigger key is from the same half, this feature allows the user to configure whether the alt role key should resolve to primary or secondary.  Never to not resolve.

kareltucek: Sorry, I am still missing something.

I am now thinking in the context of triggering by release and these two scenarios:

- scenario 1:
	- press a/shift
	- press s
	- release a/shift
	- release s

- scenario 2:
	- press a/shift
	- press s
	- release s
	- release a/shift

Normal behavior is that 1 produces `as` while 2 produces `S`.

--------------

> If the trigger key is from the other half, the alt role key always triggers secondary.

Consider:

- `primaryFromSameHalf` was provided.
- press a/shift
- press l
- release a/shift
- release l

I would expect this to yield primary, despite l being from the other half, which contradicts the above statement.

> If the trigger key is from the same half, this feature allows the user to configure whether the alt role key should resolve to primary or secondary.

Consider:

- `secondaryFromSameHalf` was provided
- press a/shift
- press s
- release a/shift
- release s

Now, I believe this code yields primary, which contradicts the quoted statement as well as my human understanding of the name `secondaryFromSameHalf`.

-------

Where am I making a mistake?

mhantsch: > primary/secondaryActionFromSameHalfTrigger

I like what you are building, but I think the naming needs to be improved. From just reading the names, I don't really understand what this is supposed to achieve.

Is it `allowSecondaryResolutionFromSameHalf` and `allowPrimaryResoludionFromSameHalf` or is it `sameHalfTriggerResolvesToPrimary` and `sameHalfTriggerResolvesToSecondary`?

Is there a way to allow certain keys (e.g. thumb keys) to be excluded from those special conditions, and just go through normal (advanced) primary/secondary resolution?

firngrod: @mhantsch: This PR makes two major changes in logic.  One is going to be a non-configurable change to the trigger-by-release logic as discussed in the other long thread in this PR:  Right now, trigger-by-release triggers by release of the first key pressed after the Alt role key was pressed.  This PR will make it so that trigger-by-release triggers on the first release of any key which was pressed after the Alt role key was pressed.  Timeout, safety margin and so on will otherwise work as before.
The other change is that it is will be configurable through global config, and on individual key basis through macros (`ifPrimary <someKeywordHere>`) whether an Alt role key will resolve as primary or secondary if it is triggered by a press or release of another key and that key was on the same half of the keyboard as the Alt role key.
The debate is what `someKeywordHere` should be, specifically when triggering as secondary role from trigger keys on the same keyboard half is desired as an override in a macro when global config is set to trigger primary from same-half triggers.

firngrod: @kareltucek In the pressA-pressB-releaseA-releaseB with trigger-on-release case, the trigger is from the release of the alt role key without any other key about which we can even consider half-association.  That scenario is not affected by the configuration we seek to name, and I have not considered it in my naming suggestions.

firngrod: I guess a fully descriptive name if one knows the source code variable names would be `resolve[Primary/Secondary]OnTriggerFromActionKeyOnSameKeyboardHalf`
`sameHalfTriggerResolvesToPrimary` and `sameHalfTriggerResolvesToSecondary` might be good, if kind of long.

kareltucek: > @kareltucek In the pressA-pressB-releaseA-releaseB with trigger-on-release case, the trigger is from the release of the alt role key without any other key about which we can even consider half-association.

But this conditioning requires exact and detailed knowledge of the feature and code, *and* an expectation that you expect the user to have this knowledge and actively use it when interpreting the name. That's why I don't understand them - I (think I) have the former, but I don't have the latter.

Without that knowledge, or without that expectation, I intuitively interpret the S key of a pressA-pressS-releaseA-releaseS scenario to be subject to any "same half trigger" conditioning. 

> I guess a fully descriptive name if one knows the source code variable names would be `resolve[Primary/Secondary]OnTriggerFromActionKeyOnSameKeyboardHalf`
> `sameHalfTriggerResolvesToPrimary` and `sameHalfTriggerResolvesToSecondary` might be good, if kind of long.

 Actually, still no. But I see now from where you are coming. A fixed descriptive name would have to be `sameHalfSecondaryTriggerResolvesToPrimary` and `sameHalfSecondaryTriggerResolvesToSecondary`

> I still don't get how it can be `ignoreSameHalf`.

The idea was "Ignore any logic of the same-half feature during this resolution."

----------------- 

Following variants work for me, I think:
- `sameHalfSecondaryTriggerResolvesToPrimary` and `sameHalfSecondaryTriggerResolvesToSecondary`
- `sameHalfSecondaryResolvesToPrimary` and `sameHalfSecondaryResolvesToSecondary`
- `primaryFromSameHalfEnabled` and `primaryFromSameHalfDisabled`
- `allowSecondaryFromSameHalf` and `forcePrimaryFromSameHalf`

firngrod: > The idea was "Ignore any logic of the same-half feature during this resolution."

Well that falls to similar logic in that I read that as "Ignore triggers from the same half".  :)
`primaryFromSameHalfEnabled` and `primaryFromSameHalfDisabled` work for me although I do not like the difference in two similar strings to be on the tail.  I find that harder to read.
`primaryFromSameHalf` and `noPrimaryFromSameHalf` maybe, and then ensure that the documentation is very clear on what it means?
Maybe `doPrimaryFromSameHalf` and `noPrimaryFromSameHalf` although the semantics are different in the two.

kareltucek: > Maybe `doPrimaryFromSameHalf` and `noPrimaryFromSameHalf` although the semantics are different in the two.

They are visually very similar. Also the "do" and "no" have much more room for interpretation than "enable"/"disable". 

---------

> `primaryFromSameHalfEnabled` and `primaryFromSameHalfDisabled` work for me although I do not like the difference in two similar strings to be on the tail. I find that harder to read.

`enablePrimaryFromSameHalf` and `disablePrimaryFromSameHalf`, or  `primaryFromSameHalf` and `primaryFromSameHalfDisabled` work for me too in case it improves anything.

mhantsch:
Just to give you an update: I am still very interested in trying out these changes. I am a heavy user of homerow-mods and am very curious how they work out for me. Maybe I can then also provide some newer suggestions on the naming of the keywords.

Unfortunately, my UHK80 was returned to UGL as part of a recall, so I have to wait to receive a new keyboard again. Due to holidays season, it will most likely be until at least middle of January before I can try out any newer firmware.

firngrod:
Sorry about leaving this hanging for so long.  Life has been keeping me very busy this last week.  I will get around to it soon, I promise(tm)

right/src/secondary_role_driver.c
firngrod: Yeah, that made a difference.

right/src/postponer.c
```diff
@@ -415,6 +415,26 @@ void PostponerQuery_InfoByQueueIdx(uint8_t idx, postponer_buffer_record_type_t**
     }
 }
 
+void PostponerQuery_FindFirstReleased(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release)
```
kareltucek, line 418: Very sorry for picking this apart to such detail, I am just terrified by the prospect of having to search for bugs in this later 🙈.

Shouldn't you first find the key release and just then find the corresponding press? 

I.e., search for first-by-time-release with existing press rather than first-by-time-press with existing release?

I mean:

- press ctrl
- press alt
- press delete
- release delete
- release ctrl
- release alt
- now evaluate the ctrl secondary role

In this scenario, this will give you the alt key, whereas I think you are after the delete key?

doc-dev/reference-manual.md
firngrod: If you're happy with the naming now, feel free to mark this one as resolved.

kareltucek: Sure, perfect!

right/src/postponer.c
firngrod: Not a problem at all!  It's good to keep the firmware as solid as possible.
It doesn't matter in the general case since there will only ever be one matching set of press and release.  Once there is, resolution happens and the superposition collapses.
In your example, Ctrl and Alt are resolved once Delete is released, and the release of Ctrl and Alt happens off the queue.
However, with the new minimum hold time, you are correct, there is a possibility of this causing issues.  I will redo the logic.

firngrod: Is there a possibility of two releases being registered in one cycle of the secondary driver evaluation?

kareltucek: Yes, it can happen that in one secondary resolution pass there is no release in the queue, and in the next pass there are multiple.

For instance some other postponing logic (say ifGesture) may postpone multiple dual role keys past their releases - just as in the postponeKeys example above.

The macro engine parsing and execution is slow. With multiple complex running macros it may happen that the secondary resolution macro doesn't get to run for dozens milliseconds at a time, which is sufficient for multiple key events to arrive in between two evaluations.

kareltucek (approved):
Looks good to me!

PR #1394, opened 2025-11-28
Finrgrod/Secondary role minimum hold time

firngrod:
What the title says, essentially.

I haven't explicitly tested this on it's own, only on top of the same half stuff, but it should be okay.


doc-dev/reference-manual.md
```diff
@@ -151,6 +151,7 @@ COMMAND = set secondaryRole.advanced.safetyMargin <ms, higher value adjusts sens
 COMMAND = set secondaryRole.advanced.triggerByPress <trigger immediately on action key press (BOOL)>
 COMMAND = set secondaryRole.advanced.triggerByRelease <trigger secondary role if action key is released before dual role (BOOL)
 COMMAND = set secondaryRole.advanced.triggerByMouse <trigger secondary role immediately on mouse move (BOOL)
+COMMAND = set secondaryRole.advanced.minimumHoldTime <ms, minimum time a key must be held before it can trigger secondary role 0-255>
```
kareltucek:
```suggestion
COMMAND = set secondaryRole.advanced.minimumHoldTime <ms, minimum time a key must be held before it can trigger secondary role 0-255 (INT)>
```

right/src/secondary_role_driver.c
```diff
@@ -196,6 +206,10 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
         bool actionKeyWasReleasedFirst = actionRelease != NULL && dualRoleRelease != NULL && actionRelease->time <= dualRoleRelease->time - Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
 
         if (actionKeyWasReleasedFirst || actionKeyWasReleasedButDualkeyNot) {
+            if(!dualRoleWasHeldLongEnoughToBeAllowedSecondary){
+                sleepTimeoutStrategy(Cfg.SecondaryRoles_AdvancedStrategyMinimumHoldTime);
+                return SecondaryRoleState_DontKnowYet;
+            }
             KEY_TIMING2(actionKeyWasReleasedButDualkeyNot, KeyTiming_RecordComment(resolutionKey, "SG"));
             KEY_TIMING2(actionKeyWasReleasedFirst, KeyTiming_RecordComment(resolutionKey, "SH"));
```
kareltucek, line 214: I should have told you about these earlier. 

This code along with a macro switch and a shell script allows you to capture a timing log and draw an svg graph out of it.

Haven't used it in ages and give no guarantees about the state it is in (it is a very improvised thing that can bring more trouble than benefit)... but just so that you know what it is for and that it exists.

kareltucek (approved):
Looks good to me, just please commit that one grammar suggestion and confirm whether I can merge it. (I will do so against a staging branch. I am more concerned about not creating conflicts against your workflow.)

right/src/secondary_role_driver.c
firngrod: I figured it was something like that.  By the way, I haven't kept those up to date in the other PR.  Maybe I should update the macros and add the missing ones.

firngrod:
If it creates conflicts with my other PRs, I will fix that.

right/src/secondary_role_driver.c
kareltucek: It is fine. The next person who needs it will have to revise the code anyways, so feel free to ignore them 🤷.

PR #1411, opened 2025-12-17
Doubletap primary on timeout

firngrod:
Moved the primary activation on double tap to primary to only activate on timeout.  This is important for HRMs or a Shift secondary on Space.


kareltucek:
Ok!

PR #1415, opened 2025-12-30
Heatmap.

kareltucek:
I had written this as a pre-christmas procrastination. 

It records key press statistics and then it colors keys according to how much they are pressed compared to others - black to blue to purplish to orange to red. 

It is more of a fun thing than a useful tool. (It lacks a lot of refinement and configurability for being actually useful.)

Any opinions on whether to merge it or throw it away?


mondalaci:
My gut reaction is that anything that doesn't add substantial value for users is a maintenance burden that shouldn't be merged, probably including a feature, but feel free to ask on the forum.

pcooke9:
Sounds kinda cool. I'd like to play around with it while gaming for a bit first, if you're not gonna keep it. But like Laszlo said, I imagine that if you merged it, a few people are likely to bug you for tweakability options😉.

If there's ever any weird/disposable stuff you want someone to check out, I'm in. I've got nothing else going on.😎

Is it usable as is now? If so, is it just a matter of using `set backlight.strategy heatmap`?

kareltucek:
> Is it usable as is now? If so, is it just a matter of using set backlight.strategy heatmap?

Yes.

ert78gb:
I dont have strong opinion about this feature. I just share my thought.
I could imagine these kind if exotic features as plugins. These are not part the "main" firmware, the user can upload and delete them from the keyboard And turn on or off .

The plugin system provides adds complexity to the core module but maybe worth to think about it.

mondalaci:
To minimize complexity, I'd rather close this PR. We can think about a plugin solution in the far future.

mhantsch:
I love the idea, and I would like to be able to export the heatmap from the keyboard to the host. A macro command to send heatmap numbers to the host (by typing them in some kind of table)?

kareltucek:
It shouldn't be hard to write a macro that prints them into the status buffer. 

However, as this branch is abandoned, I don't plan to invest any more time into it ;-).

PR #1418, opened 2026-01-02
Firngrod/timeout noop option

firngrod:
The biggest change in this PR is how the caching of resolutions of secondary key state is handled.  Now the fact that a key needs a new resolution is explicitly set for new presses both in the case of secondary role keys and macros.  The big difference here is that a macro will only allow this resolution to run once for it's triggering key, and any future `ifPrimary/ifSecondary` will reference the outcome of that first resolution.

Then I've also added a "no action" resolution option, currently available only on timeout.  It's not that I'm actually going to use that before I have added more logic, but it's a necessity for what I want to make next.

As mentioned in the comments, I am not really happy with how the cache is managed.  There is a mixture of responsibility of values and variables which I don't like.
The best solution in my opinion would be to add another bit to the key state struct for resolution cache tracking, either by an explicit value, or more flexibly by expanding the state enum.  Expanding the state enum would make it 5 values, thus requiring a third bit, but then additional values could be added later.  The reason I have not done so already is that it would take another bit of memory for each key, in addition to what I've already added.

Otherwise, I have tested the changes to the logic by running it for a few days with no issues.

### DIFF
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -146,7 +146,7 @@ COMMAND = set module.touchpad.pinchZoomMode NAVIGATION_MODE
 COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
 COMMAND = set secondaryRole.advanced.timeout <ms, 0-500 (INT)>
-COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary }
+COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | noAction }
 COMMAND = set secondaryRole.advanced.safetyMargin <ms, higher value adjusts sensitivity towards primary role -50-50 (INT)>
 COMMAND = set secondaryRole.advanced.triggerByPress <trigger immediately on action key press (BOOL)>
 COMMAND = set secondaryRole.advanced.triggerByRelease <trigger secondary role if action key is released before dual role (BOOL)>
@@ -688,7 +688,7 @@ Internally, values are saved in one of the following types, and types are automa
     - simple strategy listens for other key activations until the dual-role key is released. If there is any such activation, it activates the secondary role and then the action of the other key without any further delays. If there is no such other action, it performs primary role on the dual-role key release.
     - advanced strategy may trigger secondary role depending on timeout, or depending on key release order.
       - `set secondaryRole.advanced.timeout <timeout in ms, 350 (INT)>` if this timeout is reached, `timeoutAction` (secondary by default) role is activated.
-      - `set secondaryRole.advanced.timeoutAction { primary | secondary }` defines whether the primary action or the secondary role should be activated when timeout is reached
+      - `set secondaryRole.advanced.timeoutAction { primary | secondary | noAction}` defines whether the primary action or the secondary role, or no action at all, should be activated when timeout is reached
       - `set secondaryRole.advanced.triggerByRelease BOOL` if enabled, secondary role is chosen depending on the release order of the keys (`press-A, press-B, release-B, release-A` leads to secondary action; `press-A, press-B, release-A, release-B` leads to primary action). This is further modified by safetyMargin.
       - `set secondaryRole.advanced.triggerByPress BOOL` if enabled, secondary role is triggered when there is another press, simiarly to the simple strategy. Unlike simple strategy, this allows setting timeout behaviors, and also is modified by safetyMargin.
       - `set secondaryRole.advanced.triggerByMouse BOOL` if enabled, any mouse (module) activity triggers secondary role immediately.
--- a/doc-dev/reference-manual.md
+++ b/doc-dev/reference-manual.md
@@ -146,7 +146,7 @@ COMMAND = set module.touchpad.pinchZoomMode NAVIGATION_MODE
 COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
 COMMAND = set secondaryRole.advanced.timeout <ms, 0-500 (INT)>
-COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | noAction }
+COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | none }
 COMMAND = set secondaryRole.advanced.safetyMargin <ms, higher value adjusts sensitivity towards primary role -50-50 (INT)>
 COMMAND = set secondaryRole.advanced.triggerByPress <trigger immediately on action key press (BOOL)>
 COMMAND = set secondaryRole.advanced.triggerByRelease <trigger secondary role if action key is released before dual role (BOOL)>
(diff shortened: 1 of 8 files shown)


right/src/key_states.h
```diff
@@ -11,6 +11,13 @@
 
 // Typedefs:
 
+    typedef enum {
```
firngrod, line 14: I tried to use `typedef enum : uint8_t`, but while my local compiler handled it fine, the pipeline one complained and I think I saw a warning that enums are automatically made int32_t.

right/src/secondary_role_driver.c
```diff
@@ -68,39 +68,28 @@ static key_state_t *previousResolutionKey;
...
+I dislike the arrangement because it puts double roles on the values SecondaryRoleState_DontKnowYet 
+and SecondaryRoleState_NoOp and the variable key_state_t.secondaryState.
+In some cases they are states of resolution, in other cases, they are cache management.
+The best alternative I can see is to extend the secondary_role_state_t enum with a fifth state 
+SecondaryRoleState_ResolutionUninitialized, or adding another member to key_state_t, but then yet another bit of memory
+per key is required.  Personally, I like that better, but I leave that up to the maintainers.
+One could also add two new names to the enum with duplicate values, but that still makes the usage of
+key_state_t.secondaryState confusing with it's content's meaning being dependend on the stored resolutionKey
+*/
+void SecondaryRoles_ResetResolution(key_state_t* keyState) { keyState->secondaryState = SecondaryRoleState_DontKnowYet; }
```
kareltucek: I would prefer the resetting to be done by a direct variable write from usb_report_updater.c:645 (under the `                actionCache[slotId][keyId].modifierLayerMask = 0;` line).

doc-dev/reference-manual.md
```diff
@@ -146,7 +146,7 @@ COMMAND = set module.touchpad.pinchZoomMode NAVIGATION_MODE
 COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
 COMMAND = set secondaryRole.advanced.timeout <ms, 0-500 (INT)>
-COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary }
+COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | noAction }
```
kareltucek: Would it make sense to call it `none` instead? 

It would go better with the idioms of the macro language I think.

right/src/secondary_role_driver.c
```diff
@@ -68,39 +68,28 @@ static key_state_t *previousResolutionKey;
...
+I am not happy about this arrangement, and it is very much up for discussion.
+I dislike the arrangement because it puts double roles on the values SecondaryRoleState_DontKnowYet 
+and SecondaryRoleState_NoOp and the variable key_state_t.secondaryState.
+In some cases they are states of resolution, in other cases, they are cache management.
+The best alternative I can see is to extend the secondary_role_state_t enum with a fifth state 
+SecondaryRoleState_ResolutionUninitialized, or adding another member to key_state_t, but then yet another bit of memory
+per key is required.  Personally, I like that better, but I leave that up to the maintainers.
+One could also add two new names to the enum with duplicate values, but that still makes the usage of
+key_state_t.secondaryState confusing with it's content's meaning being dependend on the stored resolutionKey
+*/
```
kareltucek: I slightly prefer to leave it as is, not "exposing" the "need of secondary resolution initialization" concept to the rest of the codebase. 

The way I think of it, the keystate value simply states the information about the state of the key, without having implementation details attached to it. 

I think of the initialization as being an inside thing of the secondary role driver blackbox.

But at the end of the day, the reasoning makes sense, so if I have not convinced you that the 2 bit variant makes sense too, and you feel sufficiently strongly about it, then feel free to do it.

right/src/key_states.h
kareltucek: I am confused about this statement. 

Is it just a fyi for me, or do you expect some reply from me? 

In any case, noted :-).

kareltucek (approved):
Looks good to me. 

Some minor comments follow.

Please make any changes that you consider desirable, and then let me know to merge it. 👍

right/src/key_states.h
firngrod: We talked briefly about enum variable sizes, so I just wanted to mention that I tried declaring it as `int8_t` and it failed, also that I think I saw indication that it might currently be an `int32_t`, and finally that I think the compiler in the pipeline is different from the one we use locally.

right/src/secondary_role_driver.c
firngrod: > The way I think of it, the keystate value simply states the information about the state of the key, without having implementation details attached to it.
> I think of the initialization as being an inside thing of the secondary role driver blackbox.

Something needs to invalidate the secondary role cached for a key.  This could happen inside the secondary role driver if the secondary role driver was always run while KeyActivatedNow was true.  This is not the case for macros.  Therefore something needs to explicitly clear the cache from outside the driver.  I put it where I did because then we only use cycles on it in cases where it might be relevant.  If you say the performance to complexity trade-off is worth it to do it every time, I will absolutely accept that.

> if I have not convinced you that the 2 bit variant makes sense too

I don't know what you mean by "the 2 bit variant".  I extended `key_state_t.secondary` to 2 bits and now use the four possible states for `uresolved/primary/secondary/noOp`, but I need a fifth state to differentiate "resolution not started" and "resolution not finished", so in the ideal world I need 5 states, requiring 3 bits.  For this reason, some states pull double duty depending on the current content of `resolutionKey`.
I think I can rewrite the logic of `SecondaryRoles_ResolveState` to structure it so that I can avoid confusion without adding a fifth state.

firngrod: I can even remove the need for cache reset IF we either cache the result in macro or accept multiple runs of resolution within a macro run.

kareltucek: >  If you say the performance to complexity trade-off is worth it to do it every time, I will absolutely accept that. 

Doing it with a direct variable write (as suggested above) should be cheap, shouldn't it? 

I think that the code complexity and robustness trade off is worth it in this case.

> I don't know what you mean by "the 2 bit variant".

I mean the variant implemented in this PR. I might be doing a bit of rationalizing there, but I am not finding anything about it to be particularly convoluted, or prone to exploding when stroked gently.

> I can even remove the need for cache reset IF we either cache the result in macro or accept multiple runs of resolution within a macro run.

I am totally fine with the cache reset. 

I am just thinking about the dilemma between the (current) 4  state solution and the (proposed) 5 state solution.

I slightly prefer using current variant compared to allocating another bit. If there was any functional benefit to the 5 states, I would say go for it, but as long as this is just an aesthetics, I find current code to be fine.

right/src/secondary_role_driver.c
```diff
@@ -325,65 +313,51 @@ static secondary_role_state_t startResolution(key_state_t *keyState, secondary_r
...
+}
 
-    return SecondaryRoleState_DontKnowYet;
+static void finishResolution(secondary_role_state_t res)
+{
+    resolutionKey->secondaryState = res;
+    previousResolutionKey = resolutionKey;
+    previousResolutionTime = resolutionStartTime;
+    if(!resolutionCallerIsMacroEngine) {
+        fakeActivation();
```
firngrod, line 316: This one may be a little questionable, but currently, there are only the two cases for secondary key resolution, do we think we will get more?

firngrod:
I redid the logic of `SecondaryRoles_ResolveState` such that it is very clear which key if any is currently under resolution.  This removes the need for a fifth state without making any of the other states ambiguous.
It also does some other stuff, like hiding the fake activation, claws back a little bit of memory from a redundant variable, and reordering the enum such that `primary` is 1 and `secondary` is 2.

right/src/secondary_role_driver.c
kareltucek: I don't think we will get more.

Anyways, rule of three says this condition here is just fine.

kareltucek: Or do you mean moving the fake activation from usb report updater here. 

Yes, it is ugly, but I can live with it.

kareltucek:
Nice refactor! 

Can I merge it?

firngrod:
I think so. I've been running it all day with no issue.

PR #1420, opened 2026-01-04
Firngrod/improved same half logic

firngrod:
This improves upon the same half secondary logic in that it no longer actively triggers primary from same half activations, but rather just prevents same half keys from triggering secondary role.
While there are a number of edge case changes from this, none are what I would call bugs, and I think the benefits outweigh the costs.
Benefits:
- Centralized logic:  The decision is now made only in the isKeyAllowedToBeActionKey function, the rest of the code functions the same regardless of same half logic
- When using TriggerOnPress, it is now possible to use multi-mod combos.  I even managed to forget I had triggerOnPress on for a brief while after testing that it worked.
- I think the naming is much more clear on this behavior than on the previous one.  Let's start a discussion on it!

Arguably negative changes are mainly to do with the timing aspect.  One can type out entire key sequences on the same half while a secondary role key is being held with nothing happening before the key is released, which is sort of counter-intuitive, but only actually ever happens if one is determined on testing it.  Otherwise, it does ever so slightly the timing of application of some keys, but it's nothing next to just the fact of using Secondary Roles.


right/src/postponer.c
```diff
@@ -398,35 +398,45 @@ void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_typ
     }
 }
 
-void PostponerQuery_InfoByQueueIdx(uint8_t idx, postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release)
```
firngrod, line 401: I agonized a little on whether to leave this function in place and add a separate for the predicated search, but in the end I decided to avoid the dead code.

right/src/postponer.h
```diff
@@ -87,9 +87,10 @@
     bool PostponerQuery_IsKeyReleased(key_state_t* key);
     bool PostponerQuery_IsActiveEventually(key_state_t* key);
     void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release);
-    void PostponerQuery_InfoByQueueIdx(uint8_t idx, postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release);
-    void PostponerQuery_FindFirstReleased(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release);
     bool PostponerQuery_ContainsKeyId(uint8_t keyid);
+    typedef bool (*key_state_predicate_t)(key_state_t *keyState);
```
firngrod: I was unsure of whether to go with a predicate or whether to bake the logic into the search function.  I like the predicate structure better as it leaves the Secondary Role business logic in `secondary_role_driver.c` file and the search logic in `postponer.c`.  I do not think there is any real performance cost to doing it this way.  Possibly there's a readability cost?

right/src/secondary_role_driver.c
```diff
@@ -102,6 +104,13 @@ void SecondaryRoles_FakeActivation(secondary_role_result_t res)
     }
 }
 
+bool isKeyAllowedToTriggerSecondary(key_state_t* keyState) {
+    return acceptTriggersFromSameHalf || !Utils_AreKeysOnTheSameHalf(
+        Utils_KeyStateToKeyId(resolutionKey),
```
firngrod: This could be cached.  Is RAM or Cycles more valuable?  My guess is RAM.

right/src/secondary_role_driver.c
```diff
@@ -130,18 +139,17 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
     uint32_t dualRolePressTime = resolutionStartTime;
     postponer_buffer_record_type_t *dummy;
     postponer_buffer_record_type_t *dualRoleRelease;
-    postponer_buffer_record_type_t *actionPress;
+    postponer_buffer_record_type_t *actionPress = NULL;
```
firngrod, line 105: Explicitly `NULL` here to ensure that even if `PostponerQuery_FindFirstReleased` doesn't set it because it's not run at all, it is `NULL`.

right/src/secondary_role_driver.c
```diff
@@ -130,18 +139,17 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
...
     postponer_buffer_record_type_t *actionRelease;
 
     PostponerQuery_InfoByKeystate(resolutionKey, &dummy, &dualRoleRelease);
     if (Cfg.SecondaryRoles_AdvancedStrategyTriggerByRelease) {
-        PostponerQuery_FindFirstReleased(&actionPress, &actionRelease);
+        PostponerQuery_FindFirstReleased(&actionPress, &actionRelease, isKeyAllowedToTriggerSecondary);
+        
     }
-    if (!Cfg.SecondaryRoles_AdvancedStrategyTriggerByRelease || actionPress == NULL) {
```
firngrod, line 110: The false assumption of a direct correlation between `SecondaryRoles_AdvancedStrategyTriggerByRelease` and `SecondaryRoles_AdvancedStrategyTriggerByPress` has cost me quite a bit of debug time by now.  I have made a commit where I collapse them into an enum of actions as seen below if that has any interest.  The way I see it, TriggerOnPress precludes TriggerOnRelease, and it's not very obvious that having neither is a real option.
```
    typedef enum {
        SecondaryRoleTriggerAction_None,
        SecondaryRoleTriggerAction_Press,
        SecondaryRoleTriggerAction_Release,
    } secondary_role_trigger_action_t;
```
Naming is obviously up for debate.

right/src/secondary_role_driver.c
```diff
@@ -38,6 +38,8 @@
  * - TriggerByRelease - if set to false, the only way to trigger secondary role is to press it for more than Timeout.
  * - DoubletapTime - if dual-role key is tapped and press within this time (press-to-press), then primary role is activated.
  * - DoubletapToPrimary - if set to false, doubletap logic is disabled. Useful if you prefer tap and hold evaluate to primary role.
+ * - MinimumHoldTime - The minimum time that a key must be held for in order to allow it to trigger as secondary.
+ * - AcceptTriggersFromSameHalf - Whether or not action keys can be from the same half as the resolution key.
```
firngrod, line 42: How up-to-date are these comments being kept?  I feel that they are quite eclipsed by the macro reference manual of which any developer is surely aware.

right/src/secondary_role_driver.c
firngrod: I could also pass a state variable through with the predicate, but I think that costs on readability.  Any opinions?

kareltucek: Depends. 

Where would you cache it? I assume in the postponer queue? A cost of one byte times 32 slots is not negligible. 

However, the actual computation is something like two instructions - a bit shift and a mask, isn't it? Cpu suffers here because of function call overheads.

right/src/secondary_role_driver.c
kareltucek: Just as any documentation or comments are in coding projects. When writing it, I felt the need to have a brief reference here, containing whatever I consider relevant for me as a programmer, not having to think about its exhaustiveness or other aspects of a documentation. 

I am not dead set on keeping it. If you have a strong desire to just delete it, then feel free to.

right/src/postponer.h
kareltucek: Well, right, what is actually relevant is that the call tree of those utils calls has become quite large, which imposes penalties on both performance and readability due to call overheads.
Also, looking at the code, my implementation of Utils_KeyStateToKeyId is hugely overengineered. That could be simplified down to a oneliner plus or minus edge-case handling - the 64 base is more or less set in stone now.

I prefer baking it into the search, unless you insist on the predicate.
Reasons:

readability cost
seems like a premature generalization / yagni
performance cost is tricky to judge because it depends on what you compare it to, but compared to an unpacked in-place computation, it is huge-ish

I guess the PostponerQuery_FindFirstReleased should issue a single call to Utils_AreKeysOnTheSameHalf with key_state_t* pointers as arguments.
If it makes sense to you, feel free to move Utils_AreKeysOnTheSameHalf logic into postponer (probably as a static function) and unpack its content there.

right/src/postponer.c
```diff
@@ -398,35 +398,45 @@ void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_typ
...
+void PostponerQuery_FindFirstPressed(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release,
```
kareltucek, line 401: Well, this is searching for a first press key that also has a relevant release in the queue, right? 

Is the existence of the release relevant in the press trigger scenarios?

kareltucek:
> I think the naming is much more clear on this behavior than on the previous one. Let's start a discussion on it!

They sound fine! :-)

kareltucek:
Do any edge cases need to be considered?

I have thought about the following, and figured it is actually the desired behavior:

- press a
- press s
- release s
- press j
- release j
- 'a' now triggers its secondary role
- if 's' is a dual-role key, it gets evaluated as primary, yet still applied as 'a+s'
- finally, j gets applied as 'a+j'
- this sounds as the desired behavior.

kareltucek (changes requested):
If you are up to it, then please do refactor the predicate part. 

Otherwise, let me know and we can merge it as is.

right/src/postponer.c
firngrod: It finds the first press and it's corresponding release if there is one.

firngrod: I will have to look through the code to see if the release is relevant.

right/src/postponer.h
firngrod: I will do that.  I'm inexperienced in microprocessor development, so I don't have a firm grasp of how much to optimize, and I tend to err or the side of generalization.

right/src/secondary_role_driver.c
firngrod: I would have cached it as resolutionKeyId, but it's going to be moved into the search function, so it's a moot point.

firngrod:
> Do any edge cases need to be considered?
> 
> I have thought about the following, and figured it is actually the desired behavior:
> 
>     * press a
> 
>     * press s
> 
>     * release s
> 
>     * press j
> 
>     * release j
> 
>     * 'a' now triggers its secondary role
> 
>     * if 's' is a dual-role key, it gets evaluated as primary, yet still applied as 'a+s'
> 
>     * finally, j gets applied as 'a+j'
> 
>     * this sounds as the desired behavior.

Yes, you are correct, the first key would go secondary and the two next keys would go primary with mod applied, even though the first one was from the same half.
I don't know what to call it.  It's well defined and expected behavior, but not the goal behavior.  Call it accepted behavior?  It's an unintended usecase of the feature which I think is not going to cause any issues.

right/src/postponer.c
kareltucek: Ah yes, sorry. 

Nevermind.

right/src/secondary_role_driver.c
kareltucek: Sounds fine to me :).

kareltucek:
I think it actually is the correct and desired behavior the way it is :). Just took a few minutes to realize that there is nothing strange about it and that it definitely isn't expected to be a multimod scenario.

firngrod:
I see it like if someone asked if you had considered that they new wooden wrist rest smelled differently from the rubber one.  Well, yes it does, it's not really a feature that was aimed for, but it is to be expected, and no one was expected to start actively sniffing it.

firngrod:
Once the other branch merges, I will rebase this one and start working on the changes.

firngrod:
Baked the predicated check into the search functions.
The release from `PostponerQuery_FindFirstPressed` is not used.  This will be fixed at some point.  Right now, I want to run these settings for a day or two to make sure they're okay.

right/src/key_states.h
```diff
@@ -53,4 +53,13 @@
     static inline bool KeyState_DeactivatedEarlier(key_state_t* s) { return !s->previous && !s->current; };
     static inline bool KeyState_NonZero(key_state_t* s) { return s->previous || s->current; };
 
+    static inline bool KeyState_IsRightHalf(key_state_t* s) { return s < &KeyStates[1][0]; };
```
firngrod, line 56: I know only one of these are in use.  I just wrote them all out to see which combination would be the most effective.
I am willing to remove the unused and move the used one elsewhere.

firngrod: I guess an `AreBothRight` could be just `if distance > 2 * moduleKeyStatesWidth`
AreBothLeft is harder, and the need to then check and flip signs makes it not worth it anyway.  Or I could just go for broke and switch the arrays up.  :D

right/src/postponer.c
```diff
@@ -398,35 +398,47 @@ void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_typ
...
+    *release = NULL;
+    *press = NULL;
+    return;
 }
 
-void PostponerQuery_FindFirstReleased(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release)
+void PostponerQuery_FindFirstReleased(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release,
+                                      key_state_t *opposingKey)
 {
+    bool opposingIsRight = KeyState_IsRightSide(opposingKey);
```
firngrod, line 429: This will be registered as right with a NULL pointer, but that's alright because we check for NULL later.

right/src/postponer.c
```diff
@@ -398,35 +398,47 @@ void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_typ
...
     if (bufferSize > 1) {
         for ( int i = 1; i < bufferSize; i++ ) {
             *release = &buffer[POS(i)];
             if ((*release)->event.type == PostponerEventType_ReleaseKey) {
-                for ( int j = 0; j < i; j++ ) {
-                    *press = &buffer[POS(j)];
-                    if((*press)->event.type == PostponerEventType_PressKey
-                        && (*press)->event.key.keyState == (*release)->event.key.keyState )
-                    return;
+                if (opposingKey == NULL || opposingIsRight != KeyState_IsRightSide((*release)->event.key.keyState)) {
```
firngrod, line 434: Every time I need an XOR operator for bools, I spend time googling and finding people tauting `(a && !b)||(!a && b)` before I remember to use `!=`.

kareltucek:
Alright, let's merge this! 

Would you write a changelog entry? (A couple of brief bullet points to be shown in the release notes about all your changes to secondary roles so far. Just post it here or anywhere.)

PR #1422, opened 2026-01-05
Ble fixes.

kareltucek:
This fixes a disconnect during ble hid pairing. 

Changes as imagined: there should be no reconnect when pairing is complete, and no suspicious logs.

Behaviors that might have been affected: switchover logic and pairing logic. 

This PR really could use a serious review and serious testing.

I think I will merge it with the max-style-slots PR and try to stress them both?

Changelog:
- Fix bluetooth disconnect during pairing.


mondalaci:
Yes, please do as suggested.

Issue #1421, opened 2026-01-05
Secondary Roles with positive safety margin does not finish resolution until timeout is reached or resolution key is released


firngrod:
The resolution is correct, it is just not reached until timeout is reached or some other key action triggers resolution after safety margin has passed.  A wakeup for the resolution is simply never set.
I will fix this myself, just putting it here to not forget.

firngrod:
Another one I just noticed in the code: Configured timeout action is ignored if
- Trigger On Release
- Timeout is reached after another key is pressed but not released
In that case, it will always trigger as secondary.

firngrod:
Another one gleaned from the code:  With negative safety margin trigger on release, release doesn't trigger resolution while an action key is pressed, not until timeout or action key release.  Again, the wakeup isn't scheduled.

All in all, the function is suffering from too much iterative development, and I'm in the process of overhauling it.

kareltucek:
> release doesn't trigger resolution while an action key is pressed, not until timeout or action key release 

(I may easily be misunderstanding/missing some context, but...)

(Reality check: the margin is _added_ to the time, so negative one should make the key resolve sooner assuming that we have an oracle, right?)

But we don't have an oracle, so it doesn't *know* that a key release is going to come within the next `abs(safetyMargin)`, so it cannot resolve before it, right? 

The only expected change in behavior is that once the release or timeout comes, the safety margin is taken into account when deciding the role.

Given this, the only possible change in behavior may be that you would expect it to resolve at `timeout - safetyMargin`, but that is not actual behavior change - it yields the same behavior one way or the other if you adjust the timeout.

kareltucek:
> Another one I just noticed in the code: Configured timeout action is ignored if
> 
>     * Trigger On Release
> 
>     * Timeout is reached after another key is pressed but not released
>       In that case, it will always trigger as secondary.

It may be better to create separate topics about these.

What is the scenario here?

- press dual role key
- press a
- press b
- now timeout is reached? 
- now you say that it resolves as secondary instead of the configured timeout action?

firngrod:
`a` has secondary role.  Timeout is large.  Safety margin is negative.   `b` is allowed to trigger secondary on `a`.
- `a` is pressed
- `b` is pressed
- `a` is released
- `b` is held for more than safety margin
- `a` does not resolve when safety margin runs out, but when `b` is released or timeout is reached

firngrod:
[...]

PR #1436, opened 2026-01-15
Firngrod/passive timeout

firngrod:
I added a new feature to do timeout passively, as I mentioned on the forum, although I had not fully realized yet that it could be implemented this simply.  Simply put, instead of applying the timeout action when timeout is reached, it will apply the timeout action once the key is released.  If timeout action is primary or secondary, that will be a tap of the key, if timeout action is none, it will do nothing,  The big difference here is that the key can be held for an arbitrarily long amount of time and still be used in a combo.
The main motivator is that I often press a secondary key to do a combo, then realize that it's on the same half.  Now I can just wait for half a second and release without my screen drowning in menus and without having to time my combos tightly.
It also allows the timeout to be effectively disabled by setting passive timeout to primary.
One downside is that now we have some setting combinations which make little sense in practice.  One that I can think of is passive timeout to secondary layer switch.  It would effectively be identical to passive timeout to none.  Maybe that could be added as a feature if someone wants it.

Also rewrote the main resolver to a fail fast style which removes a lot of ambiguity of where options apply.
EDIT:  Oops, I used a word I didn't understand.  It's not Fail Fast, it's just that I try to quit the function using as little info as possible.  What do I know of programming terms, I never went to programming school.

This PR fixes everything mentioned in #1421 and #1441

Changes:
Secondary Role keys (HRM):
Added option to postpone timeout action until release
Made Trigger by mouse subject to minimum hold time to prevent vibration on key press from triggering secondary role when using a trackball module
Made Trigger by mouse reliably trigger secondary role immediately
Fixed a number of edge cases to do with safety margin where secondary role keys would not trigger in a timely manner


doc-dev/reference-manual.md
```diff
@@ -148,6 +148,7 @@ COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
 COMMAND = set secondaryRole.advanced.timeout <ms, 0-500 (INT)>
 COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | none }
+COMMAND = set secondaryRole.advanced.timeoutType { active | passive } <defines whether the secondary role should perform it's timeout action when the timeout is reached or when the key is released after having been pressed beyond timeout without triggering secondary>
```
firngrod: Naming is obviously up for debate.  I see `passive` as the timeout passively change the primary action while `active` actively executes the timeout action.  It is entirely reasonable to see it as `passive` as timeout executes passively while you do nothing and `active` is that the timeout executes on the activity of release.  Probably different wording is in order.

doc-dev/user-guide.md
```diff
@@ -479,9 +479,10 @@ Firngrod's Home Row Mods (HRM) configuration (Recommended):
 set secondaryRole.defaultStrategy advanced
 set secondaryRole.advanced.triggeringEvent release
 set secondaryRole.advanced.timeout 300
-set secondaryRole.advanced.timeoutAction secondary
+set secondaryRole.advanced.timeoutType passive
+set secondaryRole.advanced.timeoutAction none
```
firngrod, line 483: These are my settings, don't know if they should be recommended.  The idea has been described on the Forum.  Basically, mods only activate in combinations with a non-mod, otherwise not.  If I need to activate a mod (combo) on it's own, I will activate it with a secondary key on the other half, but hold that key long enough to get a `none` timeout action.  With active secondary timeout, I would definitely recommend these settings still.

doc-dev/user-guide.md
```diff
@@ -491,8 +492,9 @@ Full set of advanced strategy config values follows (for copy-paste convenience)
 
 ```
 set secondaryRole.defaultStrategy advanced
-set secondaryRole.advanced.timeout 500
+set secondaryRole.advanced.timeout 200
```
firngrod, line 495: The text below states 200 ms.  On second thought, I should update the text instead.  200 ms timeout is aggressive.

right/src/postponer.c
```diff
@@ -404,33 +404,22 @@ void PostponerQuery_InfoByKeystate(key_state_t* key, postponer_buffer_record_typ
     }
 }
 
-void PostponerQuery_FindFirstPressed(postponer_buffer_record_type_t** press, postponer_buffer_record_type_t** release,
-                                     key_state_t *opposingKey)
+void PostponerQuery_FindFirstPressed(const postponer_buffer_record_type_t** press, const key_state_t *opposingKey)
```
firngrod, line 407: Release doesn't matter for the trigger on press and vice versa.  We only actually need the time of the event to enforce safety margin.  I don't know if we should reduce the function to just return a timestamp?

right/src/postponer.c
```diff
@@ -439,17 +428,16 @@ void PostponerQuery_FindFirstReleased(postponer_buffer_record_type_t** press, po
             if ((*release)->event.type == PostponerEventType_ReleaseKey) {
                 if (opposingKey == NULL || opposingIsRight != KeyState_IsRightSide((*release)->event.key.keyState)) {
                     for ( int j = 0; j < i; j++ ) {
-                        *press = &buffer[POS(j)];
-                        if ((*press)->event.type == PostponerEventType_PressKey
-                            && (*press)->event.key.keyState == (*release)->event.key.keyState ) {
+                        const postponer_buffer_record_type_t * const press = &buffer[POS(j)];
+                        if (press->event.type == PostponerEventType_PressKey
```
firngrod, line 432: Here we still need to find the press to ensure that it's there.

right/src/secondary_role_driver.c
```diff
@@ -64,9 +65,10 @@ static bool resolutionCallerIsMacroEngine = false;
 static bool acceptTriggersFromSameHalf = true;
 static key_state_t *resolutionKey;
 static uint32_t resolutionStartTime;
+static bool isDoubletap = false;
+static bool currentlyResolving = false;
+
 
-static key_state_t *previousResolutionKey;
```
firngrod, line 68: These values were only used for doubletap, so save some RAM and cycles there

right/src/secondary_role_driver.c
```diff
@@ -97,143 +99,114 @@ static void sleepTimeoutStrategy(uint16_t wakeTimeOffset) {
     }
 }
 
+
+#define RESOLVED(resolution)    KEY_TIMING(KeyTiming_RecordComment(resolutionKey, resolution, __LINE__))  \
```
firngrod, line 103: Not quite as readable output as before, but a lot easier to use these macros.

right/src/secondary_role_driver.c
```diff
@@ -97,143 +99,114 @@ static void sleepTimeoutStrategy(uint16_t wakeTimeOffset) {
...
+                                return resolution;
+#define AWAITEVENT(timeout)        sleepTimeoutStrategy(timeout);                                  \
+                                return SecondaryRoleState_DontKnowYet;
+
 static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
```
firngrod, line 108: Don't try to compare the new lines to the removed ones.  The entire thing is rewritten.

right/src/secondary_role_driver.c
```diff
@@ -97,143 +99,114 @@ static void sleepTimeoutStrategy(uint16_t wakeTimeOffset) {
...
         }
     }
 
-    //handle trigger by press
-    if (Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent == SecondaryRoleTriggeringEvent_Press) {
-        bool actionKeyWasPressedButDualkeyNot = actionPress != NULL && dualRoleRelease == NULL && (int32_t)(Timer_GetCurrentTime() - actionPress->time) > Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
-        bool actionKeyWasPressedFirst = actionPress != NULL && dualRoleRelease != NULL && actionPress->time <= dualRoleRelease->time - Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
+    // see if we have an activating key event
+    const postponer_buffer_record_type_t *actionEvent = NULL;
+    const key_state_t * const opposingKey = acceptTriggersFromSameHalf ? NULL : resolutionKey;
```
firngrod, line 136: Heavy const syntax

right/src/secondary_role_driver.c
```diff
@@ -97,143 +99,114 @@ static void sleepTimeoutStrategy(uint16_t wakeTimeOffset) {
...
     }
 
-    // handle primary press when action key is pressed and trigger behaviors are off
-    if (dualRoleRelease != NULL && !triggerBehaviorsActive && activeTime < Cfg.SecondaryRoles_AdvancedStrategyTimeout) {
-        KEY_TIMING(KeyTiming_RecordComment(resolutionKey, "PK"));
-        return SecondaryRoleState_Primary;
+    // handle action key activation
+    if (actionEvent != NULL) {
+        RESOLVED(SecondaryRoleState_Secondary);
     }
```
firngrod, line 203: This was so anticlimactic to write once I got here.

right/src/secondary_role_driver.c
```diff
@@ -97,143 +99,114 @@ static void sleepTimeoutStrategy(uint16_t wakeTimeOffset) {
...
 
-    sleepTimeoutStrategy(Cfg.SecondaryRoles_AdvancedStrategyTimeout);
+    // see if we should set a timer to wake up to actively time out
+    if (isActiveTimeout || isDoubletap) {
+        AWAITEVENT(Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout);
+    }
+    
+    // otherwise, keep postponing until key action
     return SecondaryRoleState_DontKnowYet;
```
firngrod, line 211: The user can fill the queue while holding indefinitely.  They will loose keystrokes.  I'm okay with that.

kareltucek:
Btw., a bit of commit message philosophy, since I am doing git archeology right now: it would be nice if I could tell straight away what part of the codebase a commit belongs to just from the message. In the case of your PRs, it might mean just prefixing messages with a short codeword like "hrm". E.g., "hrm: Embarrassing".

This is just a general feedback. No need to rebase anything ;-).

kareltucek:
(I am guilty too though: "pick c6198af1 # Reword some code.".)

firngrod:
Makes sense.  Where I work, we do the related story prefix, so I always do a double take anyway while I try to remember what the story number was.  Might as well harness that.  :)

firngrod:
This branch now also addresses https://github.com/UltimateHackingKeyboard/firmware/issues/1441

right/src/secondary_role_driver.c
```diff
@@ -298,18 +286,26 @@ static void startResolution(
...
+            Macros_SetStatusNumSpaced(Timer_GetCurrentTime(), false);
+            Macros_SetStatusString(" Got resolve request.\n", NULL);
+        }
         activateSecondaryImmediately = true;
+        if (resolutionCallerIsMacroEngine) {
+            Macros_WakeBecauseOfKeystateChange();
+        }
+        else {
+            EventVector_Set(EventVector_NativeActions);
+        }
```
firngrod, line 308: I'm not sure if these are the correct ways to force a reevaluation, but it seems to work.

firngrod:
I have run this for a week with no issues now.

doc-dev/reference-manual.md
```diff
@@ -704,6 +705,7 @@ Key actions can be parametrized with macro arguments. These arguments can be exp
     - advanced strategy may trigger secondary role depending on timeout, or depending on key release order.
       - `set secondaryRole.advanced.timeout <timeout in ms, 350 (INT)>` if this timeout is reached, `timeoutAction` (secondary by default) role is activated.
       - `set secondaryRole.advanced.timeoutAction { primary | secondary | none}` defines whether the primary action or the secondary role, or no action at all, should be activated when timeout is reached
+      - `set secondaryRole.advanced.timeoutType { active | passive }` defines whether the secondary role should perform it's timeout action when the timeout is reached or when the key is released after having been pressed beyond timeout without triggering secondary
```
kareltucek: Please extend this. I can't wrap my head around this.

doc-dev/reference-manual.md
kareltucek: Please try to keep these hints shorter. 80ish or so characters inside the <> tops

doc-dev/reference-manual.md
```diff
@@ -148,6 +148,7 @@ COMMAND = set module.touchpad.holdContinuationTimeout <0-65535 (INT)>
 COMMAND = set secondaryRole.defaultStrategy { simple | advanced }
 COMMAND = set secondaryRole.advanced.timeout <ms, 0-500 (INT)>
 COMMAND = set secondaryRole.advanced.timeoutAction { primary | secondary | none }
+COMMAND = set secondaryRole.advanced.timeoutType { active | passive } <defines whether the secondary role should perform it's timeout action when the timeout is reached or when the key is released after having been pressed beyond timeout without triggering secondary>
 COMMAND = set secondaryRole.advanced.safetyMargin <ms, higher value adjusts sensitivity towards primary role -50-50 (INT)>
 COMMAND = set secondaryRole.advanced.triggeringEvent { press | release | none } <determines whether secondary role is triggered when another key is pressed or when another key is pressed and released while the secondary role key is being held, or not at all from other keys being pressed>
```
kareltucek: Same here

right/src/postponer.c
kareltucek: No. I prefer the information to be there in case anyone needs them.

doc-dev/reference-manual.md
firngrod: The current behavior, which I have called `active` will perform the timeout action immediately on timeout. The new behavior I have added, called `passive` will perform the timeout action once the key is released, if the key has timed out. The key difference is that with `passive` timeout, it can still become secondary after timeout until the key is released. So now, with `passive` timeout action `none`, I can hold the key for a long time and still combo the secondary role, or release without activating start or context menus.

firngrod: I'm open to other ways of doing this, but this way also allows effectively disabling timeout by setting `passive` timeout `primary` action, which is nice.

kareltucek: Ah, right, the key difference is that it cannot become primary but can become secondary.

It is fine, just the documentation needs to be polished a bit. 

Suggested wording:

```
      - `set secondaryRole.advanced.timeoutType { active | passive }` 
        - active - the timeout action is activated immediately upon timeout.
        - passive - the timeout action is delayed until release. This way, after the timeout, the action cannot become primary anymore, but can still become secondary.
```

I am wondering if this should be configurable at all, or the only allowed way. But it is definitely fine like this.

firngrod: How about this phrasing:

`set secondaryRole.advanced.timeoutType { active | passive }` defines how the timeout action is applied.  With `active` timeout, the timeout action is applied the moment the timeout has been reached.  With `passive` timeout, if the key fails to become secondary and was held longer than the timeout time, when it's released, the key will activate the timeout action rather than it's primary function.

kareltucek: Sorry, reply got stuck with the rest of the review batch.

Will submit the review now and follow up later with more.

doc-dev/reference-manual.md
kareltucek: Naming sounds sensible to me :-).

right/src/secondary_role_driver.c
kareltucek: Overoptimization if you ask me, but yes, why not have it ;-).

right/src/secondary_role_driver.c
kareltucek: Yes, they are correct!

kareltucek (approved):
Nice refactor! 

Just please rephrase those docs.

kareltucek:
Also please write a changelog. (List the changes, including all fixes.)

firngrod:
I'm traveling for work until Wednesday evening and didn't bring my own computer, so I will address things once I get home.

firngrod:
> I am wondering if this should be configurable at all, or the only allowed way. But it is definitely fine like this.

The active timeout should definitely still be an option.

> Also please write a changelog. (List the changes, including all fixes.)

Where do I write that?  Just as a comment in this PR?

kareltucek:
> The active timeout should definitely still be an option.

Ok!

> Where do I write that? Just as a comment in this PR?

Ideally edit the leading comment of the PR please.

PR #1443, opened 2026-01-19
Led commands2

### LINKED ISSUE #322: Expose LEDs via USB
This issue contains a specification for exposing LED states (both of the LED display and the per-key LEDs) via getter and setter USB commands, which is useful for displaying custom information and animations.

First up, let's add the `UsbVariable_LedOverride` struct and expose it via the getter/setter interface of `right/src/usb_commands/usb_command_{get,set}_variable.[ch]`. The following byte fields should be featured:

- letters (the LEDs of the 3 letter keymap abbreviation)
- layers (the 3 layer LEDs)
- capsLock (the caps lock icon)
- agent (the Agent icon)
- adaptiveMode (the triangle icon of the adaptive mode)
- keys (per-key LEDs)

By default, these fields are set to 0, meaning that the UHK firmware is in control of all the LEDs. By setting a field to 1, the relevant LEDs won't be controlled by the UHK firmware anymore, and the user will be able to override their value.

Let's add the `GetLedValues` USB command. The input parameters of this command are `slotId` and `ledIndex`, and based on these parameters, `GenericHidOutBuffer` should be filled up with the relevant LED values taken from `LedDriverValues`.

Let's also add the `SetLedValues` USB command. The first argument is the `slotId` byte. Subsequent arguments are byte-pairs of `ledId` (within the slot) and the related `ledValue` to be set. A `ledId` of 255 value should be ignored, and it marks the end of the slot's LED settings, meaning that a new slot id and led index, value pairs follow. This way, bytes with `255` value should be used for padding the unused part of `GenericHidOutBuffer`.

I also have a good idea regarding the host-side API, so please open a new Agent issue before starting to implementing it, even if your language of preference is not TypeScript or JavaScript because I'd like to keep libraries of different languages consistent.

kareltucek:
Closes #322

See doc-dev/other/usb_led_controls.md for usage examples.

Changelog:
- Add USB API for led control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@ build
 .vscode/settings.json
 .jj
 mcux_include.json
+compile_commands.json
+.devices
+Makefile
+createTags.sh
+tags
--- a/device/src/keyboard/oled/oled.c
+++ b/device/src/keyboard/oled/oled.c
@@ -12,6 +12,8 @@
 
 #if DEVICE_HAS_OLED
 
+uint8_t OledOverrideMode = 0;
+
 #include <zephyr/kernel.h>
 #include <zephyr/drivers/gpio.h>
 #include "keyboard/oled/oled.h"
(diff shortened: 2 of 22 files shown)


mondalaci:
I think I'll need https://github.com/UltimateHackingKeyboard/agent/issues/2334 for testing.

Which will also resolve https://github.com/UltimateHackingKeyboard/agent/issues/660

Issue #1446, opened 2026-01-22
ifHold does not activate on timeout


mhantsch:
`ifHold` currently does not activate until I release the key. It should already activate on timeout.

My steps to reproduce:

**graveTilde** macro bound to the `` ` ~`` key:
```
ifShortcut 1 exec quoteLong
ifHold exec quoteLong
holdKey graveAccentAndTilde
```

When I tap it, I get a grave accent. ✔️ 
When I combo it with 1, I immediately get the quoteLong result. ✔️ 
When I hold it, nothing happens until I release it. Then I get the quoteLong result. 🤔 

This should resolve as soon as the timeout is reached.

Configuring `set holdTimeout 350` in **$onInit** doesn't make a difference.

**quoteLong** just types two lines of triple backticks with an empty line in between, and places the cursor on the empty line, ready to have some code pasted into it:
```
tapKeySeq graveAccentAndTilde graveAccentAndTilde graveAccentAndTilde
tapKeySeq enter enter
tapKeySeq graveAccentAndTilde graveAccentAndTilde graveAccentAndTilde enter
tapKeySeq up up
```

firngrod:
While you hold the key, `ifShortcut` waits indefinitely for another key.  Then when you release, `ifHold` evaluates on a now very long keypress and evaluates true.  Try adding a timeout to `ifShortcut`.

mhantsch:
Why does it wait indefinitely and not just until the timeout I've configured in Typing behaviour?

firngrod:
Because it's not a secondary key feature, it's it's own thing.  I can understand the confusion.

mhantsch:
Ah, it would help in the documentation to explain which macro commands are affected by which config parameter, and which takes which defaults from where.

mhantsch:
You were of course right, @firngrod. Modifying the macro with a timeout for `ifShortcut` makes it work as expected:

```
ifShortcut timeoutIn 200 1 exec quoteLong
ifHold exec quoteLong
holdKey graveAccentAndTilde
```

So, `ifHold` works fine. This issue can be closed.

kareltucek:
> Ah, it would help in the documentation to explain which macro commands are affected by which config parameter, and which takes which defaults from where.

Please provide a specific wording ;-)

mhantsch:
Pull request with suggested changes to reference docs.

PR #1448, opened 2026-01-23
Update reference manual for better description of ifShortcut/ifGestur…

mhantsch:
…e options

Clarify behavior of if(Not)Shortcut/if(Not)Gesture commands and their options.
(See discussion in https://github.com/UltimateHackingKeyboard/firmware/issues/1446#issuecomment-3789173819.)

Rebased version in https://github.com/UltimateHackingKeyboard/firmware/pull/1456


doc-dev/reference-manual.md
```diff
@@ -508,13 +508,13 @@ We allow postponing key activations in order to allow deciding between some scen
...
-  - `Shortcut` requires continual press of keys (e.g., Ctrl+c). By default, it timeouts with the activation key release.
-  - `Gesture` allows a noncontinual sequence of keys (e.g., vim's gg). By default, timeouts in 1000 ms since activation.
+- `ifShortcut/ifNotShortcut/ifGesture/ifNotGesture [IFSHORTCUT_OPTIONS]* [KEYID]*` will wait for next keypresses until sufficient number of keys have been pressed. The default - without use of `timeoutIn` or `cancelIn` - is to wait forever. If the next keypresses correspond to the provided arguments (hardware ids), the keypresses are consumed and the condition is performed. Both `if` and `ifNot` versions will consume the keypresses if the full list is matched. E.g., `ifShortcut 090 089 final tapKey C-V; holdKey v`. `noConsume` can be used to keep the remaining keys for further processing.
```
firngrod: I would rephrase a little:
The default - without use of `timeoutIn` or `cancelIn` - is to wait forever.  =>  The default behavior without `timeoutIn` or `cancelIn` is to wait forever or until release of the original key for additional key presses.

mhantsch: It's already clear that we are waiting "for additional key presses", so I would leave that part out.

New suggestion: The default behavior (without `timeoutIn` or `cancelIn`) is to wait forever or until release of the original key.

Also, I want to replace "hardware ids"  in the next sentence with KEYIDs. I'll update the pull request.

mhantsch (commented):
Thinking more about this, it's also not completely true. `ifShortcut` will indeed end once the original key is released, but `ifGesture` will continue waiting for more keys, won't it? The idea is that it waits for a sequence of separately tapped keys. What's the default timeout for `ifGesture`?

kareltucek:
Default timeout is 1000ms, but Laszlo complains that it is too long, so I will probably reduce it to 500ms.

mhantsch:
For me, it would make sense to have a default timeout configuration for `ifGesture` and `ifShortcut` so that you don't have to write `timeoutIn` into all your macros. This would be similar to `holdTimeout`. Maybe something like:

```
// this one already exists:
set holdTimeout 350
// the next two are new:
set gestureBehaviour.Timeout 400
set shortcutBehaviour.Timeout 250
```

Next, I'll try rephrasing the reference docs again. I am not satisfied yet.
Though.... do we need defaults for `cancelIn` also? That feels like overkill.

firngrod:
I'm not sure I would agree with a default timeout.  I think coupling it with an `ifHold` is a very rare usecase anyway.
Whether a blanket config value is necessary depends on whether we expect users to use the command a lot across many macros or not I guess.  But it's not inconceivable for them to do so, so I will not actively resist the idea, though clever use of variables can also solve that problem.

mhantsch (commented):
Let's try this variant of a description.

doc-dev/reference-manual.md
```diff
@@ -508,13 +508,13 @@ We allow postponing key activations in order to allow deciding between some scen
...
+- `ifShortcut/ifNotShortcut/ifGesture/ifNotGesture [IFSHORTCUT_OPTIONS]* [KEYID]*` will wait for next keypresses until a sufficient number of keys have been pressed, either simultaneously (Shortcut) or in a sequence (Gesture). If the next keypresses correspond to the provided arguments (KEYIDs), the keypresses are consumed and the condition is executed. Both `if` and `ifNot` versions will consume the keypresses if the full list is matched; the `noConsume` option can be used to keep the remaining keys for further processing.
+  - `Shortcut` requires continuous press of two or more keys (e.g., Ctrl+c), sometimes also called a *combo* or *chord*. When all the keys are pressed together, it executes the conditional command. It aborts when either the activation key is released, or a specific timeout is reached (`timeoutIn` or `cancelIn`). In the absence of timeouts, it can wait forever; there is no default timeout.
+  - `Gesture` matches a non-continuous sequence of keys (e.g., vim's gg). When all the keys are tapped in order, and within the timeout, it executes the conditional command. By default, the timeout is 1000 ms, and can be overridden with `timeoutIn` and `cancelIn`.
```
firngrod: Non-continuous doesn't sound right to me. How about 'non-simultaneous'?

kareltucek:
Well, I would rather keep that npm that actually doesn't break local, and fix the ci later, and in a separate PR. 

Would you rebase it, squash it and force push again?

mhantsch:
I'll try. I'm not the best git expert.

kareltucek:
I can do it myself then.

mhantsch:
Thanks

kareltucek:
Superseeded by https://github.com/UltimateHackingKeyboard/firmware/pull/1456

PR #1450, opened 2026-01-25
Add a notify macro command.

kareltucek:
(Not yet tested.)


pcooke9:
Nice! It would be great if we could change text size. Maybe an option to scale the notifications to fill the screen?

kareltucek:
Sorry, but I don't want to complicate neither the code nor the macro syntax and definitely not now 😉.

pcooke9:
Understandable, a notify macro is a welcome addition regardless.

PR #1470, opened 2026-01-30
Bluetooth: always exchange MTU right after connection

kareltucek:
## Summary
- Proactively initiate ATT MTU exchange in the `connected` callback for all BLE connections
- Previously, HID connections relied on the host to initiate MTU exchange, which sometimes failed after host switching
- This fixes "No ATT channel for MTU 32" and "No buffer available to send notification" warnings

## Background
The default ATT MTU is 23 bytes (20 bytes payload). When HID reports exceed 20 bytes, notifications fail if MTU hasn't been negotiated higher. NUS connections already performed MTU exchange, but HID connections did not.

## Test plan
- [x] Test BLE HID connection to multiple hosts
- [x] Test switching between BLE hosts multiple times
- [x] Verify "MTU exchange done" log appears for connections
- [x] Verify no more "No ATT channel for MTU" warnings after switching

🤖 Generated with [Claude Code](https://claude.ai/code)


kareltucek:
This seems to significantly reduce the number of ble hid interface fails (mouse or keys not working).

PR #1476, opened 2026-02-04
Fix i2c slave scheduler regression from uart refactors.

kareltucek:
Changelog:
- Fix uhk60 i2c driver bugs introduced in 16.1.


kareltucek:
[firmware.zip](https://github.com/user-attachments/files/25093513/firmware.zip)

Issue #1453, opened 2026-01-26
hrm: exclude certain keys from "same half" detection


mhantsch:
I have assigned **Tab** to the `left fn` thumb key of the left hand. 
I have assigned **LAlt** as secondary to the `a` key (also left hand).
I would like to tap Alt-Tab (switch windows) quickly with my left hand, so I press `a`+`left fn` quickly and release it quickly. Now the same hand detection kicks in and I get the letter **a** followed by a **Tab** press.

I'd like to keep same hand detection for general use of the `a` key to protect me from spurious activations of **LAlt** during fast typing.
But I want to exclude `left fn`.

*Note:* in ZMK and Kanata, I can freely configure which keys trigger immediate ("same hand") resolution to primary. I can even set up different behaviour per key-with-secondary-function.

kareltucek:
Well, lets say that I might be willing to allocate a bit of data per key for it.

mhantsch:
As an option, I could see this working as a macro code option to `ifPrimary`/`ifSecondary`:

```
ifSecondary ignoreTriggersFromSameHalf acceptTriggersFromKeys leftFn final holdKey leftAlt
holdKey a
```

I don't know if that would change and/or require at all that bit allocation per key.

It also would allow a lot more flexibility as it could be changed for every `ifSecondary` call, so different dual-role keys could have different trigger key lists.

However, to make this usable in macros, other issues need to be resolved, too:
https://github.com/UltimateHackingKeyboard/firmware/issues/1445 (required)
https://github.com/UltimateHackingKeyboard/firmware/issues/1421 (may be needed)

firngrod:
#1421 is fixed in my current PR.
May I suggest a solution where you do a postponed `ifSecondary accepTriggersFromSameHalf`, then look through the queue for first released key, and use that info to decide whether to do secondary or primary? Would be tricky to handle if there are more than one released key though.

mhantsch:
> May I suggest a solution where you do a postponed `ifSecondary accepTriggersFromSameHalf`, then look through the queue for first released key, and use that info to decide whether to do secondary or primary?

You lost me, sorry. Not sure how I could do that from the macro.

mhantsch:
Now that https://github.com/UltimateHackingKeyboard/firmware/issues/1445 is fixed, I was able to work around it with this macro:

```
ifShortcut timeoutIn 100 orGate noConsume leftFn goTo secondary
ifPrimary final holdKey a
secondary:
[...]

PR #1477, opened 2026-02-04
Firngrod/global doubletap handling

firngrod:
Running this now, seems to work fine.
I want to run it for a bit to make sure.

I only propagated the press as far as it needed to go down the tree in `usb_report_updater.c`.
I think maybe the sequence number for macros should go into the key press info.
Do you have a procedure for remembering to remove removed config values in the future?

Changelog:
Fixed an issue where secondary role could incorrectly trigger doubletaps
Fixed an issue where macros could incorrectly trigger doubletaps
Macros will now remember and reuse secondary role resolutions for the rest of their runtime, regardless of key events

### DIFF
--- a/right/src/config_parser/parse_config.c
+++ b/right/src/config_parser/parse_config.c
@@ -179,7 +179,6 @@ parser_error_t parseConfig(config_buffer_t *buffer)
     uint16_t keystrokeDelay = Cfg.KeystrokeDelay;
 
     secondary_role_strategy_t secondaryRoles_Strategy = SecondaryRoleStrategy_Simple;
-    uint16_t secondaryRoles_AdvancedStrategyDoubletapTimeout = Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout;
     uint16_t secondaryRoles_AdvancedStrategyTimeout = Cfg.SecondaryRoles_AdvancedStrategyTimeout;
     int16_t secondaryRoles_AdvancedStrategySafetyMargin = Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
     secondary_role_triggering_event_t secondaryRoles_AdvancedStrategyTriggeringEvent = Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent;
@@ -188,7 +187,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
     if (DataModelVersion.major >= 7) {
         secondaryRoles_Strategy = ReadUInt8(buffer);
-        secondaryRoles_AdvancedStrategyDoubletapTimeout = ReadUInt16(buffer);
+        ATTR_UNUSED uint16_t secondaryRoles_AdvancedStrategyDoubletapTimeout = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategyTimeout = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategySafetyMargin = ReadInt16(buffer);
         secondaryRoles_AdvancedStrategyTriggeringEvent = ReadBool(buffer) ? SecondaryRoleTriggeringEvent_Release : SecondaryRoleTriggeringEvent_Press;
@@ -346,7 +345,6 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
         if (DataModelVersion.major >= 7) {
             Cfg.SecondaryRoles_Strategy = secondaryRoles_Strategy;
-            Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout = secondaryRoles_AdvancedStrategyDoubletapTimeout;
             Cfg.SecondaryRoles_AdvancedStrategyTimeout = secondaryRoles_AdvancedStrategyTimeout;
             Cfg.SecondaryRoles_AdvancedStrategySafetyMargin = secondaryRoles_AdvancedStrategySafetyMargin;
             Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent = secondaryRoles_AdvancedStrategyTriggeringEvent;
--- a/right/src/config_parser/parse_config.c
+++ b/right/src/config_parser/parse_config.c
@@ -179,7 +179,6 @@ parser_error_t parseConfig(config_buffer_t *buffer)
     uint16_t keystrokeDelay = Cfg.KeystrokeDelay;
 
     secondary_role_strategy_t secondaryRoles_Strategy = SecondaryRoleStrategy_Simple;
-    uint16_t secondaryRoles_AdvancedStrategyDoubletapTimeout = Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout;
     uint16_t secondaryRoles_AdvancedStrategyTimeout = Cfg.SecondaryRoles_AdvancedStrategyTimeout;
     int16_t secondaryRoles_AdvancedStrategySafetyMargin = Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
     secondary_role_triggering_event_t secondaryRoles_AdvancedStrategyTriggeringEvent = Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent;
@@ -188,7 +187,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
     if (DataModelVersion.major >= 7) {
         secondaryRoles_Strategy = ReadUInt8(buffer);
-        secondaryRoles_AdvancedStrategyDoubletapTimeout = ReadUInt16(buffer);
+        uint16_t _ = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategyTimeout = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategySafetyMargin = ReadInt16(buffer);
         secondaryRoles_AdvancedStrategyTriggeringEvent = ReadBool(buffer) ? SecondaryRoleTriggeringEvent_Release : SecondaryRoleTriggeringEvent_Press;
@@ -346,7 +345,6 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
         if (DataModelVersion.major >= 7) {
             Cfg.SecondaryRoles_Strategy = secondaryRoles_Strategy;
-            Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout = secondaryRoles_AdvancedStrategyDoubletapTimeout;
             Cfg.SecondaryRoles_AdvancedStrategyTimeout = secondaryRoles_AdvancedStrategyTimeout;
             Cfg.SecondaryRoles_AdvancedStrategySafetyMargin = secondaryRoles_AdvancedStrategySafetyMargin;
             Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent = secondaryRoles_AdvancedStrategyTriggeringEvent;
--- a/right/src/config_parser/parse_config.c
+++ b/right/src/config_parser/parse_config.c
@@ -179,6 +179,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
     uint16_t keystrokeDelay = Cfg.KeystrokeDelay;
 
     secondary_role_strategy_t secondaryRoles_Strategy = SecondaryRoleStrategy_Simple;
+    uint16_t secondaryRoles_AdvancedStrategyDoubletapTimeout = Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout;
     uint16_t secondaryRoles_AdvancedStrategyTimeout = Cfg.SecondaryRoles_AdvancedStrategyTimeout;
     int16_t secondaryRoles_AdvancedStrategySafetyMargin = Cfg.SecondaryRoles_AdvancedStrategySafetyMargin;
     secondary_role_triggering_event_t secondaryRoles_AdvancedStrategyTriggeringEvent = Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent;
@@ -187,7 +188,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
     if (DataModelVersion.major >= 7) {
         secondaryRoles_Strategy = ReadUInt8(buffer);
-        uint16_t _ = ReadUInt16(buffer);
+        secondaryRoles_AdvancedStrategyDoubletapTimeout = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategyTimeout = ReadUInt16(buffer);
         secondaryRoles_AdvancedStrategySafetyMargin = ReadInt16(buffer);
         secondaryRoles_AdvancedStrategyTriggeringEvent = ReadBool(buffer) ? SecondaryRoleTriggeringEvent_Release : SecondaryRoleTriggeringEvent_Press;
@@ -345,6 +346,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
         if (DataModelVersion.major >= 7) {
             Cfg.SecondaryRoles_Strategy = secondaryRoles_Strategy;
+            Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout = secondaryRoles_AdvancedStrategyDoubletapTimeout;
             Cfg.SecondaryRoles_AdvancedStrategyTimeout = secondaryRoles_AdvancedStrategyTimeout;
             Cfg.SecondaryRoles_AdvancedStrategySafetyMargin = secondaryRoles_AdvancedStrategySafetyMargin;
             Cfg.SecondaryRoles_AdvancedStrategyTriggeringEvent = secondaryRoles_AdvancedStrategyTriggeringEvent;
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -4,40 +4,42 @@
 
 typedef enum {
     DoubletapState_PossibleFirst,
-    DoubletapState_Doubletap,
-    DoubletapState_Blocked,
+    DoubletapState_PossibleSecond,
+    DoubletapState_IneligibleForFirst,
 } doubletap_state_t;
 
 typedef struct {
     const key_state_t *keyState;
     uint32_t timestamp;
-    doubletap_state_t doubletapState: 2;
+    uint32_t previousPressTime;
+    doubletap_state_t doubletapState;
 } previous_key_event_type_t;
 
 static previous_key_event_type_t lastPress;
 
 void KeyHistory_RecordPress(const key_state_t *keyState)
 {
-    bool isDoublePress = 
-        keyState == lastPress.keyState
-        && lastPress.doubletapState == DoubletapState_PossibleFirst
-        && CurrentPostponedTime < lastPress.timestamp + Cfg.DoubletapTimeout;
-
     lastPress = (previous_key_event_type_t){
         .keyState = keyState,
         .timestamp = CurrentPostponedTime,
-        .doubletapState = isDoublePress ? DoubletapState_Doubletap : DoubletapState_PossibleFirst,
+        .previousPressTime = lastPress.timestamp,
+        .doubletapState = lastPress.doubletapState == DoubletapState_PossibleFirst ? DoubletapState_PossibleSecond : DoubletapState_PossibleFirst,
     };
 }
 
 void KeyHistory_RecordRelease(const key_state_t *keyState)
 {
     if (keyState != lastPress.keyState) {
-        lastPress.doubletapState = DoubletapState_Blocked;
+        lastPress.doubletapState = DoubletapState_IneligibleForFirst;
     }
 }
 
-bool KeyHistory_WasLastDoubletap()
+bool KeyHistory_WasLastDoubletap(const uint32_t maxInterval)
 {
-    return lastPress.doubletapState == DoubletapState_Doubletap;
+    bool wasDoubletap = lastPress.doubletapState == DoubletapState_PossibleSecond
+                        && lastPress.timestamp - lastPress.previousPressTime < maxInterval;
+    if (wasDoubletap) {
+        lastPress.doubletapState = DoubletapState_IneligibleForFirst;
+    }
+    return wasDoubletap;
 }
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -1,0 +1,46 @@
+#include "key_history.h"
+#include "postponer.h"
+
+typedef enum {
+    DoubletapState_Eligible,
+    DoubletapState_Consumed,
+    DoubletapState_Ineligible,
+} doubletap_state_t;
+
+typedef struct {
+    const key_state_t *keyState;
+    uint32_t timestamp;
+    doubletap_state_t doubletapState: 2;
+} previous_key_event_type_t;
+
+static previous_key_event_type_t lastPress;
+
+void KeyHistory_RecordPress(const key_state_t *keyState)
+{
+    lastPress = (previous_key_event_type_t){
+        .keyState = keyState,
+        .timestamp = CurrentPostponedTime,
+        .doubletapState = lastPress.doubletapState == DoubletapState_Consumed 
+            ? DoubletapState_Ineligible
+            : DoubletapState_Eligible,
+    };
+}
+
+void KeyHistory_RecordRelease(const key_state_t *keyState)
+{
+    if (keyState != lastPress.keyState) {
+        lastPress.doubletapState = DoubletapState_Ineligible;
+    }
+}
+
+bool KeyHistory_IsDoubletap(const key_state_t *keyState, const uint16_t doubletapInterval)
+{
+    uint32_t interval = CurrentPostponedTime - lastPress.timestamp;
+    if(keyState != lastPress.keyState 
+        || lastPress.doubletapState != DoubletapState_Eligible
+        || interval > doubletapInterval) {
+        return false;
+    }
+    lastPress.doubletapState = DoubletapState_Consumed;
+    return true;
+}
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -1,13 +1,11 @@
 #include "config_manager.h"
 #include "key_history.h"
-#include "macros/status_buffer.h"
 #include "postponer.h"
 
 typedef enum {
     DoubletapState_PossibleFirst,
-    DoubletapState_PossibleSecond,
-    DoubletapState_IneligibleForFirst,
-    DoubletapState_UsedSecond,
+    DoubletapState_Doubletap,
+    DoubletapState_Blocked,
 } doubletap_state_t;
 
 typedef struct {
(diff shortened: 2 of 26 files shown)


right/src/config_parser/parse_config.c
```diff
@@ -188,7 +187,7 @@ parser_error_t parseConfig(config_buffer_t *buffer)
 
     if (DataModelVersion.major >= 7) {
         secondaryRoles_Strategy = ReadUInt8(buffer);
-        secondaryRoles_AdvancedStrategyDoubletapTimeout = ReadUInt16(buffer);
+        uint16_t _ = ReadUInt16(buffer);
```
firngrod: Generates a warning, I know.  I could remove the warning, I just like the reminder that something isn't as it ought to be.

right/src/macros/commands.c
```diff
@@ -539,27 +539,7 @@ static bool processIfDoubletapCommand(bool negate)
...
-            S->ms.currentMacroStartTime - MacroState[i].ms.currentMacroStartTime <= Cfg.DoubletapTimeout &&
-            S->ms.currentMacroIndex == MacroState[i].ms.currentMacroIndex &&
-            &MacroState[i] != S
-        ) {
-            doubletapFound = true;
-        }
-    }
-
-    return doubletapFound != negate;
+    return S->ms.keyPress.isDoubletap != negate;
```
firngrod: Felt good.  The reason for this whole endeavor.

right/src/macros/core.c
```diff
@@ -433,8 +439,8 @@ uint8_t initMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset,
 
     S->ms.macroPlaying = true;
     S->ms.currentMacroIndex = index;
-    S->ms.currentMacroKey = keyState;
-    S->ms.currentMacroKeyStamp = timestamp;
+    S->ms.keyPress = keyPress != NULL ? *keyPress : (key_press_info_t){.keyState = NULL};
```
firngrod: I swear I handled the null case before I flashed it the first time and crashed on `$onInit`!

right/src/macros/core.c
```diff
@@ -433,8 +439,8 @@ uint8_t initMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset,
...
+    S->ms.keyActivationSeq = keyActivationSeq;
```
firngrod: Hmm, maybe the activation sequence should be part of the keyPress actually.

right/src/macros/set_command.c
```diff
@@ -323,8 +323,9 @@ static macro_variable_t secondaryRoleAdvanced(parser_context_t* ctx, set_command
         ASSIGN_BOOL(Cfg.SecondaryRoles_AdvancedStrategyDoubletapToPrimary);
     }
     else if (ConsumeToken(ctx, "doubletapTime")) {
+        Macros_ReportErrorPos(ctx, "Configuration variable has been merged with general doubletapTimeout, please switch to using that variable");
         DEFINE_INT_LIMITS(0, 65535);
-        ASSIGN_INT(Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout);
+        ASSIGN_INT(Cfg.DoubletapTimeout);
```
firngrod: I wasn't sure what would happen to the variable expansion if I didn't have some variable, so I did this.

right/src/key_history.h
```diff
@@ -0,0 +1,20 @@
...
+
+// Typedefs:
+
+// Variables:
+
+// Functions:
+
+void KeyHistory_RecordPress(const key_state_t *keyState);
+void KeyHistory_RecordRelease(const key_state_t *keyState);
+bool KeyHistory_WasLastDoubletap();
```
firngrod, line 18: Not much of a history yet, but I still chose to put it behind a compatible interface in case we wanted more in the future.

right/src/key_states.h
```diff
@@ -31,21 +31,28 @@
...
+        bool isDoubletap: 1;
+    } key_press_info_t;
+
 // Variables:
 
     extern key_state_t KeyStates[SLOT_COUNT][MAX_KEY_COUNT_PER_MODULE];
 
 // Inline functions
-
+    static inline bool KeyState_NoActivity(const key_state_t* s) { return (*((uint8_t*)s) + sizeof(s->debounceTimestamp)) == 0;}
```
firngrod: I feel that this function needs a better name, but putting it here, along with the comment above should prevent too much of a hassle in the future.

right/src/key_states.h
```diff
@@ -31,21 +31,28 @@
...
         uint8_t debounceTimestamp;
-        uint8_t activationTimestamp;
+        // Remember to update KeyState_NoActivity if adding or removing stuff before the bitfield
         volatile bool hardwareSwitchState : 1;
         bool debouncedSwitchState : 1;
         bool current : 1;
         bool previous : 1;
         bool debouncing : 1;
         secondary_role_state_t secondaryState : 2;
+        bool padding : 1; // This allows the KeyState_NoActivity() function to not trigger false because of sequence
```
firngrod: Doing this was just too tempting.

right/src/secondary_role_driver.c
```diff
@@ -204,7 +204,7 @@ static secondary_role_state_t resolveCurrentKeyRoleIfDontKnowTimeout()
 
     // see if we should set a timer to wake up to actively time out
     if (isActiveTimeout || isDoubletap) {
-        AWAITEVENT(Cfg.SecondaryRoles_AdvancedStrategyDoubletapTimeout);
+        AWAITEVENT(Cfg.SecondaryRoles_AdvancedStrategyTimeout);
```
firngrod, line 208: Oops!

right/src/usb_report_updater.c
```diff
@@ -727,8 +734,12 @@ static void updateActionStates() {
                 Trace_Printc("w4");
 
                 //apply active-layer action
-                ApplyKeyAction(keyState, cachedAction, actionBase, &NativeKeyboardReports);
+                ApplyKeyAction(&pressInfo, cachedAction, actionBase, &NativeKeyboardReports);
 
+                if(KeyState_DeactivatedNow(keyState)) {
+                    KeyHistory_RecordRelease(keyState);
+                    keyState->secondaryState = SecondaryRoleState_DontKnowYet;
```
firngrod, line 734: Couldn't resist.  Since we react to releases anyway, setting this along with the padding means we have removed the unneeded evaluations.

right/src/config_parser/parse_config.c
kareltucek: Isn't it? 

Please make it a `ATTR_UNUSED uint16_t secondaryRoles_AdvancedStrategyDoubletapTimeout = ...` so that we don't loose the information of what it used to be.

right/src/key_history.h
kareltucek: 👍

right/src/key_states.h
```diff
@@ -31,21 +31,28 @@
...
+        uint8_t activationSeq: 3;
     } key_state_t;
 
+    typedef struct {
+        key_state_t *keyState;
+        bool isDoubletap: 1;
+    } key_press_info_t;
```
kareltucek: I feel like `uint32_t activationTime` should become part of this structure.

right/src/macros/core.h
```diff
@@ -143,19 +143,19 @@
...
             uint16_t bufferOffset;
-            uint8_t currentMacroKeyStamp;
             uint8_t parentMacroSlot;
             uint8_t currentMacroIndex;
             uint8_t postponeNextNCommands;
             // ---- 4-aligned ----
             uint8_t nextSlot;
             uint8_t oneShot : 2;
             bool macroInterrupted : 1;
+            uint8_t keyActivationSeq: 3;
```
kareltucek: Shouldn't this rather become part of the key_press_info_t?

right/src/key_states.h
kareltucek: As well as the sequence number and secondary state.

right/src/key_states.h
kareltucek: Since we have already allocated another byte, shouldn't we move the secondary role bits there too?

kareltucek: Let me be super paranoid for a moment.

Macros can live and *ask* about secondary state after the key has already been released, in which case proactively clearing the secondaryState breaks them earlier then they need to. This is especially a problem if a key is activated as part of a secondary role resolution or activateKeyPostponed, because in such cases they are activated just for a couple cycles. Postponing should probably not allow the keys to be released so fast as to break the macros, but I am not sure if we guarantee that by specification.

Of course ideal solution is probably for macros to have their own secondaryState copy. Namely in the key_press_info_t.

right/src/macros/commands.c
kareltucek: 😊

right/src/macros/core.c
```diff
@@ -462,11 +468,17 @@ uint8_t initMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset,
...
 
 //partentMacroSlot == 255 means no parent
-uint8_t Macros_StartMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset, uint8_t timestamp, uint8_t parentMacroSlot, bool runFirstAction, const char *inlineText)
+uint8_t Macros_StartMacro(uint8_t index,
+                          const key_press_info_t *keyPress,
+                          uint16_t argumentOffset,
+                          uint8_t keyActivationsSeq,
+                          uint8_t parentMacroSlot,
+                          bool runFirstAction,
+                          const char *inlineText)
```
kareltucek: [philosophical nitpick] 

I have very mixed feelings about this kind of formatting. 

My scala/functional background screems loudly for uniform parenthesis formatting handling and says it should be:

```
uint8_t Macros_StartMacro(
    uint8_t index,
    const key_press_info_t *keyPress,
    uint16_t argumentOffset,
    uint8_t keyActivationsSeq,
    uint8_t parentMacroSlot,
    bool runFirstAction,
    const char *inlineText
) {
```

But I do recognize that this is not a C conventional style, and the clang format we provide isn't configured properly, so this is more of an opinion exchange than any request for any kind of reformat. 

If you have your opinion on this, you are welcome to share 😊. Otherwise this is a resolve-without-any-action case ;-).

right/src/macros/set_command.c
kareltucek: A cleaner alternative would be to create a dummy local variable: 

```
        Macros_ReportErrorPos(ctx, "Configuration variable has been merged with general doubletapTimeout, please switch to using that variable");
        uint32_t dummy = Cfg.DoubletapTimeout;
        DEFINE_INT_LIMITS(0, 65535);
        ASSIGN_INT(dummy);
```

right/src/key_history.c
```diff
@@ -0,0 +1,43 @@
...
+} previous_key_event_type_t;
+
+static previous_key_event_type_t lastPress;
+
+void KeyHistory_RecordPress(const key_state_t *keyState)
+{
+    bool isDoublePress = 
+        keyState == lastPress.keyState
+        && lastPress.doubletapState == DoubletapState_PossibleFirst
+        && CurrentPostponedTime < lastPress.timestamp + Cfg.DoubletapTimeout;
```
kareltucek, line 25: (Using CurrentPostponedTime as an activation time is correct, but gives me a headache every time I see it.)

right/src/usb_report_updater.c
```diff
@@ -677,19 +678,25 @@ static void updateActionStates() {
             key_action_cached_t *cachedAction;
             key_action_t *actionBase;
 
-            if(((uint8_t*)keyState)[2] == 0) {
+            if(KeyState_NoActivity(keyState)) {
```
kareltucek: Please use the dirty way here. Feel free to add a comment about it.

Reasoning: 

This loop is a performance-critical place.

If I am not mistaken, an inline function is still a function, which means that it introduces memory barriers which may hinder compiler optimizations.

right/src/usb_report_updater.c
```diff
@@ -677,19 +678,25 @@ static void updateActionStates() {
...
+                    .isDoubletap = false,
+                };
                 if (KeyState_ActivatedNow(keyState)) {
                     // cache action so that key's meaning remains the same as long
                     // as it is pressed
                     actionCache[slotId][keyId].modifierLayerMask = 0;
-                    keyState->secondaryState = SecondaryRoleState_DontKnowYet;
+                    ++keyState->activationSeq;
+                    KeyHistory_RecordPress(keyState);
+                    pressInfo.isDoubletap = KeyHistory_WasLastDoubletap();
```
kareltucek: (I find this place intuitively a bit surprising, but probably correct. I would expect these to be evaluated closer to commitKeystate)

It means that the isDoubletap that we pass around in any future evaluation is wrong, isn't it? 

The isDoubletap flag should be kept in the key_state_t structure I think. 

Passing the key_press_info_t around the usb report updater is clumsy, as we are loosing the activation time and the doubletap flag. I mean, it is correct to have them in there, but not to pass it around in this context. 

My idea was to use the key_press_info_t to pass the information through postponer and then from activatedNow to macro engine and keep it in macro state, but accept loosing the information for "normal actions", which would mean to *not* pass it around in the native action (usb report updater) call tree.

kareltucek (commented):
I have a good feeling about this refactor (thanks for it! I think it really benefits from your clean plate perspective). 

We could merge it as is I think, but at the same time, to get the most benefit from it, we should make one more iteration of refactors.

kareltucek:
If my perfectionistic nitpicks and requests are too much, please let me know. 

Even if merged as is, this definitely is a step forward.

right/src/macros/core.c
kareltucek: Yes please.

right/src/usb_report_updater.c
firngrod: I'll find a different kind of dirt.

firngrod:
No problem, I do this for fun, and obsessing over details is fun.

right/src/key_states.h
firngrod: Secondary state does indeed have a wonky nature when it comes to macros.  We cannot control when they start the resolution, but it certainly makes sense to store the result in macro scope instead of relying on the key state.  Once we do that, I see no problem with clearing on release.

right/src/macros/core.h
firngrod: It makes sense to store the key sequence in key_press_info, yes, but on the other hand, every place which needs it stores it on the press action anyway.

right/src/key_states.h
firngrod: Secondary state does not make sense on the press info.  It is only resolved later.  See my other comment on those.

right/src/macros/core.c
firngrod: I have no strong feelings other than lines should be short enough to fit in my editor, and once the line breaks, I want only one argument on each line.

firngrod:
More to come, putting kids to bed.

right/src/macros/core.h
kareltucek: I mean, some of those places that need it (macros) wouldn't have to store it separately.

right/src/usb_report_updater.c
firngrod: Evaluating it earlier and putting it through the postponer gains us nothing, I think.  Especially as the user can, at any time, modify the postponer queue, possibly requiring us to rewrite history.  I think this is the best place to evaluate doubletaps.

right/src/macros/core.h
firngrod: Yes, since macros may outlast the keystroke.

firngrod:
I lean in a very opposing direction from you on key state vs key press info, but the more I think about it, the more I'm certain that we should go all the way one way or all the way the other.

But before that, I actually think that secondary role state cache should be moved.  I think that secondary role driver should own the state, not key state.  Put 8 or 16 slots in there for caching resolutions and clear them on release.  It can be done pretty cleanly.  Macros should always manage their own memory of secondary roles.

Back to key state vs key event.  I think that if we keep key events, the key state should only contain information about the key state.  Basically what it contains now, minus secondary role.  Everything else should live in events.  Ideally, there should be no need to pass around key state pointers, key ID should be enough for the purpose of differentiating keys, nothing else.  Everything that the logic needs in order to execute should be encapsulated in the event type:  Is it a new press (ActivatedNow), new release (DeactivatedNow), is it doubletap, what was the ID of the press for macros to check back if that ID has been released, so on.  Macro engine is the only place where any of this information needs to live past the event, so it can store what it needs to store.  Realistically, nothing else needs the keyState really, does it?  To be clear, I don't want to get rid of passing key state around (yet), but it should theoretically be straightforward to do.

The alternative to this would be to drop the event idea and instead letting the places which need the information extract that information when needed.  This requires that these things are not changed during ApplyKeyAction, which I'm pretty sure is already true.  For example, as things are, there is no reason that Secondary Role driver and macro engine could not just call KeyHistory_WasLastDoubletap.

The upside of the former approach is that it's formalized and clear, but wastes resources on operations which are not always needed.  The upside of the latter is that it's leaner and more efficient, but the flow of information is less clear.

I'm ambivalent.  Now that I've thought about it, dropping the event type is probably my preferred way to go.  The macro engine needs to be tuned a bit for it, it needs to gather the information it needs on start, then pass it on to spin-offs.

The main thing about the event approach is that one can construct an event and inject it.  If Secondary Role calls WasLastDoubletap, I cannot easily execute something as a doubletap, say from a macro.  Now I lean that way again!  Argh!

firngrod:
I'm going to move secondary role state cache.  That one I really like.  I'll make it one commit which can be reverted if you don't like it.

firngrod:
Now that I've slept on it, I think I still like the idea of passing a key event around rather than the key state, for the following reasons:
- I like the idea of having access to things centralized.  Having one place where we determine "This is the event that we're going to process now!" makes it a lot easier to reason about as a human.
- Writing testable code for my job has given me a dislike of functions which depend on other things than what they are provided as parameters.
- I like the idea of processing events rather than starting a process which then inspects state.  It conceptually simplifies the mouse caret handling, for instance.

I think the additional processing to prepare this state is worth it, but I will let you be the judge.

kareltucek:
> But before that, I actually think that secondary role state cache should be moved. I think that secondary role driver should own the state, not key state. Put 8 or 16 slots in there for caching resolutions and clear them on release. It can be done pretty cleanly. Macros should always manage their own memory of secondary roles.

I like about the key_state_t solution that it guarantees a full nkey rollover for native actions. 

I do realize that it is more of an emotional fondness than an rational reasoning though. 

> Writing testable code for my job has given me a dislike of functions which depend on other things than what they are provided as parameters.

Well, I have been doing pure functional programming for my other jobs... but this is not a pure functional language and codebase.

> I like the idea of processing events rather than starting a process which then inspects state. It conceptually simplifies the mouse caret handling, for instance.
>
> I think the additional processing to prepare this state is worth it, but I will let you be the judge.

I think our ideas about what this refactor should be have been diverging in two directions which are not incompatible, just not one and the same thing (probably falling apart into two press info structures - one event-oriented for the native actions logic, and one state-oriented for macro states), so I am dropping my branch for another time and lets focus on your branch.

I am not sure what changes your idea would imply. Let's say that refactors that objectively improve state of affairs are welcome. On the other hand, if it is not broken, don't fix it. I don't think I am open to reconstructing the entire codebase just to get code that looks nicer.

I think I have stated my case and will look forward to whatever implementation you come up with.

right/src/secondary_role_driver.c
```diff
@@ -71,6 +71,41 @@ static bool currentlyResolving = false;
 
 static bool activateSecondaryImmediately;
 
+#define resolutionCacheSize 16
```
kareltucek: Please, macros always ("always") in upper case: RESOLUTION_CACHE_SIZE.

right/src/usb_report_updater.c
kareltucek: Good point.

kareltucek:
This is getting a bit messy (🙈), but I’ve read through all your comments and I don’t think I have any objections.

right/src/secondary_role_driver.c
firngrod: Sure, and I think I will also change the structure of the cache.  With a buffer size of 16, I think we can sacrifice 12 bytes to make it an array of structs to save on bitshifting.

kareltucek: Yes, I definitely prefer a 16 byte array for readability. 

I still lean towards the key_state_t storage because of the true-n-key-rollover-in-native-actions invariant. 

But I see you have a strong opinion on this, so feel free to go on with this.

firngrod:
> Well, I have been doing pure functional programming for my other jobs... but this is not a pure functional language and codebase.

I get it, and I'm not really advocating for making everything injectable, it's just a reasoning for why I lean that way.

> I don't think I am open to reconstructing the entire codebase just to get code that looks nicer.

Absolutely!  I don't want to overhaul the code, I just want to settle on which approach to take for the flow of information to the handling of an event.

> This is getting a bit messy (🙈), but I’ve read through all your comments and I don’t think I have any objections.

Agreed.  Let's clean this up to be only about the global doubletap handling.  I will make a minimally disruptive version of global double check, without the press event concept.  Regardless of where any discussion of the event concept goes, it will do so elsewhere.  This implementation is not good enough, so I will remove it.

I will also split the secondary role state thing to another PR where we can discuss the merits of it in isolation, and later, I will make a PR about macro initial state snapshot handling, I have some ideas in that regard.

mhantsch:
> macro initial state snapshot handling

Now you got me curious. Please keep me in the loop on this (when you are starting the PR).

firngrod:
It's just that when the macro starts, it should gather the information it might need during it's lifetime, but which might not be available later, such as doubletap, keytap id and so on.  It does so now, but not in a very formalized way.  I also want it to remember secondary role resolution and pass all of these things on to macros it might call.

firngrod:
It builds, haven't tested it, still needs code cleanup.

firngrod:
Also some of your comments are unaddressed.

right/src/macros/core.c
firngrod: I wasn't even consistent in my breaking of lines.

right/src/usb_report_updater.c
```diff
@@ -677,7 +676,7 @@ static void updateActionStates() {
             key_action_cached_t *cachedAction;
             key_action_t *actionBase;
 
-            if(((uint8_t*)keyState)[2] == 0) {
+            if(KEYSTATE_KEYINACTIVE(keyState)) {
```
firngrod, line 679: Told you I would find a different kind of dirty.

firngrod:
This now only has the global doubletap and a couple of, I think, objective improvements and/or fixes.

Since doubletap evaluation went on-site, I just made it use individual timeout values again.

By the way, you mentioned that you had a branch with a key event type implementation.  I would love to take a look at that when it's ready.

kareltucek:
> By the way, you mentioned that you had a branch with a key event type implementation. I would love to take a look at that when it's ready.

Have I? 

More context please?

I guess it is deprecated by this PR.

kareltucek:
> Since doubletap evaluation went on-site, I just made it use individual timeout values again.

I don't understand this part either.

What do you mean "went on-site"? 

Which one is the relevant commit?

firngrod:
> By the way, you mentioned that you had a branch with a key event type implementation. I would love to take a look at that when it's ready.
>> so I am dropping my branch for another time

Maybe you meant dropping your idea of a press event implementation?

> What do you mean "went on-site"?

I no longer call `KeyHistory_WasLastDoubletap` centrally and pass the result around, it is called where it's used.  That's what I mean by on-site.  It strikes me by the way that there is now a condition where secondary role evaluation in macros may be wrong about doubletap if there has been key action in the time from macro started until secondary role evaluation started.  I am personally willing to live with that, but I guess that it's not impossible that [someone](https://github.com/UltimateHackingKeyboard/firmware/issues/1453#issuecomment-3818060033) might make an `ifGesture` followed by `ifSecondary`.  *sigh*  I'll see if I can do something about that.  Maybe a key press info type might help.

firngrod:
Okay, now that we're not in any way bound by my previous implementation of the key press event concept, if I was to do it again, what do you think it should look like?
I still believe that the proper place to evaluate such a structure is when a keypress is being irrevocably applied, in `updateActionStates`.  It would have to make it's way down to macro engine, secondary role driver and layer switch logic at least, for the doubletap to be used there.  Is there a better way to do it, you think, than what I did last time, wrapping it around the keyState?

kareltucek:
> This now only has the global doubletap and a couple of, I think, objective improvements and/or fixes.
>
> Since doubletap evaluation went on-site, I just made it use individual timeout values again.

I think that unifying doubletap timeouts into a single one was a good thing.

> I still believe that the proper place to evaluate such a structure is when a keypress is being irrevocably applied, in updateActionStates.

Totally agreed.

> Is there a better way to do it, you think, than what I did last time, wrapping it around the keyState?

No, I think it is fine.

> Okay, now that we're not in any way bound by my previous implementation of the key press event concept, if I was to do it again, what do you think it should look like?

I believe you have better idea what to do and how to do it.

If I am to sum up my understanding of what has been going on:

------------------------------

My original idea was there to be a new struct to store the info in the macro state. It would probably look something like this:

```
struct macro_press_info_t {
    keyId or keystate pointer
    uint32_t activationTime
    ... secondaryRoleState
    ... isDoubletap
    ... seqId
}
```

But that is basically just gathering stuff that is already lying around and putting it at one more centralized place.

The concept which you had in mind was about passing things inside report updater, which kind of sounds similar, but turns out to be a very different thing.

--------------------

> I think that if we keep key events, the key state should only contain information about the key state. Basically what it contains now, minus secondary role. Everything else should live in events. 

> Ideally, there should be no need to pass around key state pointers, key ID should be enough for the purpose of differentiating keys, nothing else.

Sounds fine, except that there is a number of states that don't have keyid assigned - the navigation mode states for instance. These would have to be identified and dealt with.

> Everything that the logic needs in order to execute should be encapsulated in the event type: Is it a new press (ActivatedNow), new release (DeactivatedNow), is it doubletap, what was the ID of the press for macros to check back if that ID has been released, so on. Macro engine is the only place where any of this information needs to live past the event, so it can store what it needs to store. Realistically, nothing else needs the keyState really, does it? To be clear, I don't want to get rid of passing key state around (yet), but it should theoretically be straightforward to do.

If I understand correctly, you are suggesting to create a new structure similar to this and pass it around the native action processing (usb_report_updater):

```
struct key_press_info {
    keyId or keystate pointer; 
    eventType (activatedNow, active, deactivatedNow, nonActive);
    isDoubletap
    seqId
}
```

I would add: 
```
    uint32_t activationTime; // (I prefer to pass it around like this instead of those occasional cryptic readings of CurrentPostponedTime.)
```

-----------------

In any case, you seem to have a pretty good understanding of the keystate handling as well as an opinion on how it should be done. 

As for whether to do it or how or not, you be the judge. I trust the sanity of your jugdment and would prefer to not have to get too engaged into this.

firngrod:
I have re-deprecated secondary role specific doubletap timeout.

For the issue with macros doing secondary role evaluation after other key actions, I personally am fine with that for now for the following reasons:
- People are rarely going to do a bunch of stuff and then do an `ifPrimary/ifSecondary`
- In the case I referred to previously with `ifGesture` followed by `ifPrimary`, postpone should still be in effect from the `ifGesture`, right?
- In other cases, even if we handle doubletap correctly, if the key order is already broken by macro delays, it's broken, even if we patch doubletap specifically.  This is always going to be a fundamentally broken scenario
- It's only the doubletap within secondary role resolution which *may* be off.  It's a pretty specific case which is almost certainly not coming into play

I think this branch is now at a stage where it is ready for a final review and, hopefully, merge, without key press events.
I will look into a key event implementation in another PR where nothing else is on the line, both in terms of communication clarity and risk of loss if PR is rejected.  :slightly_smiling_face:

That might not be for at least a week as I will be visiting the in-laws for some days this coming week.

Thank you for your patience, and sorry for all the mess.

kareltucek:
I think everything said is correct.

> I think this branch is now at a stage where it is ready for a final review

Alright, will give it another look and probably merge soon!

> Thank you for your patience, and sorry for all the mess.

No, no, no, thank *you* for all the "mess". 😊

right/src/macros/core.c
```diff
@@ -434,10 +441,11 @@ uint8_t initMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset,
     S->ms.macroPlaying = true;
     S->ms.currentMacroIndex = index;
     S->ms.currentMacroKey = keyState;
-    S->ms.currentMacroKeyStamp = timestamp;
+    S->ms.keyActivationId = keyActivationId;
     S->ms.currentMacroStartTime = CurrentPostponedTime;
     S->ms.currentMacroArgumentOffset = argumentOffset;
     S->ms.parentMacroSlot = parentMacroSlot;
+    S->ms.isDoubletap = KeyHistory_WasLastDoubletap();
```
kareltucek, line 448: Some macros run without owning a key - those started from macroEvents or via usb, or those bound in navigation modes.

Iinm, this line can make them think they were a doubletap. It would be safer to ask for a "keystate" rather than "last key".

Of course, such macros shouldn't be asking for doubletaps in the first place, so this shouldn't be a problem in practice.

right/src/mouse_controller.c
```diff
@@ -247,17 +247,17 @@ static bool feedTapHoldStateMachine(touchpad_events_t events)
...
     }
     if (action & Action_Press) {
-        PostponerCore_TrackKeyEvent(singleTap, true, 0xff, Timer_GetCurrentTime());
+        PostponerCore_TrackKeyEvent(singleTap, true, 0xff);
     }
     if (action & Action_Doubletap) {
         // some systems do debouncing because mouse switches are unreliable
-        uint16_t delay = 20;
-        PostponerCore_TrackKeyEvent(singleTap, false, 0xff, Timer_GetCurrentTime());
+        const int16_t delay = 20;
```
kareltucek, line 257: Why signed? Is this a typo?

right/src/usb_report_updater.c
kareltucek: This is kinda same kind of dirty ;-).

right/src/key_states.h
```diff
@@ -31,13 +35,15 @@
...
-        uint8_t activationTimestamp;
+        // Remember to update KEY_INACTIVE() if adding or removing stuff before the bitfield
         volatile bool hardwareSwitchState : 1;
         bool debouncedSwitchState : 1;
         bool current : 1;
         bool previous : 1;
         bool debouncing : 1;
         secondary_role_state_t secondaryState : 2;
+        bool padding : 1; // This allows the KEY_INACTIVE() macro to not trigger false because of sequence
+        uint8_t activationId: 3;
```
kareltucek, line 46: I would feel safer with 4 bits.

kareltucek:
Let's merge this. 

If you want to solve any of the nitpicks, please do so in a new PR.

PR #1481, opened 2026-02-06
introduction of macroArg command

mhantsch:
`macroArg` declaration command, introduction of argument types, parsing of `$macroArg.`*`name`* and `$macroArg.`*`number`*, processing of `$macroArg` values for typed values of `int`, `float`, `bool`, `string`, `keyid` and (modded) `scancode` (=shortcut), in addition to backwards-compatible behaviour for `any` type.

`&macroArg` syntax is left unmodified by this PR.

The PR also contains some cleanup / refactoring (mainly in str_utils) but also a few other places to improve consistent code style and ensure consistent safety checks of context bounds.

User manual and reference docs have been updated as well.


right/src/macros/core.h
```diff
@@ -167,8 +173,11 @@
             bool autoRepeatInitialDelayPassed: 1;
             macro_autorepeat_state_t autoRepeatPhase: 1;
             // ---- 4-aligned ----
+            macro_arg_t arguments[MAX_MACRO_ARGUMENT_SIZE];
```
kareltucek: No. Definitely not allowing this into origin. 

As discussed, these should be allocated from a shared pool on demand.

right/src/macros/commands.c
```diff
@@ -2031,388 +2492,20 @@ static macro_result_t processCommand(parser_context_t* ctx)
...
-        case CommandId_activateKeyPostponed:
-            return processActivateKeyPostponedCommand(ctx);
-        case CommandId_autoRepeat:
-            return processAutoRepeatCommand(ctx);
-        case CommandId_addReg:
-            Macros_ReportErrorPos(ctx, "Command was removed, please use command similar to `setVar varName ($varName+1)`.");
-            return MacroResult_Finished;
+        // We assume the next command is a non-header command and will therefore
+        // finish the header block.
+        bool headersProcessed = true;
```
kareltucek: Does this have to live here? Why not manipulate the ms->... state directly?

right/src/str_utils.c
```diff
@@ -320,6 +320,11 @@ void ConsumeAnyIdentifier(parser_context_t* ctx)
     consumeWhite(ctx);
 }
 
+void ConsumeIdentifier(parser_context_t* ctx)
```
kareltucek: 😱

No. Don't publish this very unsafe thing as an Api. 

Also, it is never used.

right/src/macros/commands.c
```diff
@@ -2010,6 +2081,396 @@ static macro_result_t processZephyrCommand(parser_context_t* ctx) {
     } \
     break;
 
+static macro_result_t dispatchCommand(parser_context_t* ctx, command_id_t commandId, bool *headersFinished) {
```
kareltucek: Could we maybe leave the switch where it was? 

At the very minimum in order to make the diffs readable for now.

right/src/macros/core.h
mhantsch: Yes, understood. I'm not even filling the structure out currently. I will remove it for now until I have a better storage.

right/src/macros/commands.c
mhantsch: `processMacroArgCommand()` (or future other header commands) needs the state before its invocation. It will thrown a macro error if we're not in the header anymore. So I will only update the state after the command has been processed, but I need to know whether it was a header command or not. This information needs to flow back from dispatchCommand().

Alternatively, we could allocate another MacroResult bit to indicate that a header command was processed; I just didn't want to extend the enum beyond 8 bits. But maybe that's actually not an issue, and I should go that route. What do you think?

right/src/str_utils.c
mhantsch: Do all your Consume... interfaces move past White, or why do you consider this unsafe?

You are right, it's not being used. Must be a leftover from an earlier attempt. Will remove it.

right/src/macros/commands.c
mhantsch: I really wanted to pull out the long switch because it makes the logic in processCommand() more compact and obvious. I want to express the core of the command processing while loop, and which results trigger which next activities (some loop, some don't). I would really like to leave it in dispatchCommand(). It's a one-time diff, then it'll stay where it is.

right/src/str_utils.c
kareltucek: Whites now handle template expansions.

I am generally trying to keep context handling contained here in str_utils and isolate higher layers from having to deal with separators, whites, and generally tokenization. This api, including the ":" syntax, goes against this effort.

As I said, I strongly favor a 'macroArg int name "string literal"' format. (I am generally fond of the scala/kotlin 'name: type' style, but I feel we are here in a c-standard domain and so should stick to its conventions of 'type name'.)

right/src/macros/commands.c
kareltucek: Having this diff live on a long lived feature branch (that this is probably going to be) makes it unsafe - it will tend to create conflicts that are not easily detectable and resolvable. 

If we are to do this refactor (I still don't see it as important or necessary), then we need to do it via a dedicated PR.

right/src/macros/commands.c
mhantsch: Found a different solution that needs no additional bits. See `MacroResult_Header`.

right/src/str_utils.c
mhantsch: I do not want "type name". That's old school. Modern languages use "name first, followed by type" order.

My view is that the types are additional, optional properties of the parameters. The ':' notation is more towards the style of e.g. Rust. It is optional; `macroArg the_param My beautiful param` will map as `any` type.

Readability: all the names are at the same position, if you have the type first the indentation will vary.
```
macroArg this_argument:int Number of customers
macroArg the_other_argument:string Country name
```

Dynamic typing or type inference: leave out the optional `:type` for an `any` type.

Type theory: Type is a property of a value, name-first mimics this.

My stream of thoughts:
`macroArg x:int Initial counter`
- This macro has an argument
- It is called 'x'
- btw, it is an integer
- Fill it out with the initial counter value

We may need to agree to disagree here.

right/src/macros/commands.c
mhantsch: I can create a mini-refactor just for this, if that's your preference. I understand you don't want this block of change to live for weeks or even months of diffs; that makes sense.

right/src/str_utils.c
kareltucek: Alright.

But please really enclose those comments in quotes:

```
macroArg this_argument:int "Number of customers"
macroArg the_other_argument:string "Country name"
```

Unquoted strings are allowed only for backward compatibility.

mhantsch: OK, I will make the description a proper string.

mhantsch: OK, I will make the description a proper string.

mhantsch: I have loosened the parser a bit so that it will accept two variants:

```
macroArg this_argument:int "Some number"
macroArg that_argument: int "Some other number"
```

And of course, without type (meaning `any`):

```
macroArg another_argument "Anything really"
```

right/src/macros/commands.c
```diff
@@ -1998,13 +2090,404 @@ static macro_result_t processZephyrCommand(parser_context_t* ctx) {
...
+}
+
 static macro_result_t processCommand(parser_context_t* ctx)
 {
     const char* cmdTokEnd = TokEnd(ctx->at, ctx->end);
     if (cmdTokEnd > ctx->at && cmdTokEnd[-1] == ':') {
         //skip labels
         ConsumeAnyToken(ctx);
-        if (ctx->at == ctx->end && IsEnd(ctx)) {
+        // TODO: why not just IsEnd() in this condition?
```
kareltucek, line 2487: Much faster.

right/src/macros/display.c
```diff
@@ -211,6 +218,7 @@ void processList(parser_context_t* ctx, bool show, uint16_t time) {
 
 macro_result_t Macros_ProcessSetLedTxtCommand(parser_context_t* ctx)
 {
+    // TODO: I guess ATTR_UNUSED is not correct here?
```
kareltucek, line 221: Indeed probably no longer needed.

right/src/macros/keyid_parser.c
```diff
@@ -166,12 +166,15 @@ static const lookup_record_t* lookup(uint8_t begin, uint8_t end, const char* str
...
 
+// this attempts to consume a KEYID_ABBREV (textual key id)).
+// returns the key id, or 255 if not found (invalid key id).
 uint8_t MacroKeyIdParser_TryConsumeKeyId(parser_context_t* ctx)
 {
     // this gets identifier till the next dot only
     const char* end1 = IdentifierEnd(ctx);
     const lookup_record_t* record = lookup(0, lookup_size-1, ctx->at, end1);
 
+    // TODO: WHY this second attempt with dot??
```
kareltucek, line 177: Because there are abbreviations like `rightModule.leftButton`.

right/src/macros/vars.c
```diff
@@ -198,7 +271,8 @@ static macro_variable_t* consumeVarAndAllocate(parser_context_t* ctx)
         }
     }
 
-    CTX_COPY(bakCtx, *ctx);
+    // TODO: Is this needed at all? Looks like something left over
```
kareltucek, line 274: Probably left over, yes.

right/src/macros/vars.c
```diff
@@ -236,6 +310,14 @@ static macro_variable_t* consumeVarAndAllocate(parser_context_t* ctx)
     return res;
 }
 
+static macro_variable_t reportUnexpectedVariableType(parser_context_t* ctx, macro_variable_type_t type)
+{
+    // TODO: would there be any way to trace the current position down the 
+    //       context stack and report it here?
+    Macros_ReportErrorNum("Unexpected variable type:", type, NULL);
```
kareltucek, line 317: You mean the expansion stack? That should be reported automatically.

kareltucek (commented):
Just a quick peek and some replies to some questions. 

I am generally missing documentation, but we can talk about all that I guess.

right/src/macros/vars.c
kareltucek: I mean, if the context is provided. Migrate it to a more modern Macros_Report.* format that takes ctx pointer.

right/src/macros/string_reader.c
```diff
@@ -12,12 +12,287 @@
...
 #endif
 
-typedef enum {
-   StringType_Raw,
-   StringType_DoubleQuote,
-   StringType_SingleQuote,
-} string_type_t;
+// new code:
 
+static char consumeExpressionCharOfInt(const macro_variable_t* variable, uint16_t* idx);
```
kareltucek, line 17: Do I understand correctly that you have here just copied the code so that you have an original reference for coding?

mhantsch:
This is now updated with all the changes up to firmware 17.1.1.

Issue #1406, opened 2025-12-15
[sleep-wake] Slow scrolling after wake up on WINDOWS


jscheidtmann:
As requested in #1155 I am opening an issue for WINDOWS specifically. 

Experienced behavior:
- After Windows 11 resumes from sleep, scrolling with the keycluster module is awfully slow. 
- This happens to me both when connected directly to the computer and when connected via an USB-hub.
- Plugging the keyboard out and in again resets the scroll speed.

Expected behavior:
- scroll speed should be unaffected by the OS going to sleep and awakening.

My configuration: 
- UHK60, Agent 8.0.0, Firmware 15.0.1 on all components (UHK60, Key Cluster, Trackball)

I will report back when I have updated to 8.0.1 and the respective firmware is installed.

daviian:
My configuration:
- UHK60, Agent 8.0.1, Firmware 15.3.0 on all components

Unfortunately, the issue still persists. However I cannot confirm that it is tied to sleep / awakening.
I noticed it simply after a while. Disconnect/reconnect the keyboard fixes it temporarily.

dustyculler:
I shut my Windows 11 desktop down at the end of the day and each morning I have this problem until I use the key combo that I setup to run the reboot command.  My UHK is plugged directly into a usb port on the motherboard.

My configuration:

- UHK60 v2, Agent 9.0.0, Firmware 16.0.0 on all components (UHK60 v2, Key Cluster, Trackball)

This has been occurring for a few months and while I don't know which firmware version this started on, I recall that agent had me update due to a new configuration format before it would let me make changes, and I first noticed this issue after that update a few months ago.  I don't typically install every update so I was likely a few versions back when that update I first experienced this issue on was applied.

RedDirtDog:
I'm on Win10 - do you want my stuff in here or in #1155 ?

UHK60, Trackball & Key cluster
Just upgraded from R11 (in steps thru each major release) to R16.1.1 & Agent 9.0.1

Slow scrolling occurs regardless of waking up - after reboot/replug, scrolling is fine for a short while then dies. Keyboard was connected via Dell dock, am now trying directly connected to laptop (USB-C).

kareltucek:
Here, since it is a windows case.

jscheidtmann:
[...]

Issue #1169, opened 2025-03-07
Failed to allocate net_buf


kareltucek:
From https://github.com/UltimateHackingKeyboard/firmware/issues/1124

> I got this in the UART adapter log today:
> 
> ```
> [2025-03-06 20:16:43] E: Failed to allocate net_buf 1026
> [2025-03-06 20:16:48] E: Failed to allocate net_buf 1026
> [2025-03-06 20:17:03] E: Failed to allocate net_buf 1026
> [2025-03-06 20:17:08] E: Failed to allocate net_buf 1026
> [2025-03-06 20:17:53] E: Failed to allocate net_buf 1026
> [2025-03-06 20:17:58] E: Failed to allocate net_buf 1026
> [2025-03-06 20:18:23] E: Failed to allocate net_buf 1026
> [2025-03-06 20:18:28] E: Failed to allocate net_buf 1026
> ```
> 
> When it started, I think I was just typing and spamming the backspace key a few times. I didn't see any OLED messages, or notice any abnormal behavior while it was happening.
> 
> **FW v12.4.0 #efe3a64 Bridge cable & USB both halves to PC Dongle was plugged into PC, but disabled via** `set bluetooth.enabled 0`.

Grepping the filesystem leads to usb: 
```
./zephyr/drivers/usb/udc/udc_common.c:          LOG_ERR("Failed to allocate net_buf %zd", size);
./zephyr/subsys/usb/device_next/class/usbd_hid.c:               LOG_ERR("Failed to allocate net_buf");
./zephyr/subsys/usb/device_next/class/bt_hci.c:                 LOG_ERR("Failed to allocate net_buf");
```

Not sure how severe that is, or how much that is in c2usb's domain, but thought I should notify you @benedekkupper .

According to @pcooke9, there was no abnormal behavior, so it may be safe to ignore?

kareltucek:
I guess this may no longer be relevant.

pcooke9:
I forgot all about this, but it still happens. Idk what these log messages mean, but I don't notice any abnormal behavior otherwise. I'll leave it to you guys if you think it can be ignored, but I figured I'd note this here in case it's helpful later or whatever...

[...]

PR #1482, opened 2026-02-07
Uhk60 c2usb integration

benedekkupper:
major refactoring of UHK60 right USB interface

* start using Kconfig for device configuration values (see right/Kconfig.device)
* use generated usb_device_config.h header
* organize HID report manipulation into self-contained sources and headers
* remove kusb interfacing code, except for buspal
* update c2usb dependency handling
* merge zephyr c2usb interface with UHK60
* use single report for system and media usages
* remove unused gamepad interface

This PR is work in progress, the following tasks are still ahead:
- [x] initial design review of the changes @kareltucek
- [x] verification of UHK60 functionality @kareltucek @mondalaci
- [x] fix UHK80 build @benedekkupper
- [x] fix UHK60 crashes @benedekkupper 
- [x] ~investigate~ try to reproduce uhk80 hogp crashes @kareltucek
- [x] look into dongle slow scrolling @kareltucek 
- [x] investigate and fix UHK80 agent flashing @benedekkupper @kareltucek
- [x] create zephyr RTOS design for throttling report sending in `usb_report_updater` @kareltucek @benedekkupper
- [x] merge Kconfigs of UHK60 and UHK80 @benedekkupper
- [x] https://github.com/UltimateHackingKeyboard/firmware/pull/1482#issuecomment-4220822193 - April 7th regressions reported by Phil: @benedekkupper
- [x] https://github.com/UltimateHackingKeyboard/firmware/pull/1482#issuecomment-4199749417 - hogp failures after pairing - @benedekkupper, @kareltucek (if you need assistence, logs, debugging...)
- [x] test both UHK60 and UHK80 extensively @benedekkupper @kareltucek @mondalaci


mondalaci:
@benedekkupper Please let me know when I should test a CI build.

right/src/macros/scancode_commands.c
```diff
@@ -22,20 +18,14 @@ typedef struct {
 
 static void addBasicScancode(uint8_t scancode, macro_usb_keyboard_reports_t* reports)
 {
-    if (!scancode) {
```
kareltucek, line 25: I am wondering if it is safe to remove these...

right/src/event_scheduler.h
```diff
@@ -71,7 +71,7 @@
        // some other minor triggers
        EventVector_KeyboardLedState =                      1 << 10,
        EventVector_UsbMacroCommandWaitingForExecution =    1 << 11,
-       EventVector_ProtocolChanged =                       1 << 12,
+       // EventVector_ProtocolChanged =                       1 << 12, // unused
```
kareltucek, line 75: Where is the protocol change (nkro vs 6kro) now handled?

kareltucek (commented):
As for initial review: looks great. Thanks for all those refactors - they are spot on. I didn't expect you to go this deep into our side of code.

Are there any areas that should I focus on specifically with review? 

I mean, are there any notable changes in logic, or potentially controversial areas? (I have tried to read through all the changes, but it is easy to miss things among minor changes, mass renames, etc....)

kareltucek:
> create zephyr RTOS design for throttling report sending in usb_report_updater

What exactly do you have in mind?

What comes to mind given this prompt:
- It is fine if the send function blocks for the next couple transport windows. Considering hogp, this means it is fine if you block the main thread for 20ms or so. If you need longer, an error should be returned to trigger resending.
- If you give us a function which tells us whether c2usb is `readyForNextReport`, we can use it and wait with report calculation. With current implementation, this would mean that we poll you once every millisecond. As for alternatives:
  - If you also tell us for how long to sleep, we can use that information in sleep calculation.
  - Proper solution would be to wait via a semaphore until we receive a callback, but I don't think this is worth the code complexity.
  - (The simplest polling solution is preferred.)

Whether and how the reports should be queued - I assume a queue of length 1 - the switching of active and "inactive" report buffers:
- In case of usb transport, I think it makes sense for one report to be waiting in the usb stack for transport while another is already being constructed. I.e., "ready" function should return true once there is a free spot in the queue.
- In hogp case, due to high latency, optimal would be to calculate the time when a report should be constructed with respect to the next transport window. The simple version would be to calculate next report once the previous report has been sent. In any case, this can be encapsulated in the suggested "readyForNextReport" function. 

Assume it takes 1ms to construct a report. (Well, I should make some statistics of this.)

kareltucek:
> verification of UHK60 functionality @kareltucek @mondalaci

UHK60 crashes somewhere in initUsb.

What is the state of testing on your end @benedekkupper . Can you efficiently test and debug with actual UHK60 hardware, or is that up to me?

right/src/event_scheduler.h
kareltucek: Although, I think it is correct that the usb_report_udpater logic shouldn't be needed these days, as everything is stored in nkro format, so indeed this should be a no issue.

kareltucek:
I have pushed a minor PID fix of west_agent.py.

device/src/usb/app_base.cpp
```diff
@@ -1,40 +0,0 @@
...
-void app_base::send(const std::span<const uint8_t> &buffer)
-{
-    if (!active()) {
-        return;
-    }
-    if (!sending_sem_.try_acquire_for(SEMAPHORE_RESET_TIMEOUT)) {
-        //return;
-    }
-    std::copy(buffer.begin(), buffer.end(), in_buffer_.data() + sizeof(in_id_));
-    auto result = send_report(in_buffer_);
```
kareltucek, line 17: Have you been dropping all errors here all along?

right/src/macros/scancode_commands.c
benedekkupper: It is safe to remove them, because the call chain checks for zero values too. It does mean that zero values are processed longer, so if you want to speed up the error handling, I can restore these checks.

right/src/event_scheduler.h
benedekkupper: Now the UHK60 handles this internally, just like the UHK80 already does.

benedekkupper:
> Are there any areas that should I focus on specifically with review?
>
> I mean, are there any notable changes in logic, or potentially controversial areas? (I have tried to read through all the changes, but it is easy to miss things among minor changes, mass renames, etc....)

The key areas are the `usb_report_updater.c` (as always) and `hid/transport.cpp`, I'm not entirely confident about the changes affecting the keyboard LED states, and the `UsbReportUpdateSemaphore`, so please take a closer look at those.

> > create zephyr RTOS design for throttling report sending in usb_report_updater
> 
> What exactly do you have in mind?
> 
> What comes to mind given this prompt:
> 
>     * It is fine if the send function blocks for the next couple transport windows. Considering hogp, this means it is fine if you block the main thread for 20ms or so. If you need longer, an error should be returned to trigger resending.
> 
>     * If you give us a function which tells us whether c2usb is `readyForNextReport`, we can use it and wait with report calculation. With current implementation, this would mean that we poll you once every millisecond. As for alternatives:
>       
>       * If you also tell us for how long to sleep, we can use that information in sleep calculation.
>       * Proper solution would be to wait via a semaphore until we receive a callback, but I don't think this is worth the code complexity.
>       * (The simplest polling solution is preferred.)

Now it will return `-EBUSY` when the previous report wasn't sent out yet. `Hid_KeyboardReportSentCallback` and co. with `HID_TRANSPORT_BLE` will be called when the report was sent (that's your `readyForNextReport`), so you can pick up from there.


> Whether and how the reports should be queued - I assume a queue of length 1 - the switching of active and "inactive" report buffers:
> 
>     * In case of usb transport, I think it makes sense for one report to be waiting in the usb stack for transport while another is already being constructed. I.e., "ready" function should return true once there is a free spot in the queue.

This is doable, but it needs to split handling USB and BLE report sending, so I'd do it in another round.

>     * In hogp case, due to high latency, optimal would be to calculate the time when a report should be constructed with respect to the next transport window. The simple version would be to calculate next report once the previous report has been sent. In any case, this can be encapsulated in the suggested "readyForNextReport" function.
> 
> 
> Assume it takes 1ms to construct a report. (Well, I should make some statistics of this.)

Don't design too much around this high latency, there is a new BLE spec version already supported by nRF Connect SDK, that can drastically reduce this ([LE Shorter Connection Intervals](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/releases_and_maturity/releases/release-notes-3.2.0.html))

> UHK60 crashes somewhere in initUsb.
>
> What is the state of testing on your end @benedekkupper . Can you efficiently test and debug with actual UHK60 hardware, or is that up to me?

I have debugged the firmware a bit, but not in a production environment, I flashed the firmware without bootloader present. The keys and mouse movement is working in general, I didn't test further so far due to the limited time availability. Do you have a JTAG/SWD debugger attached, or you use some kind of log and trace functionality? Did yo flash the CI build that crashed, or your local build? Make sure you run the usual west command sequence before build: `west config manifest.file west_mcuxsdk.yml && west update && west patch`

kareltucek:
> Do you have a JTAG/SWD debugger attached, or you use some kind of log and trace functionality?

I tried to attach to a gdb via an swd, but failed. Can look into it deeper on monday I think.

It was a local build.

kareltucek:
As for those crashes:

- when deploying from vscode without bootloader, everything works fine for both release and debug builds. 
- when I flash the regular way - via bootloader and connect to vscode gdb and let the code continue, I get a crash (again with both (export) DEBUG=0 and DEBUG=1 firmwares):

```
bt
#0  HardFault_Handler () at /opt/west2/firmware/right/src/trace_reasons.c:123
#1  <signal handler called>
#2  0x00034452 in usb::df::function::deinit(usb::df::config::interface const&) ()
#3  0x000323e8 in usb::df::device::set_config(usb::df::config::view, usb::df::device::event) ()
#4  0x2000ee40 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

info registers
r0             0x880bd301          2282476289
r1             0x780               1920
r2             0x787               1927
r3             0x880bd301          2282476289
r4             0x15881             88193
r5             0x1                 1
r6             0x0                 0
r7             0x2000ee00          536931840
r8             0x0                 0
r9             0x1fff1c60          536812640
r10            0x1                 1
r11            0x0                 0
r12            0x0                 0
sp             0x2000ede0          0x2000ede0
lr             0xfffffff1          -15
pc             0x2b7d8             0x2b7d8 <HardFault_Handler>
xpsr           0x10f0003           17760259
fpscr          0x60000010          1610612752
msp            0x2000ede0          0x2000ede0
psp            0x0                 0x0
primask        0x0                 0
basepri        0x0                 0
faultmask      0x0                 0
control        0x0                 0
apsr           0xf0000             983040
epsr           0x1000000           16777216
ipsr           0x3                 3
iapsr          0xf0003             983043
eapsr          0x10f0000           17760256
iepsr          0x1000003           16777219

x/x 0xE000ED28
0xe000ed28:	0x00008200
x/x 0xE000ED2C
0xe000ed2c:	0x40000000
x/x 0xE000ED34
0xe000ed34:	0x880bd301
x/x 0xE000ED38
0xe000ed38:	0x880bd301
```

kareltucek:
> create zephyr RTOS design for throttling report sending in usb_report_updater

Is there any specific concern regarding this? It seems to me that the UsbReportUpdateSemaphore takes care of things quite optimally for now, and so can be checked off.

-----------------

I went through the usb_report_updater and hid/transport.cpp, and I don't see any obvious issues. 

I am running this firmware on uhk80 and it works like a charm so far.

right/src/hid/transport_ble.cpp
```diff
@@ -60,3 +62,31 @@ extern "C" void HOGP_Disable()
 {
     hogp_manager::instance().main_service().stop();
 }
+
+extern "C" int HOGP_HealthCheck()
```
kareltucek, line 66: Is there anything more we can do to check health of the connection? 

(I do realize that this is a bit redundant since we should get failures on report sends.)

benedekkupper: That's as much as I can think of

kareltucek:
I have lost the ability to flash uhk80 via Agent with this PR.

benedekkupper:
> I have lost the ability to flash uhk80 via Agent with this PR.

What error message are you getting?

kareltucek:
It wasn't able to find the bootloader.

kareltucek:
(I mean the Agent.)

benedekkupper:
Is the USB HID index of the agent command interface still hardcoded in the agent?

kareltucek:
No iirc, but I have no hard data for it.

benedekkupper:
@kareltucek I can reproduce the crash, but I'm unable to debug it, as the crash only happens when the bootloader starts the application, and not the debugger. Can you provide some further information about the crash site? Maybe a wireshark capture of the USB communication up to the point of the crash?

kareltucek:
Probably not today I am afraid.

So the problem is that you don't know how to reproduce it with a debugger attached, am I correct? 

I used the agent flashing procedure with a local build. These commands in my case:

```
export DEBUG=1
./build.sh switchMcux
./build.sh right build flashUsb
```

Which translates to: 
```
west build --build-dir "$BUILD_DIR" "$DEVICE_DIR" --pristine -- --preset "$VARIANT"
west agent --build-dir $BUILD_DIR
```

I used this launch.json: 

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "J-Link Debug",
            "type": "cortex-debug",
            "request": "launch",
            "serverpath": "/opt/SEGGER/JLink/JLinkGDBServer",
            "servertype": "jlink",
            "device": "MK22FN512VLH12",
            "interface": "swd",
            "executable": "${workspaceFolder}/right/build/v2-debug/uhk-60-right.elf",
            "loadFiles": ["${workspaceFolder}/right/build/v2-debug/uhk-60-right.elf"],
            "gdbPath": "/opt/arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi/bin/arm-none-eabi-gdb",
            "svdFile": "",
            "runToEntryPoint": "main",
            "overrideLaunchCommands": [
                "monitor halt",
                "monitor reset",
                "load",
                "set $sp = *(0xC000)",
                "set $pc = *(0xC004)"
            ],
            "overrideRestartCommands": [
                "monitor halt",
                "monitor reset",
                "load",
                "set $sp = *(0xC000)",
                "set $pc = *(0xC004)"
            ]
        }
    ]
}
```

I don't remember the exact procedure though. But should be straightforward.

benedekkupper:
@kareltucek How is the suspend and resume supposed to be signalled to the firmware? I'm calling `PowerMode_SetUsbAwake`, but this looks like a dummy endpoint.

kareltucek:
It is the correct way. Or the intended way anyways. I just might have disabled it due to past issues with suspend detection 🤔. Also, for uhk80, we need to be awake when any host is active, which makes that api a bit inconvenient as the information source.

What is the intended transport behavior during suspends? I mean, there is that signaling of set connection state. I would expect transport to not be available during suspend, but I am not sure about the semantics.

benedekkupper:
> It is the correct way. Or the intended way anyways. I just might have disabled it due to past issues with suspend detection 🤔. Also, for uhk80, we need to be awake when any host is active, which makes that api a bit inconvenient as the information source.

I had to use different calls to make it work as before, something to solve at some point.

> What is the intended transport behavior during suspends? I mean, there is that signaling of set connection state. I would expect transport to not be available during suspend, but I am not sure about the semantics.

The transport isn't available for sending during suspend. Preparing to receive is still allowed, but nothing will be delivered until the bus is resumed.

benedekkupper:
The crashes are fixed, please retest @kareltucek

kareltucek:
I have used uhk60 and uhk80 on a couple of occasions, and they work fine except for occasional lockups. 

I will try to debug these more extensively (wireshark, gdb) when I return from vacation towards the end of March.

benedekkupper:
> I have used uhk60 and uhk80 on a couple of occasions, and they work fine except for occasional lockups.

I also had lockups (e.g. mouse not sending new reports after a while of movement) in previous versions of the PR, they have been fixed in the last two builds. @mondalaci if you have a chance, you can try the latest build.

mondalaci:
@benedekkupper Should I test https://github.com/UltimateHackingKeyboard/firmware/actions/runs/22808403305 ?

benedekkupper:
> @benedekkupper Should I test https://github.com/UltimateHackingKeyboard/firmware/actions/runs/22808403305 ?

Yes

mondalaci:
Just started testing. So far so good!

mondalaci:
I've been using the referred release for three days, and it has been rock stable.

@kareltucek Have you also been testing it? What's your experience?

@benedekkupper Unlike on my UHK 80, I can wake up my PC by pressing any key on my UHK 60 v2 with this firmware, even though both UHKs are now controlled by c2usb. How so?

pcooke9:
A couple of issues with FW v16.2.0 #801794e on UHK80:
Both halves connected via USB, and bridge cable
Touchpad module
Win 11

(The issues below are reproducible using default config. Neither issues happen with official FW v16.2.0 or latest master branch v16.2.0-146b61b. Also no difference with/without the bridge cable.)


1.  Mouse keys via dongle - both movement and scrolling is very slow/sporadic, and _occasionally_ (rarely) the touchpad movement seems affected as well. Tested with a couple of dongles. Everything else seems to work fine over dongle though (typing output, macros, etc.).

-  Resetting/rebooting the UHK80 doesn't help. Doesn't seem related to Windows fast startup like mentioned in other issues, as I don't have fast startup enabled, and rebooting the UHK80 doesn't resolve it.

-  In case it's helpful:
If I use the dongle with firmware on official v16.2.0, but the UHK80 on FW v16.2.0 #801794e, it seems to work fine; maybe a little choppier than usual, but not slow.

2.  When trying to pair a device via BLE, it consistently fails upon typing in the third digit of the pairing code. The OLED screen shows that the output of every keypress is doubled in the pairing code prompt, and tapping backspace deletes two numbers at a time as well.

-  This happens regardless of using the UHK80's bluetooth pairing mode, or when advertising normally (happens 100% of the time). Double output is only experienced when entering the pairing code, not during normal typing, etc..

-  As a result, I am unable to test BLE with this FW.
___

UHK60 seems stable for me so far. 😎

pcooke9:
Slow mouse keys issue is also present when using a dongle on my TV (Samsung Tizen OS), as well as on my Android phone. So that confirms it's not just a Windows OS issue.

kareltucek:
I have fixed the pairing screen issue. However, shortly after pairing or connecting ble hid device, the keyboard freezes. Stopping it yields this callstack:

```
z_SysNmiOnReset() (/opt/west3/zephyr/arch/arm/core/nmi_on_reset.S:27)
z_arm_nmi() (/opt/west3/zephyr/arch/arm/core/nmi.c:58)
hid::application::setup(hid::protocol prot, hid::transport * tp, hid::application * const this) (/opt/west3/c2usb/c2usb/hid/application.hpp:194)
bluetooth::zephyr::hid::service::start_app(bluetooth::zephyr::hid::service * const this, bt_conn * conn, hid::protocol protocol) (/opt/west3/c2usb/c2usb/port/zephyr/bluetooth/hid.cpp:546)
bluetooth::zephyr::hid::service::ccc_cfg_write(bt_conn * conn, const bluetooth::zephyr::gatt::attribute * attr, bluetooth::zephyr::gatt::ccc_flags flags) (/opt/west3/c2usb/c2usb/port/zephyr/bluetooth/hid.cpp:475)
bt_gatt_attr_write_ccc(struct bt_conn * conn, const struct bt_gatt_attr * attr, const void * buf, uint16_t len, uint16_t offset, uint8_t flags) (/opt/west3/zephyr/subsys/bluetooth/host/gatt.c:2243)
write_cb(const struct bt_gatt_attr * attr, uint16_t handle, void * user_data) (/opt/west3/zephyr/subsys/bluetooth/host/att.c:2081)
gatt_foreach_iter(uint16_t * num_matches, uint16_t end_handle, uint16_t handle, const struct bt_gatt_attr * attr) (/opt/west3/zephyr/subsys/bluetooth/host/gatt.c:2032)
foreach_attr_type_dyndb(void * user_data, bt_gatt_attr_func_t func, uint16_t num_matches, const void * attr_data, const struct bt_uuid * uuid, uint16_t end_handle, uint16_t start_handle) (/opt/west3/zephyr/subsys/bluetooth/host/gatt.c:2064)
bt_gatt_foreach_attr_type(uint16_t start_handle, uint16_t end_handle, const struct bt_uuid * uuid, const void * attr_data, uint16_t num_matches, bt_gatt_attr_func_t func, void * user_data) (/opt/west3/zephyr/subsys/bluetooth/host/gatt.c:2113)
bt_gatt_foreach_attr(void * user_data, uint16_t end_handle, uint16_t start_handle) (/opt/west3/zephyr/include/zephyr/bluetooth/gatt.h:663)
att_write_rsp(struct bt_att_chan * chan, uint8_t req, uint8_t rsp, uint16_t handle, uint16_t offset, const void * value, uint16_t len) (/opt/west3/zephyr/subsys/bluetooth/host/att.c:2128)
att_write_req(struct bt_att_chan * chan, struct net_buf * buf) (/opt/west3/zephyr/subsys/bluetooth/host/att.c:2155)
bt_att_recv(struct bt_l2cap_chan * chan, struct net_buf * buf) (/opt/west3/zephyr/subsys/bluetooth/host/att.c:2992)
l2cap_chan_recv(_Bool complete, struct net_buf * buf, struct bt_l2cap_chan * chan) (/opt/west3/zephyr/subsys/bluetooth/host/l2cap.c:2710)
bt_l2cap_recv(struct bt_conn * conn, struct net_buf * buf, _Bool complete) (/opt/west3/zephyr/subsys/bluetooth/host/l2cap.c:2744)
bt_acl_recv(uint8_t flags, struct net_buf * buf, struct bt_conn * conn) (/opt/west3/zephyr/subsys/bluetooth/host/conn.c:476)
hci_acl(struct net_buf * buf) (/opt/west3/zephyr/subsys/bluetooth/host/hci_core.c:656)
rx_work_handler(struct k_work * work) (/opt/west3/zephyr/subsys/bluetooth/host/hci_core.c:4246)
work_queue_main(void * workq_ptr, void * p2, void * p3) (/opt/west3/zephyr/kernel/work.c:688)
z_thread_entry(k_thread_entry_t entry, void * p1, void * p2, void * p3) (/opt/west3/zephyr/lib/os/thread_entry.c:48)
```

kareltucek:
> @kareltucek Have you also been testing it? What's your experience?

I have been actively using uhk60 with this firmware and it is rock stable. 

My office uhk80 is unusable in practice due to those busted matrix rows. Will probably bring another unit, but didn't use it in practice. Given above hogp crashes, I consider uhk80 unstable with this firmware.

benedekkupper:
@kareltucek about the dongle mouse issue, I see that in messenger.c there is no retry mechanism for calling the `Hid_Send*Report` methods, do you have an idea what could be done there?

@pcooke9 do you have a chance to retest the #1406 problem with this build? Do you see any difference on UHK60 or UHK80?

I will check the BLE HID connection crash issue earliest Sunday, latest next weekend.

pcooke9:
> @pcooke9 do you have a chance to retest the #1406 problem with this build? Do you see any difference on UHK60 or UHK80?

The following results are with firmware v16.2.0 #801794e, or should I retest with a newer FW from the uhk60-c2usb-integration branch?


**FW v16.2.0 #801794e**

**UHK60** - No problems with this firmware, regardless of Fast Startup on/off.

**UHK80** - No problems with Fast Startup off, but something interesting with it enabled:

-  Booting up with UHK80 connected via USB; scroll speed is normal.
-  Then `switchHost` to dongle (no additional reboot); slow scrolling issue is present with both touchpad and mouse keys (_in addition to_ slow mouse keys movement issue).
-  `switchHost` back to USB again, and USB's scroll speed remains unaffected.
-  Rebooting UHK80 resumes normal dongle scroll speed, _kinda_. (The slow mouse keys scroll/movement via dongle is still present.)

**NOTE:** For clarity, the slow mouse keys via dongle issue is definitely much slower than normal, but nowhere near as bad as the scroll speed experienced in https://github.com/UltimateHackingKeyboard/firmware/issues/1406.

I've also never been able to reproduce #1406 when resuming from sleep (hibernate or S3 states). It only happens to me when powering back on after shutdown with Fast Startup enabled.

kareltucek:
> @kareltucek about the dongle mouse issue, I see that in messenger.c there is no retry mechanism for calling the Hid_Send*Report methods, do you have an idea what could be done there? 

@benedekkuper, you mean in the dongle's messenger? 

Since the bluetooth messages can't come more frequently than once every 7ms or so atm., I would expect the usb transport to be always available, or drop only rarely.

What exactly is the issue you are seeing?

------------------

If current implementation turns out to be a problem, then proper solution would be to run some minimal subset of the standard usb_report_updater and accumulate the reports there in the active report. That would probably be a task for me though.

kareltucek:
> **UHK80** - No problems with Fast Startup off, but something interesting with it enabled:
> 
>     * Booting up with UHK80 connected via USB; scroll speed is normal.
> 
>     * Then `switchHost` to dongle (no additional reboot); slow scrolling issue is present with both touchpad and mouse keys (_in addition to_ slow mouse keys movement issue).
> 
>     * `switchHost` back to USB again, and USB's scroll speed remains unaffected.
> 
>     * Rebooting UHK80 resumes normal dongle scroll speed, _kinda_. (The slow mouse keys scroll/movement via dongle is still present.)

The dongle is connected to the same pc? Is that dongle flashed to this firmware too? 

It seems like either:
- the dongle usb exchange is affected while right's isn't. (This would be weird if it was the same pc, etc...)
- or uhk80 right <-> dongle fail to communicate scroll multipliers to each other.

benedekkupper:
> @benedekkuper, you mean in the dongle's messenger?
> 
> Since the bluetooth messages can't come more frequently than once every 7ms or so atm., I would expect the usb transport to be always available, or drop only rarely.

I'm just going with the information given by Phil. `Hid_SendMouseReport` is called from `usb_report_updater` and from `messenger`, IIUC the latter is the data path for the dongle. It is true that the packet rate is at best 7ms, but one packet may deliver more than one mouse report at a time (I assume), which means that every subsequent mouse report in a packet is silently dropped at the moment.

kareltucek:
Ah, right. I will take care of that.

pcooke9:
> The dongle is connected to the same pc? Is that dongle flashed to this firmware too?

Yes, same PC and firmware.

pcooke9:
Does https://github.com/UltimateHackingKeyboard/firmware/actions/runs/23657501779 need tested? I'm trying to find something to kill a few hours with today.

kareltucek:
Yes please.

pcooke9:
**UHK80
16.2.0-d1944c4**

Oh, boy! That was the wackiest firmware I've tried yet!🫨🤪

The good news; _I think_ you fixed the slow mouse keys via dongle. _Maybe_?😅

The fun news; everything else is completely bonkers!🥴

-  So, when connected to a host via USB, every keypress gets stuck repeating.

-  When switching to dongle, every keypress is doubled, and sometimes gets scrambled with the last pressed key. Usually when pressing a key, it's doubled, but there's no output when pressing the same key again (no stuck keys via dongle though).

Resetting/rebooting the UHK80 doesn't resolve it. That's as far as I got before rolling back the FW. I tried to tame it for a few minutes, but to no avail.🙃

kareltucek:
Ah, right, I see 🫣

kareltucek:
Yeah, that was dumb. 

I have reverted that, and implemented simple semaphore blocking. That should work fine for dongle.

pcooke9:
**Testing FW v16.2.0 #2ff017c on UHK80:**

-  The issue with slow mouse keys via dongle is fixed with this FW.

-  [x] [#1406](https://github.com/UltimateHackingKeyboard/firmware/issues/1406) remains present via dongle when powering the PC back on with Windows Fast Startup enabled. (I'm still unable to reproduce it when resuming from hibernation or S3 states.)

-  Now that BT pairing is fixed, I can confirm that neither issue [#1406](https://github.com/UltimateHackingKeyboard/firmware/issues/1406), nor the other slow movement/scrolling via mouse keys is present when connected to my PC via BT6.0-USB adapter.

-  Connecting via BLE to my phone and to my PC seems to behave about the same as with previous firmwares. No noticeable changes there.

So far, everything else seems fine on UHK80 as well. I haven't tried this one on UHK60 yet.

benedekkupper:
> **Testing FW v16.2.0 #2ff017c on UHK80:**
> 
>     * The issue with slow mouse keys via dongle is fixed with this FW.
> 
>     * [#1406](https://github.com/UltimateHackingKeyboard/firmware/issues/1406) remains present via dongle when powering the PC back on with Windows Fast Startup enabled. (I'm still unable to reproduce it when resuming from hibernation or S3 states.)
> 
>     * Now that BT pairing is fixed, I can confirm that neither issue [#1406](https://github.com/UltimateHackingKeyboard/firmware/issues/1406), nor the other slow movement/scrolling via mouse keys is present when connected to my PC via BT6.0-USB adapter.
> 
>     * Connecting via BLE to my phone and to my PC seems to behave about the same as with previous firmwares. No noticeable changes there.
> 
> 
> So far, everything else seems fine on UHK80 as well. I haven't tried this one on UHK60 yet.

So are you saying that #1406 is only present on the dongle, but not when the keyboard is directly connected over USB?

pcooke9:
> So are you saying that #1406 is only present on the dongle, but not when the keyboard is directly connected over USB?

Correct, it only happens over dongle.

pcooke9:
Let me be more precise, in case it matters.

#1406 only happens via dongle. 
It doesn't matter if the UHK80 is plugged in with a USB cable or not.

If the UHK80 is connected with both USB and dongle to the same PC:
-  It will not happen when USB is the active host.
-  It will not happen when a BLE device is the active host.
-  It _will happen_ when the dongle is the active host.

kareltucek:
Well, that may be an issue of dongle <-> right sync rather than c2usb. Will look into it. 

> Connecting via BLE to my phone and to my PC seems to behave about the same as with previous firmwares. No noticeable changes there.

@pcooke9 by previous firmwares, you mean master (normal releases)? 

Is that a good thing or a bad thing in this context?

Please feel free to mark those points that require attention with `- [ ]`.

benedekkupper:
> Well, that may be an issue of dongle <-> right sync rather than c2usb. Will look into it.

@kareltucek for your attention: https://github.com/UltimateHackingKeyboard/firmware/pull/1482/changes/f66708e09ee5de70ba332d547e77bfbacdcf1f04

pcooke9:
> @pcooke9 by previous firmwares, you mean master (normal releases)?

Yes, I meant in comparison to normal releases.

Basically, when using only one BLE device, it's pretty stable. Using more than one BLE device still has connection issues after switching hosts a few times.

> Is that a good thing or a bad thing in this context?

Neutral, I guess? I only mentioned it because I wasn't really sure if BLE improvements were to be expected with this one (there's a lot to keep track of, sorry 😅).

benedekkupper:
I tried to test the BLE pairing problem on UHK80, but I'm unable to discover the device over BLE (using https://github.com/UltimateHackingKeyboard/firmware/actions/runs/23749469339/job/69187090071), any idea what's wrong?

kareltucek:
> I tried to test the BLE pairing problem

The double keypress in pairing screen is fixed. I am not aware of any other pairing problem?

> unable to discover the device over BLE

 Apart from too many connections already connected, no.

I would see what `uhk connections` in zephyr shell says.

benedekkupper:
@kareltucek @mondalaci @pcooke9 can you sum up the list of regressions on your side in this PR compared to master?

kareltucek:
I am now aware only of this, but given that only dongle is affected, I suspect that it is not a c2usb issue:

>  #1406 remains present via dongle when powering the PC back on with Windows Fast Startup enabled. (I'm still unable to reproduce it when resuming from hibernation or S3 states.)

I am a bit apprehensive about the hogp freezes that I reported above, as I am not aware of any related fixes, but it looks like Phil is not affected, and the last time I tested, I didn't run into it.

pcooke9:
>  [#1406](https://github.com/UltimateHackingKeyboard/firmware/issues/1406) remains present via dongle when powering the PC back on with Windows Fast Startup enabled.

This appears to be the only issue I'm aware of now.

pcooke9:
> I am a bit apprehensive about the hogp freezes that I reported above, as I am not aware of any related fixes, but it looks like Phil is not affected, and the last time I tested, I didn't run into it.

I don't recall any freezes here. Was it a hard freeze, requiring a reboot/reset?

Can you give more details to help me try to reproduce it? Did you also have a dongle, USB, or more than one BLE device connected when it froze?

kareltucek:
> I don't recall any freezes here. Was it a hard freeze, requiring a reboot/reset? 

Actually, after some time (10 or so seconds) the board would usually have crashed, so I don't recall having to hard reset it often. But yes, it would require power-cycle. 

The freeze kicked in right after a hogp host connected. (Either right after pairing, or right after connection if it was already paired.)

> Did you also have a dongle, USB, or more than one BLE device connected when it froze?

Argh. I didn't write that down. 🙈

kareltucek:
~~@benedekkupper do you have some idea on how to make the error numbers traceable? On uhk-code level, they are anonymous integers that lead into hid::result which is practically ungreppable, but after some digging leads into c2usb.hpp, where resides a clearly incomplete "enum" that however contains no numbers - just more defines, this time without a clear trace as to where they go.~~

~~One hypothesis is that they might be standard errno.h numbers. But as a reader of the code, how should I know that?~~

-------

Once a host is successfully paired, things seem to work reliably. 

However, often after a "successful" pairing, the device ends up in a non-functional state, returning some of -16, -104 and -128 error codes.

The only solution is to forget the connection on the device and repair again :-(. Just disconnecting such connection after we detect the problem doesn't seem to suffice.

This seems like a regression to me. Not sure about its source.

One such failed attempt (android device): 

```
<wrn> Bt: Pairing completed: n/a (n/a, d4:5e:89:65:a3:f8), bonded 1
--- Pairing ended, success = 1: Successfuly bonded! ---
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (BtPair_EndPairing)
<inf> Bt: Pairing complete, passing connection 7 to authenticatedConnection handler. Selected conn is 7
<inf> Conn: Host: 6(Unregistered Ble, Connected), Peer host2, Selected
<inf> Bt: Established HID connection with host2 (Unregistered Ble, d4:5e:89:65:a3:f8)
<inf> Conn: Host: 6(Unregistered Ble, Ready), Peer host2, Selected
<inf> Conn: Host: 6(Unregistered Ble, Ready), Peer host2, Active
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (connectHid)
<inf> Notify: Notification: Pairing succeeded!
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (pairing_complete)
<inf> Bt: BtManager: Start 'advertising' (0).
<inf> Bt: Adv: 'NUS' started
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Send failed: -104
Probable jump detected at site UhkModuleDriver2! -193, -47
Probable jump detected at site UhkModuleDriver2! -193, -47
Probable jump detected at site UhkModuleDriver2! -193, -47
Probable jump detected at site UhkModuleDriver2! -193, -47
Probable jump detected at site UhkModuleDriver2! -196, -47
Probable jump detected at site UhkModuleDriver2! -198, -46
Send failed: -104
Probable jump detected at site MouseController1! -198, -46
Probable jump detected at site MouseController2! -198, -46
Send failed: -128
Send failed: -104
Send failed: -104
Send failed: -128
<inf> hogp: starting HID app REPORT: 1
<wrn> bt_gatt: Device is not subscribed to characteristic
<wrn> bt_gatt: Device is not subscribed to characteristic
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
Send failed: -16
```

Steps to reproduce:
- Pair android phone.
- Start mousing immediately.
- Observe no cursor movement, just a long stream of errors. 
- If it works fine, then:
- Long press the corresponding device slot switch action to clear the bonds, start over.

benedekkupper:
> @benedekkupper do you have some idea on how to make the error numbers traceable? On uhk-code level, they are anonymous integers that lead into hid::result which is practically ungreppable, but after some digging leads into c2usb.hpp, where resides a clearly incomplete "enum" that however contains no numbers - just more defines, this time without a clear trace as to where they go.

The c2usb error type is just an alias to the C++ standard [`std::errc`](https://en.cppreference.com/w/cpp/error/errc.html), which itself is an alias to POSIX error codes. (With the notable difference that `c2usb::result` includes the negation in the type, unlike the others.) It would be futile to try to map the different error codes that can occur through zephyr's USB and BLE APIs to a different type...

> Once a host is successfully paired, things seem to work reliably.

Before I was testing with the right half only, not knowing that it doesn't advertise anymore if it doesn't have a connection to the left side yet. Now I could connect over BLE to Windows 11 PC, and haven't found any issues. I will test other OSes on Friday.

> This seems like a regression to me. Not sure about its source.

On my side there hasn't been any noteworthy changes to BLE, other than the report sending, which now doesn't block but returns an error instead.

pcooke9:
Regarding hogp freezes; I tested with both v16.2.0-2ff017c and 16.2.1-acebaf1, but the UHK80 never froze or crashed. I tried to reproduce it for at least thirty minutes or so.🤷🏻‍♂️

The following is the only unresponsive mouse output that I observed in this testing (only happens occasionally):

16.2.1-acebaf1
Bridge cable and USB-both halves to PC
Android phone BLE

After pairing phone, the mouse cursor will temporarily be visible on the phone, but unresponsive. Mouse keys and touchpad output don't work, but other keystrokes and functions are fine. Switching host to USB works as normal, but mouse output remains unresponsive when switching back to phone.

Rebooting the UHK80 resolves it.

UART adapter output:

```
<wrn> Bt: Pairing completed: n/a (n/a, 94:e1:29:c9:80:67), bonded 1
--- Pairing ended, success = 1: Successfuly bonded! ---
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (BtPair_EndPairing)
<inf> Bt: Pairing complete, passing connection 6 to authenticatedConnection handler. Selected conn is 6
<inf> Conn: Host: 5(Unregistered Ble, Connected), Peer host1, Selected
<inf> Bt: Established HID connection with host1 (Unregistered Ble, 94:e1:29:c9:80:67)
<inf> Conn: Host: 5(Unregistered Ble, Ready), Peer host1, Selected
<inf> Conn: Host: 5(Unregistered Ble, Ready), Peer host1, Active
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (connectHid)
<inf> Notify: Notification: Pairing succeeded!
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (pairing_complete)
<inf> Bt: BtManager: Start 'advertising' (0).
<inf> Bt: Adv: 'NUS' started
<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<wrn> UsbReports: Send failed: Connection reset by peer

<inf> hogp: starting HID app REPORT: 1
<wrn> bt_gatt: Device is not subscribed to characteristic
<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<inf> Bt: Kicking HID connections to check if HOGP is healthy.
HOGP HealthCheck: OK (registered, peer connected, interval 9)
<inf> Bt: BtConn_KickHid: the connection looks fine.

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy

<wrn> UsbReports: Send failed: Device or resource busy
```

___


Even more rare, but similar variation to the above; sometimes the OLED shows the phone (BLE) as the active host, but output is going to the USB host instead. (lost UART output for this one, sorry)

pcooke9:
In case I'm missing something; is simply having `set devMode true` in $onInit enough, or do I need to do anything else to get better logs?

kareltucek:
Not at the moment. 

It is possible to tweak log levels using `log status` and `log enable {dbg|inf|wrn|err} {logChannel}` in the zephyr console (or the same with `zephyr ` prefix in macro), but I don't know what to look for in any of the increased log levels. In practice, I am seeing exactly just as little as you do. 

Anyways, this is one variation of what I am seeing - one of them is busy, one of them is connection reset by peer and the last one is key has been revoked. I hope Benedek will be able to shed some light on it, otherwise it is going to be a couple daunting sessions for me in order to trace the origins of those messages.

kareltucek:
> Even more rare, but similar variation to the above; sometimes the OLED shows the phone (BLE) as the active host, but output is going to the USB host instead. (lost UART output for this one, sorry)

When you get into this state, what happens if you switchHost to usb and then again to the BLE?

pcooke9:
> one of them is connection reset by peer

I've seen this as well.



> When you get into this state, what happens if you switchHost to usb and then again to the BLE?

I can't remember for sure, but I'll let you know if I see it again.

kareltucek:
> @kareltucek for your attention: https://github.com/UltimateHackingKeyboard/firmware/commit/f66708e09ee5de70ba332d547e77bfbacdcf1f04

Right. That explains why only dongle operation is affected. It is also not a regression.

My guess is that the Hid_MouseScrollResolutionsChanged doesn't get called when the instance().ms_enum_.msos2_support() state changes in the above workaround. (Or doesn't get called with the correct arguments.)

```
extern "C" void Hid_MouseScrollResolutionsChanged(
    hid_transport_t transport, float verticalMultiplier, float horizontalMultiplier)
{
#if DEVICE_IS_UHK_DONGLE
    DongleScrollMultipliers.vertical = verticalMultiplier;
    DongleScrollMultipliers.horizontal = horizontalMultiplier;
    StateSync_UpdateProperty(StateSyncPropertyId_DongleScrollMultipliers, NULL);
#endif
}
```

I am adding a naive update check into the main loop for it... I guess it is not worth to spend more time on it.

pcooke9:
I just noticed a couple of issues that are present with FW v16.2.1-3db9354, but not v16.2.0-2ff017c (so something done on April 7th?).

1.  UHK80 shows an error when booting my PC. Only seems to happen when USB is plugged in; doesn't happen if dongle is the only thing connected.

When reaching the Windows login screen, the LEDs turn white (if left-half USB isn't connected, left side LEDs aren't affected) and then OLED shows triangle warning icon.

Agent error panel shows:

```
udc: Failed to allocate net_buf 1012
ASSERTION FAIL @ WEST_TOPDIR/zephyr/lib/net_buf/buf.c:460
```

___


2.  UHK80 crash loop when opening the Epic Games Launcher on PC.

**EDIT:** Also happens when clicking the "Sign in" button on the Epic Games website https://store.epicgames.com/en-US (Google Chrome or Microsoft Edge browsers. Doesn't happen with Firefox.)

Logs collected when using FW v16.2.1-2cbc849:

<details>
<summary>UART Log</summary>

```
<err> udc: Failed to allocate net_buf 2050

ASSERTION FAIL @ WEST_TOPDIR/zephyr/lib/net_buf/buf.c:460
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f38
<inf> fs_nvs: data wra: 2, 278
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
CRASH DETECTED, waiting for 5 seconds to allow Agent to reenumerate
<wrn> udc_nrf: USB Address incorrect 0x02

<inf> Conn: Host: 0(USB, Ready)

<inf> Conn: Host: 0(USB, Ready), Active
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 288
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
Enabling UART
<inf> NusServer: NUS Server module initialized.
<inf> Conn: UartLeft: (Ready)
<inf> StateSync: Resetting left right link! Bidirectional
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 288
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
CRASH DETECTED, waiting for 5 seconds to allow Agent to reenumerate
<wrn> udc_nrf: USB Address incorrect 0x02
<inf> Conn: Host: 0(USB, Ready)
<inf> Conn: Host: 0(USB, Ready), Active
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 288
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
Enabling UART
<inf> NusServer: NUS Server module initialized.
Scan module initialized
<inf> NusClient: NUS Client module initialized
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 298
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
CRASH DETECTED, waiting for 5 seconds to allow Agent to reenumerate
<wrn> udc_nrf: USB Address incorrect 0x01
<inf> Conn: Host: 0(USB, Ready)
<inf> Conn: Host: 0(USB, Ready), Active
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 298
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
Enabling UART
<inf> NusServer: NUS Server module initialized.
Scan module initialized
<inf> NusClient: NUS Client module initialized
 <inf> udc_nrf: Preinit
*** Booting nRF Connect SDK v2.8.0-1d4b8cd3280e ***
*** Using Zephyr OS v3.7.99-c6951670cc67 ***
Wormhole is open, reboot to power mode 0 0
Going to resume!
----------
UHK 80 Right USB started
<inf> fs_nvs: 3 Sectors of 4096 bytes
<inf> fs_nvs: alloc wra: 2, f30
<inf> fs_nvs: data wra: 2, 298
<inf> bt_sdc_hci_driver: SoftDevice Controller build revision:
                         fe 2c f9 6a 7f 36 22 2e  a0 79 c0 40 be 2c 03 20 |.,.j.6". .y.@.,.
                         40 c2 f3 32                                      |@..2
<inf> bt_hci_core: HW Platform: Nordic Semiconductor (0x0002)
<inf> bt_hci_core: HW Variant: nRF52x (0x0002)
<inf> bt_hci_core: Firmware: Standard Bluetooth controller (0x00) Version 254.63788 Build 573996906
<inf> bt_hci_core: No ID address. App must call settings_load()
Settings: Found peer 'left' with address e93e9ac473e2
<inf> bt_hci_core: Identity: CD:EE:3C:39:FD:1A (random)
<inf> bt_hci_core: HCI: version 6.0 (0x0e) revision 0x104e, manufacturer 0x0059
<inf> bt_hci_core: LMP: version 6.0 (0x0e) subver 0x104e

uhk80:right$ Reading hardware config
Reading user config
Applying user config
Flashed User Config version: 13.0.0 (native version: 13.0.0., at 2cbc849 / 16f4a0371ac3035d1848c4ad59a04388)
<inf> StateSync: Setting battery mode to 0
Allocating unregistered bonds
<inf> StateSync: Resetting left right link! Unidirectional
User config applied
CRASH DETECTED, waiting for 5 seconds to allow Agent to reenumerate
<wrn> udc_nrf: USB Address incorrect 0x02
<inf> Conn: Host: 0(USB, Ready)
<inf> Conn: Host: 0(USB, Ready), Active
Enabling UART
<inf> NusServer: NUS Server module initialized.
Scan module initialized
<inf> NusClient: NUS Client module initialized
<inf> Bt: Starting bluetooth services.
<inf> Bt: BtManager: Start 'advertising and scanning' (0).
Bt: filling scan filters
<inf> Conn: UartLeft: (Ready)
<inf> StateSync: Resetting left right link! Bidirectional
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
<inf> StateSync: Skipping wake up, tid is 0
Filters matched: left (n/a, e2:73:c4:9a:3e:e9), connectable:1
Scan connecting: left (n/a, e2:73:c4:9a:3e:e9)
<inf> StateSync: Setting battery mode to 1
<inf> StateSync: Batteries: 63% 0% (3869 / 4066 mV; 0 / 0 mV)
<inf> StateSync: Batteries: 63% 0% (3869 / 4066 mV; 0 / 0 mV)
<inf> Bt: MTU exchanging left (n/a, e2:73:c4:9a:3e:e9)
<inf> Bt: Connected left (n/a, e2:73:c4:9a:3e:e9), 29 7
<wrn> bt_smp: Unexpected SMP code 0x0b
<inf> Bt: MTU exchange done for left (n/a, e2:73:c4:9a:3e:e9)
<inf> Conn: NusServerLeft: (Connected), Peer left
<inf> Bt: Initiating NUS connection with left (n/a, e2:73:c4:9a:3e:e9)
<wrn> NusClient: MTU exchange failed with left (n/a, e2:73:c4:9a:3e:e9), err -120
<inf> Conn: NusServerLeft: (Ready), Peer left
<inf> Bt: BtManager: Scheduling scan/adv in 100ms. (SetConnectionConfigured)
<inf> Bt: BtManager: Start 'advertising' (0).
<inf> Bt: Adv: 'HID "and NUS"' started
<inf> StateSync: Batteries: 63% 83% (3869 / 4066 mV; 3804 / 4000 mV)
<inf> StateSync: Batteries: 63% 83% (3869 / 4066 mV; 3808 / 4000 mV)
<inf> StateSync: Batteries: 63% 83% (3869 / 4066 mV; 3808 / 4000 mV)
<inf> StateSync: Batteries: 63% 83% (3868 / 4066 mV; 3808 / 4000 mV)
<wrn> bt_hci_core: opcode 0x2013 status 0x3a
```

</details>


After closing the Epic Games app or affected internet browser, the UHK80 settles down and displays the following error in Agent:

```
Scan module initialized
NusClient: NUS Client module initialized
```
___


I'm assuming this might be related to https://github.com/UltimateHackingKeyboard/firmware/issues/1169.

pcooke9:
FW v16.2.1-2cbc849 does appear to resolve #1406 via dongle with Fast Startup enabled. But I can't be sure if it resolves it when resuming from S3 or hibernation states, since I was never able to reproduce it that way.

benedekkupper:
@pcooke9 these crashes are on me, can you retry them with the new build?

pcooke9:
> @pcooke9 these crashes are on me, can you retry them with the new build?

Ok, FW v16.2.1-6bd914a resolves the error when booting the PC, and the crash loop.😎

The `udc: Failed to allocate net_buf 1026` and `udc: Failed to allocate net_buf 2050` messages mentioned in #1169 are still present when interacting with the Epic Games app/site, but I don't see any negative effects otherwise.

benedekkupper:
> The `udc: Failed to allocate net_buf 1026` and `udc: Failed to allocate net_buf 2050` messages mentioned in #1169 are still present when interacting with the Epic Games app/site, but I don't see any negative effects otherwise.

This is expected (error log printing without any negative effect), the good news is that zephyr is migrating to a different allocation strategy in its upcoming release, which will eliminate this issue.

benedekkupper:
I tested with Windows 11, mac and Android, they work fine for me at least for a 5 minute test each. From your logs I could reconstruct the problem of freezing BLE input, I pushed the fix here, feel free to test again.

kareltucek:
Thanks! You are the best!

Will test on monday. (I am away from my uhks now.)

pcooke9:
> the problem of freezing BLE input, I pushed the fix here, feel free to test again.

Seems fixed on Win 11 and Android for me as well. 😃

kareltucek:
Also works well for me.

kareltucek:
Works like a charm! 

Let's merge it.

kareltucek:
@pcooke9 @benedekkupper one more thing: can you help me compile changelog?

kareltucek:
So far I have:

C2usb refactor:
    - Refactor: port uhk60 onto c2usb in order to increase codebase maintainability.
    - Refactor: rewrite usb stack integration.
    - Fix: increase bluetooth reliability - bluetooth hid interfaces should no longer stop functioning randomly.
    - Fix: implement workaround for slow scrolling on microsoft hosts.

mhantsch:
Uh oh, will this finally resolve my BT issues with the UHK80? ;)

kareltucek:
Well, I am not *that* optimistic.

mhantsch:
> Well, I am not _that_ optimistic.

Honestly, I merged master into my fork, recompiled, updated and now I have my work laptop and my phone both paired to UHK80 on bluetooth, and seem to be able to switch between them. That's the first time that has worked consistently. I'll try some more over the next days... but it's looking better than ever before.

benedekkupper:
> So far I have:
> 
> C2usb refactor: - Refactor: port uhk60 onto c2usb in order to increase codebase maintainability. - Refactor: rewrite usb stack integration. - Fix: increase bluetooth reliability - bluetooth hid interfaces should no longer stop functioning randomly. - Fix: implement workaround for slow scrolling on microsoft hosts.

I couldn't have said it better myself - there's a lot more changes, but they aren't really relevant for the users. Glad to see this finally land :)

pcooke9:
> can you help me compile changelog?

It feels weird having so little to say about such a hefty PR, but I can't think of anything else to add.

PR #1495, opened 2026-02-19
Don't restart scanning / advertising from callbacks.

kareltucek:
This may fix the issue or give better logs: 

https://forum.uhk.io/t/dongle-stuck-in-pairing-mode-violet/2782


kareltucek:
[uhk-firmware-16.2.0-e799aa8.tar.gz](https://github.com/user-attachments/files/25417677/uhk-firmware-16.2.0-e799aa8.tar.gz)

PR #1501, opened 2026-02-26
Doubletap:  Itroduced multitaps and made secondary role trigger primary on those instead of strict doubletaps

### LINKED ISSUE #1500: Possible further improvement to doubletap
This is not meant as an "It must be so!" kind of thing, just a thought:
With doubletap as is is now, if I start spamming a key, every other activation will be registered as a doubletap.  All fine and good, but when for instance I'm navigating through something tapping my dual role arrow key, if I then start holding it to doubletap to primary, if I had tapped an uneven number of times, I will not get a doubletap-to-primary activation, which is about half the time.
It is possible in the implementation that I made with global doubletap to not actually "consume" a doubletap every time.  Basically, a key will be eligible as a doubletap, but unless it's actually used as a doubletap, it won't block the next tap from being one.  That is to say that we could make it so that a key is not used as a doubletap unless something triggers from it, be that primary activation, an `ifDoubletap` in a macro, or a layer lock in the layer case.  The primary effect would, in the immediate term, be that a "spam-to-hold" on secondary role keys would always allow double-tap to primary.

I will do a quick PR soon, unless someone adamantly wants me not to.

firngrod:
The main benefit is that secondary role keys can now doubletap to primary after any number of taps, not just even numbers.  So when you spam a key like an arrow key, or a Vim navigation key and then decide to hold it, it will consistently trigger primary hold.
It could also theoretically have some effect in macros, but that is much less likely.
Since I cannot run tests, only having an OG 60V1, I decided to not write any blindly.  I have verified that `ifDoubletap`, secondary role doubletap-to-primary and layer doubletap-to-toggle work as expected.

Fixes https://github.com/UltimateHackingKeyboard/firmware/issues/1500

### DIFF
--- a/right/src/key_history.h
+++ b/right/src/key_history.h
@@ -15,7 +15,8 @@
 
 void KeyHistory_RecordPress(const key_state_t *keyState);
 void KeyHistory_RecordRelease(const key_state_t *keyState);
-bool KeyHistory_WasLastDoubletap();
-void KeyHistory_ConsumeDoubletap(const key_state_t *keyState, uint8_t keyActivationId);
+bool KeyHistory_WasLastStrictDoubletap();
+bool KeyHistory_WasLastLooseDoubletap();
+
 
 #endif
--- a/right/src/key_history.h
+++ b/right/src/key_history.h
@@ -15,8 +15,8 @@
 
 void KeyHistory_RecordPress(const key_state_t *keyState);
 void KeyHistory_RecordRelease(const key_state_t *keyState);
-bool KeyHistory_WasLastStrictDoubletap();
-bool KeyHistory_WasLastLooseDoubletap();
+bool KeyHistory_WasLastDoubletap();
+bool KeyHistory_WasLastMultitap();
 
 
 #endif
--- a/right/src/key_history.h
+++ b/right/src/key_history.h
@@ -16,5 +16,6 @@
 void KeyHistory_RecordPress(const key_state_t *keyState);
 void KeyHistory_RecordRelease(const key_state_t *keyState);
 bool KeyHistory_WasLastDoubletap();
+void KeyHistory_ConsumeDoubletap(const key_state_t *keyState, uint8_t keyActivationId);
 
 #endif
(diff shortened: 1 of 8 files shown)


firngrod:
Actually, another approach could be to track doubletaps both strictly (as we do now) and loosely (not caring about even/uneven) at every press and then use whichever is relevant. Could even be exposed as an argument for `ifDoubletap` if there is ever a request for it.  Will see if that is simpler tomorrow. It probably is.

mhantsch:
I would do it loosely, i.e. any tap that is within a doubletap timeout of a previous tap is considered a doubletap (of course, without interruption by other keys...), no matter whether that previous tap was a first tap, a double tap, a third tap....

firngrod:
I will insist on the current behavior as the default for other cases than Secondary Role.  Just like if I doubleclick a shortcut twice on my desktop, I want two instances to open, not three.

right/src/key_history.h
```diff
@@ -15,6 +15,8 @@
 
 void KeyHistory_RecordPress(const key_state_t *keyState);
 void KeyHistory_RecordRelease(const key_state_t *keyState);
-bool KeyHistory_WasLastDoubletap();
+bool KeyHistory_WasLastStrictDoubletap();
+bool KeyHistory_WasLastLooseDoubletap();
```
firngrod: A better name might be `WasLastMultiTap`

mhantsch:
> if I doubleclick a shortcut twice on my desktop, I want two instances to open, not three.

Yes, ok, that makes sense for general doubletaps.

firngrod:
Don't merge, it has a bug!

firngrod:
It should be good now.  I will run it on my board and update if I find any issues.

kareltucek:
> Since I cannot run tests, only having an OG 60V1, I decided to not write any blindly.

Still, would you suggest scenarios that you find worth to regress test for?

firngrod:
I'll look at it tonight, I hope.  I may try writing some blind tests.

firngrod:
Added two tests.  One testing that a release between taps prevent doubletap triggers, and one that double- and tripletaps trigger primary on secondary keys.

kareltucek:
Yep, thanks for a PR!

Will push test fixes into master - will be easier. (Mostly delays for debouncing, nothing major.)

PR #1507, opened 2026-03-08
Key History and Macros

firngrod:
Did some more key history work and exposed it in macros:
`ifDoubletap` has support for detecting a series of taps any number long, within reason, such as triple and so on.
Added `$previousKeyId` which can be used to create a repeat key, which I have seen requested.
Added `$previousKeyPressTime` which can be used to detect idle time before the key was pressed.  I know that has also been requested.

Also did a couple of small refactors where I ran across code which I thought could be optimized.

I want to do one more small refactor in macros:  Macros started using `call`, `fork` and `exec` should inherit more variables from their parents than `currentMacroKey` and `keyActivationId`.

I am okay with any part of this being rejected, I mostly did this for fun.

### DIFF
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -9,7 +9,7 @@
 typedef struct {
     const key_state_t *keyState;
     uint32_t timestamp;
-    uint8_t multiTapCount;
+    uint8_t multiTapCount : 6;
     uint8_t keyActivationId : 4;
     bool multiTapBreaker : 1;
 } key_press_event_t;
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -2,6 +2,10 @@
 #include "key_history.h"
 #include "postponer.h"
 
+#define HISTORY_SIZE 2
+#define LAST (position % HISTORY_SIZE)
+#define POS(p) ((position - p) % HISTORY_SIZE)
+
 typedef enum {
     DoubletapState_Blocked,
     DoubletapState_First,
@@ -14,20 +18,23 @@ typedef struct {
     uint8_t keyActivationId;
     uint32_t timestamp;
     doubletap_state_t doubletapState;
-} previous_key_event_type_t;
+} key_press_event_t;
 
-static previous_key_event_type_t lastPress;
+static key_press_event_t history[HISTORY_SIZE];
+static uint8_t position = 0;
 
 void KeyHistory_RecordPress(const key_state_t *keyState)
 {
+    const key_press_event_t * const lastPress = &history[LAST];
     const bool isMultitap = 
-        keyState == lastPress.keyState
-        && lastPress.doubletapState != DoubletapState_Blocked
-        && CurrentPostponedTime < lastPress.timestamp + Cfg.DoubletapTimeout;
+        keyState == lastPress->keyState
+        && lastPress->doubletapState != DoubletapState_Blocked
+        && CurrentPostponedTime < lastPress->timestamp + Cfg.DoubletapTimeout;
     const bool isDoubletap = isMultitap &&
-        (lastPress.doubletapState == DoubletapState_First || lastPress.doubletapState == DoubletapState_Multitap);
+        (lastPress->doubletapState == DoubletapState_First || lastPress->doubletapState == DoubletapState_Multitap);
 
-    lastPress = (previous_key_event_type_t){
+    ++position;
+    history[LAST] = (key_press_event_t) {
         .keyState = keyState,
         .keyActivationId = keyState->activationId,
         .timestamp = CurrentPostponedTime,
@@ -39,17 +46,29 @@ void KeyHistory_RecordPress(const key_state_t *keyState)
 
 void KeyHistory_RecordRelease(const key_state_t *keyState)
 {
-    if (keyState != lastPress.keyState) {
-        lastPress.doubletapState = DoubletapState_Blocked;
+    if (keyState != history[LAST].keyState) {
+        history[LAST].doubletapState = DoubletapState_Blocked;
     }
 }
 
-bool KeyHistory_WasLastDoubletap()
+bool KeyHistory_WasDoubletap(const key_state_t *keyState, uint8_t activationId)
 {
-    return lastPress.doubletapState == DoubletapState_Doubletap;
+    for (uint8_t i = 0; i < HISTORY_SIZE; ++i) {
+        const key_press_event_t * const event = &history[POS(i)];
+        if (event->keyState == keyState && event->keyActivationId == activationId) {
+            return event->doubletapState == DoubletapState_Doubletap;
+        }
+    }
+    return false;
 }
 
-bool KeyHistory_WasLastMultitap()
+bool KeyHistory_WasMultitap(const key_state_t *keyState, uint8_t activationId)
 {
-    return lastPress.doubletapState >= DoubletapState_Multitap;
+    for (uint8_t i = 0; i < HISTORY_SIZE; ++i) {
+        const key_press_event_t * const event = &history[POS(i)];
+        if (event->keyState == keyState && event->keyActivationId == activationId) {
+            return event->doubletapState >= DoubletapState_Multitap;
+        }
+    }
+    return false;
 }
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -6,18 +6,12 @@
 #define LAST (position % HISTORY_SIZE)
 #define POS(p) ((position - p) % HISTORY_SIZE)
 
-typedef enum {
-    DoubletapState_Blocked,
-    DoubletapState_First,
-    DoubletapState_Multitap,
-    DoubletapState_Doubletap,
-} doubletap_state_t;
-
 typedef struct {
     const key_state_t *keyState;
-    uint8_t keyActivationId;
     uint32_t timestamp;
-    doubletap_state_t doubletapState;
+    uint8_t multiTapCount;
+    uint8_t keyActivationId : 4;
+    bool multiTapBreaker : 1;
 } key_press_event_t;
 
 static key_press_event_t history[HISTORY_SIZE];
@@ -28,47 +22,31 @@ void KeyHistory_RecordPress(const key_state_t *keyState)
     const key_press_event_t * const lastPress = &history[LAST];
     const bool isMultitap = 
         keyState == lastPress->keyState
-        && lastPress->doubletapState != DoubletapState_Blocked
+        && !lastPress->multiTapBreaker
         && CurrentPostponedTime < lastPress->timestamp + Cfg.DoubletapTimeout;
-    const bool isDoubletap = isMultitap &&
-        (lastPress->doubletapState == DoubletapState_First || lastPress->doubletapState == DoubletapState_Multitap);
 
     ++position;
     history[LAST] = (key_press_event_t) {
         .keyState = keyState,
         .keyActivationId = keyState->activationId,
         .timestamp = CurrentPostponedTime,
-        .doubletapState = isDoubletap ? DoubletapState_Doubletap :
-            isMultitap ? DoubletapState_Multitap :
-            DoubletapState_First,
+        .multiTapCount = 1 + (isMultitap ? lastPress->multiTapCount : 0),
+        .multiTapBreaker = false,
     };
 }
 
 void KeyHistory_RecordRelease(const key_state_t *keyState)
 {
     if (keyState != history[LAST].keyState) {
-        history[LAST].doubletapState = DoubletapState_Blocked;
+        history[LAST].multiTapBreaker = true;
     }
 }
 
-bool KeyHistory_WasDoubletap(const key_state_t *keyState, uint8_t activationId)
-{
-    for (uint8_t i = 0; i < HISTORY_SIZE; ++i) {
-        const key_press_event_t * const event = &history[POS(i)];
-        if (event->keyState == keyState && event->keyActivationId == activationId) {
-            return event->doubletapState == DoubletapState_Doubletap;
+uint8_t KeyHistory_GetMultitapCount(const key_state_t *keyState, uint8_t activationId) {
+    for (uint8_t i = 0; i < HISTORY_SIZE - 1; ++i) {
+        if(history[POS(i)].keyState == keyState && history[POS(i)].keyActivationId == activationId) {
+            return history[POS(i)].multiTapCount;
         }
     }
-    return false;
-}
-
-bool KeyHistory_WasMultitap(const key_state_t *keyState, uint8_t activationId)
-{
-    for (uint8_t i = 0; i < HISTORY_SIZE; ++i) {
-        const key_press_event_t * const event = &history[POS(i)];
-        if (event->keyState == keyState && event->keyActivationId == activationId) {
-            return event->doubletapState >= DoubletapState_Multitap;
-        }
-    }
-    return false;
+    return 0;
 }
--- a/right/src/key_history.c
+++ b/right/src/key_history.c
@@ -4,15 +4,7 @@
 
 #define HISTORY_SIZE 2
 #define LAST (position % HISTORY_SIZE)
-#define POS(p) ((position - p) % HISTORY_SIZE)
-
-typedef struct {
-    const key_state_t *keyState;
-    uint32_t timestamp;
-    uint8_t multiTapCount : 6;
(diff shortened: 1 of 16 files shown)


right/src/macros/commands.c
```diff
@@ -136,18 +136,22 @@ static bool isCurrentMacroPostponing()
...
-        bool isSameActivation = (S->ms.currentMacroKey->activationId == S->ms.keyActivationId);
-        bool keyIsActive = (KeyState_Active(S->ms.currentMacroKey) && !PostponerQuery_IsKeyReleased(S->ms.currentMacroKey));
-        return  (isSameActivation && keyIsActive) || S->ms.oneShot == 1;
-    } else {
-        bool isSameActivation = (S->ms.currentMacroKey->activationId == S->ms.keyActivationId);
-        bool keyIsActive = KeyState_Active(S->ms.currentMacroKey);
-        return (isSameActivation && keyIsActive) || S->ms.oneShot == 1;
+        return !PostponerQuery_IsKeyReleased(S->ms.currentMacroKey);
     }
+    return true;
```
firngrod: I think this spells out the logic more, and it should also be more performant to exit as soon as any logic allows it?

right/src/key_history.c
```diff
@@ -2,54 +2,56 @@
...
-    lastPress = (previous_key_event_type_t){
+    ++position;
+    history[LAST] = (key_press_event_t) {
         .keyState = keyState,
         .keyActivationId = keyState->activationId,
         .timestamp = CurrentPostponedTime,
-        .doubletapState = isDoubletap ? DoubletapState_Doubletap :
-            isMultitap ? DoubletapState_Multitap :
-            DoubletapState_First,
+        .multiTapCount = 1 + (isMultitap ? lastPress->multiTapCount : 0),
```
firngrod, line 26: I can't believe I didn't do something like this the first time around.

mhantsch:
I like both `$previousKeyId` and `$previousKeyPressTime`. Appreciate!

mhantsch:
But why do they have to live in the macrostate? Wouldn't this be global info? Assume a macro runs for a longer time, and other keys are pressed while it is running. Would `$previousKeyId` and `$previousKeyPressTime` change? You are capturing them when the macro starts? Curious what makes more sense...

firngrod:
They are meant to be previous key from thisKey, not last key, which would generally be thisKey. So either I expand history enough to be _reasonably_ sure that it still has the relevant event, which, fair enough, would be 4-8 events, or I do this caching and become _certain_ I have the info.
It may well be better to expand history and move the source of truth to history for what values this can be done for. Will find out how much history I can get for the same RAM sometime.
It's a trade-off of where to put priorities. I will wait for Karel's input before I decide either way.

mhantsch:
Yes, RAM concerns I have. @kareltucek made sure I don't store macroArgs in the macro state but use a global pool across all macros. Balance for keeping memory consumption compact.

But key history size: hold shift key and type A LONGER SENTENCE ALL IN CAPITALS. How would that hold up regarding key history and a macro bound to the shift key?

I guess macro may need to cache the info at start of macro to avoid running out of 4-8 key history, e.g. `setVar prevKey $previousKeyId`

Not sure I like this solution.

Lock certain history items as "don't delete" because they were active when a macro started and just refer to the history index of that item to retrieve keyId and time when needed by the macro? Release the lock when macro ends?

Exiting times.

pcooke9:
> Exiting times.

"Exiting times" are always exciting.

I'll show myself out...
😉

firngrod:
Well it depends on how we would feel about a world in which a macro may loose access to key press information if it lives long enough.  I'm thinking we don't feel too good about it, even if long-lived macros are discouraged.  Sure, individual macros can mitigate by using variables, but it would have to be communicated very clearly in documentation, and going from **certainty** to **near-certainty** is a big leap.

I could make key history persist as long as a macro has reference to it, but it adds a lot of complexity and increases RAM usage in that history now realistically needs to be long enough that each macro slot can hold a reference on an individual key event and it's predecessor.  Finally, in a language like C without scoped deconstruction, I do not like such patterns.

Further, I don't currently don't see the value in extending key history except having lazy fetching of some information in some macro commands/variables.

firngrod:
One thing I have really appreciated about the UHK over the years, and since I started working in the codebase, is the lack of jank-by-design.  Even if it limits the available options, and bugs notwithstanding, one can be reasonably sure that things work as intended, robustly, even in edge cases.

mhantsch:
> make key history persist as long as a macro has reference to it

That's what I did with macroArgs. Each argument in the pool stores the macro state slot index as its owner, and when the macro ends, I clean up the pool and remove the arguments owned by that macro. Memory consumption is 8 bits * pool size (32), vs. storing pointers or even complete arguments in the macro state (size * maximum number of concurrent macros (16)).

Clean up is in core.c/endMacro().

firngrod:
Having thought about it some more, this is where I'm at:

Either we accept that macros may loose access to information if they live long enough, or we don't.  First, let's presume that we don't accept that:

The information has to live somewhere.  Having it live in macro state allows history to be only 2 long but increases macro state size.  Having it live in macro state allows history to be only 2 long but increases macro state size.  It can live in macro state or in history, but it has to live.  As such, there is no fundamental argument RAM-wise to be made for storing it in one place or the other.

Here are the arguments against having it live in history:
- Since history stores more information that macros require, and macros require their activation event, AND the one before, _AND_ we need to plan for every macro referencing events without overlap, **_AND_** we still need at least one event for history to function for the rest of the firmware, this is the more costly option in terms of RAM.
- Implementing a history which allows for frozen events adds considerable complexity.  It's fun, but also the costly option in terms of complexity, and in terms of processing *on every keystroke*, not just within macro scope.

And the arguments for having it live in history:
- We have history.  We currently do not have any requests for history beyond previous key, do we?

Now, let's say we accept the possibility that history leaves a macro behind and they may lose access to data:
We can make educated guesses and cost-benefit judgements on how long of a history we will allow before we loose data in macros.  We can save on RAM by not keeping everything forever, and we don't incur the cost of a complex history implementation.
There could be a clear disclaimer in the reference manual about historical data, like "Some variables and commands rely on historical data which will become unavailable a number of key presses after macro start, after which default values will be used.  These are marked as volatile."
And then "`ifPlaytime` - Volatile, defaults to true after X presses," and "`$previousKeyId` - Volatile, defaults to 255 after X-1 presses," and "`$previousKeyPressTime` - Volatile, defaults to 0 after ...".  Something like that.  Still something of a paradigm shift.

I will not start making that choice, but I would be happy to implement it if it was made.

mhantsch:
I realise in the comparison with `macroArg` there are some differences. Not every macro will have `macroArg`, nor will they all have the maximum number of `macroArg` statements, but (almost) every macro will have an activating keypress, an activation time, and a previous key. (The few exceptions are macros like `$onInit`...) Because every macro has this information, it is probably suited to live in macro state. As you said, it has to be stored somewhere. Just referencing a place where it is stored, and managing the lifetime of that referenced storage is probably overkill, and will not offer a significant reduction of memory.

Thanks for the conversation!

kareltucek:
I am mot sure we actually need these. Afaik, the only person that ever requested these was Max, and it was for his own macro-HRM implementation, which is a usecase that is kind of not supported on purpose. 

But assume we actually want them. Then keeping the info in the macro state is indeed the only reasonable choice. 

> I could make key history persist as long as a macro has reference to it, but it adds a lot of complexity 

Indeed. Let's not go that way. 

> That's what I did with macroArgs. 

Macro arguments are a very different case. 

----------------

My bottom line: I definitely pick the macro state. Current implementation I believe.

I will have to look the code over yet though.

right/src/postponer.c
```diff
@@ -33,7 +33,7 @@ uint8_t Postponer_LastKeyMods = 0;
 //static uint8_t cyclesUntilActivation = 0;
 static uint32_t lastPressTime;
 
-#define POS(idx) ((bufferPosition + POSTPONER_BUFFER_SIZE + (idx)) % POSTPONER_BUFFER_SIZE)
+#define POS(idx) ((bufferPosition + (idx)) % POSTPONER_BUFFER_SIZE)
```
kareltucek: This will fail with negative indexes. I am not sure how much illegal they are, but would prefer to leave the original version in place.

firngrod:
Don't yet.

right/src/macros/commands.c
kareltucek: Well, I think that the refactor logic is correct, but I don't think that the result is more readable. 

For me, this unrolled variant feels very alien since in order to understand how it behaves in real life, I have to hold the entire function in my mind and evaluate it, while in the original, I see the return on a single line: ` (isSameActivation && keyIsActive) || S->ms.oneShot == 1`, and if needed, I can backtrack from there to see how keyIsActive is evaluated. 

I have a strong preference to not include this  Macros_CurrentMacroKeyIsActive refactor.

right/src/key_history.c
kareltucek: ¯\\_(ツ)_/¯

right/src/key_history.h
```diff
@@ -9,14 +9,26 @@
...
 void KeyHistory_RecordPress(const key_state_t *keyState);
 void KeyHistory_RecordRelease(const key_state_t *keyState);
-bool KeyHistory_WasLastDoubletap();
-bool KeyHistory_WasLastMultitap();
 
+// Querying whole events
+const key_press_event_t * KeyHistory_GetPreceedingPress(const key_state_t *keyState, uint8_t activationId);
+
+// Querying specific info
+uint8_t KeyHistory_GetMultitapCount(const key_state_t *keyState, uint8_t activationId);
```
kareltucek, line 32: Is there any chance that any of these two functions gets called with any but the last pressed keyState? 

If yes, then I suspect that history size 2 might be insufficient. 

(Just thinking out loud. I don't think it is the case.)

kareltucek (changes requested):
A couple change requests, but otherwise, seems solid.

right/src/postponer.c
firngrod: Aha, that's why it was there!  Will it fail with negative indexes though?  I guess that depends on what type the compiler chooses to calculate in, but if it calculates in unsigned integers, the buffer will underflow, and since the modulo is a power of two, it will all work out.
Will put it back the way it was, of course. :)

right/src/key_history.h
firngrod: Currently, there is not, no.  I just put the interface in for good measure.  I think you mentioned it when I originally made the history.

firngrod:
Funny how preferences are different.  I much prefer the other factorization of `Macros_CurrentMacroKeyIsActive`.  I can easily see what each variable means for the output, and their hierarchy.  I will put it back the way it was of course, this is your house I just walked into and started moving the furniture around.  :)

Also, as I mentioned somewhere, there is one more thing I want to do before this PR is done:  I want child macros to inherit more state.

Also also, there [has been talk of a repeat key](https://forum.uhk.io/t/does-uhk-have-a-repeat-key/2809/11), which can be easily implemented with `$lastKeyPressId`.  I have also seen that concept talked about elsewhere in keyboard blogs, so I thought that since we have a concept of key history now, might as well...

mhantsch:
And even for the repeat key it may make sense to have timing information since that last keypress, e.g. only trigger repeat if that previous keypress was within the last x milliseconds...

kareltucek:
> Funny how preferences are different.

Oh, yeah 😅.

> this is your house

I don't feel that alone is a sufficient justification 🤔.

I think that with me being responsible for long-term maintenance is the more relevant one. Still, thanks :-).

> Also, as I mentioned somewhere, there is one more thing I want to do before this PR is done: I want child macros to inherit more state.

Sorry, I had missed that one. Yes, feel free to.

> Also also, there has been talk of a repeat key, which can be easily implemented with $lastKeyPressId. I have also seen that concept talked about elsewhere in keyboard blogs, so I thought that since we have a concept of key history now, might as well...

Sounds good!

right/src/macros/core.c
```diff
@@ -402,51 +402,14 @@ static bool macroIsValid(uint8_t index)
...
-
 uint8_t initMacro(
     uint8_t index,
     key_state_t *keyState,
     uint16_t argumentOffset,
     uint8_t keyActivationId,
     uint8_t parentMacroSlot,
-    const char *inlineText
+    const char *inlineText,
+    const macro_state_t *parentMacroState
```
firngrod, line 461: Fork does not get parentMacroSlot but should still inherit the info.  Therefore the function refactor.

right/src/macros/core.c
```diff
@@ -523,14 +506,64 @@ uint8_t Macros_StartMacro(
...
+//partentMacroSlot == 255 means no parent
+uint8_t Macros_StartMacro(
+    uint8_t index,
+    key_state_t *keyState,
+    uint16_t argumentOffset,
+    uint8_t keyActivationId,
+    bool runFirstAction,
+    const char *inlineText
+) {
+    return startMacro(index, keyState, argumentOffset, keyActivationId, MacroIndex_None, runFirstAction, inlineText, NULL);
```
firngrod, line 567: Without the forward declaration, these had to move here.

right/src/macros/core.h
```diff
@@ -277,7 +281,7 @@
     macro_result_t Macros_SleepTillTime(uint32_t time, const char* reason);
     uint8_t Macros_ConsumeLayerId(parser_context_t* ctx);
     uint8_t Macros_QueueMacro(uint8_t index, key_state_t *keyState, uint8_t keyActivationId, uint8_t queueAfterSlot);
-    uint8_t Macros_StartMacro(uint8_t index, key_state_t *keyState, uint16_t argumentOffset, uint8_t keyActivationId, uint8_t parentMacroSlot, bool runFirstAction, const char *inlineText);
```
firngrod, line 280: With the internal `startMacro`, this parameter was now `255` everywhere, so I removed it.

right/src/key_history.c
```diff
@@ -2,54 +2,57 @@
...
 
-typedef enum {
-    DoubletapState_Blocked,
-    DoubletapState_First,
-    DoubletapState_Multitap,
-    DoubletapState_Doubletap,
-} doubletap_state_t;
+#define HISTORY_SIZE 2
+#define LAST (position % HISTORY_SIZE)
+#define POS(p) ((position + HISTORY_SIZE - p) % HISTORY_SIZE) // Proceeding backwards in time
```
firngrod, line 7: You were right, underflows do not go well, even though I have this nagging feeling that they should.

right/src/utils.c
```diff
@@ -26,24 +26,15 @@
...
     return 0;
 #else
-    if (key == NULL) {
-        return 0;
-    }
-    uint32_t ptr1 = (uint32_t)(key_state_t*)key;
-    uint32_t ptr2 = (uint32_t)(key_state_t*)&(KeyStates[0][0]);
-    uint32_t res = (ptr1 - ptr2) / sizeof(key_state_t);
-    return recodeId(res, MAX_KEY_COUNT_PER_MODULE, 64);
+    uint16_t arrayPos = (uint16_t)(key - (key_state_t*)KeyStates);
```
firngrod, line 39: Oops, I should put the null-check back.  Should it not return 255 on null by the way?  0 is the F7 key, isn't it?

firngrod:
> I think that with me being responsible for long-term maintenance is the more relevant one.

That is what I mean.  You are the custodian, the one responsible.  Also the responsible one.

right/src/utils.c
kareltucek: 255 sounds good

right/src/macros/core.c
kareltucek: I would prefer forward declarations, since moving code wreaks havoc in git archeology sessions.

But this is an unimportant nitpick. Feel free to leave it as is.

firngrod: I would not be happy with having to maintain the parameter list in two places.  If you are okay with that, I will put the functions back and do forward declaration.

kareltucek: I am ok with that actually 😃

firngrod:
Maybe we should hold off on this merge, I want to consider how this will work with chords.

kareltucek:
What is the state of this? Should I go through the code once more and merge it?

firngrod:
It's going to complicate chords, so I'm not sure which should be merged first.

kareltucek:
Alright, no stress, just let me know when you want some activity from me ;-).

PR #1526, opened 2026-04-17
Usb throttling

kareltucek:
Changelog:
- Lower bluetooth latency.


kareltucek:
Rough estimate:

Expected latency before:
- sensor scanning: 10ms, 5ms average
- i2c: up to some 6ms, 3ms average, actually lower on uhk80 since only two or so drivers are in the loop.
- general processing by the firmware 1ms. Twice for dongle.
- on average half of a window before getting processed
- bluetooth in general - waiting in the queue for two ble windows
- Resulting into:
    - hogp: `2*11+0.5*11+5+3+1 = 36.5ms`
    - dongle: `2*7+0.5*7+5+3+2 = 27.5ms`

Expected latency after:
- sensor scanning, i2c: still the same, general processing: still the same
- on average half of a window before getting processed
- 6ms for ble stack processing instead of the two ble windows 
- Resulting into:
    - hogp: `6+0.5*11+5+3+1 = 20.5ms`
    - dongle: `6+0.5*7+5+3+2 = 19.5ms`
    
Lower bound limits:
- sensor scanning: 5ms average
- half of a transport window: 3.5 ms  average
- bluetooth internal processing? Can this be sped up?: 6ms
- something between 8.5 and 14.5 seems to be the lowest achievable latency

Possible further improvements: 
- synchronize i2c polling

PR #1539, opened 2026-05-05
Fix #1090: don't drop USB reports on transient hub stalls

### LINKED ISSUE #1090: UHK60 v2 getting stuck on infinite key repeat
Sometime in the last couple weeks, my UHK60 v2 started to get stuck several times a day, infinitely repeating a keypress. When this happens, the last keystroke repeats forever until I press another key on the keyboard. This happens with single key strokes as well as chorded ones e.g., ctrl-v. There is no clear pattern to what causes it or when. Seems random.

This is not key chattering. It is strictly a case of getting stuck on infinite repeat.

Unfortunately I can’t confidently say which firmware version this started with. I do check and install firmware updates every few days, and can recall 2-3 firmware updates in the last month. So it started sometime in there. Although I am by no means certain it's a firmware problem, it might be a hardware problem with my UHK.

In the meantime, I downgraded to firmware 12.0, and if the problem goes away, I will start slowly moving back up.

Per the request in https://forum.ultimatehackingkeyboard.com/t/uhk-v2-recently-getting-stuck-on-infinite-key-repeat/1935, I have attached a copy of my user config.
[UserConfiguration.json](https://github.com/user-attachments/files/18412478/UserConfiguration.json)

semnil:
## Summary

Fixes #1090. Under transient USB hub stalls, the previous `Hid_Send*Report()` failure path silently committed unsent reports as the "previously sent" buffer in the double-buffer scheme, breaking the host's view of key state. User-visible symptoms:

- **Stuck-key auto-repeat** — a dropped release report leaves the key in the host's pressed set until the next press happens to differ.
- **Dropped short presses** — same mechanism, with a press dropped before its release lands.
- **Swapped consecutive bigrams** (e.g. typing `se` produced `es`) — both keys merged into a single multi-key report whose slot order matches matrix order, not press order.
- **Disturbance to other HID devices on the same hub** during sustained UHK input, because the unrate-limited retry path saturated the upstream IN-token queue.

## Changes

Three commits for the fix plus one independent UHK60 build prerequisite:

1. `9616cb4` **Preserve report state on USB send failure** — branch on `ret != 0`; never call `switchActive*Report()` when the send fails. The next merge cycle rebuilds the active buffer from the latest cached state and `CheckReportReady` retries once the link recovers.
2. `a8aa0d4` **Throttle USB resend retries via EventScheduler** — replace the synchronous `EventVector_Set(EventVector_ResendUsbReports)` with `EventScheduler_Schedule(now + 4 ms, EventSchedulerEvent_UsbResend, ...)`. Keyboard / controls / mouse retry wakeups collapse into a single ~250 Hz upper bound across all three paths, freeing the hub upstream queue between attempts.
3. `373cf4f` **Rename `ShouldResendReport` → `IsWithinResendWindow`** — the helper now only manages the per-counter retry window (consumed for log routing in `usb_report_updater.c` and as the resend-loop predicate in `device/src/messenger.c`); the new name reflects that role.
4. `8b36cd3` **Move `logger.h` include in `debug.c`** — independent of #1090 but bundled as a prerequisite UHK60 build fix: `Debug_RecordBleSendResult()` references `LogU()` in a dead-code branch and GCC 15 promotes implicit-declaration warnings to errors.

Applied identically to the keyboard, controls, and mouse send paths. New `EventSchedulerEvent_UsbResend` added to `right/src/event_scheduler.{h,c}`.

## Detailed design notes

Full root-cause analysis, per-symptom mechanism walkthrough, and pre/post-fix mermaid sequence diagrams are kept on a separate docs branch to keep this PR focused on code:

- https://github.com/semnil/firmware/blob/docs/issue-1090-usb-resend-state/doc-dev/other/issue-1090-usb-resend-state.md

### DIFF
--- a/device/src/messenger.c
+++ b/device/src/messenger.c
@@ -241,7 +241,7 @@ static void processSyncablePropertyDongle(device_id_t src, const uint8_t* data, 
 
 #if DEVICE_IS_UHK_DONGLE
     uint8_t retryCounter = 0;
-    while (ShouldResendReport(ret == 0, &retryCounter)) {
+    while (IsWithinResendWindow(ret == 0, &retryCounter)) {
         k_sem_take(&dongleUsbSem, K_MSEC(128));
         ret = sendDongleReport(propertyId, message);
     }
--- a/right/src/debug.c
+++ b/right/src/debug.c
@@ -1,8 +1,8 @@
 #include <string.h>
 #include "debug.h"
+#include "logger.h"
 
 #ifdef __ZEPHYR__
-#include "logger.h"
 #include "keyboard/oled/screens/screen_manager.h"
 #include <zephyr/kernel.h>
 #else
--- a/right/src/event_scheduler.c
+++ b/right/src/event_scheduler.c
@@ -246,6 +246,9 @@ static void processEvt(event_scheduler_event_t evt)
             BtConn_KickHid();
 #endif
             break;
+        case EventSchedulerEvent_UsbResend:
+            EventVector_Set(EventVector_ResendUsbReports);
+            break;
         default:
             return;
     }
--- a/right/src/event_scheduler.h
+++ b/right/src/event_scheduler.h
@@ -49,6 +49,7 @@
         EventSchedulerEvent_UnselectHostConnection,
         EventSchedulerEvent_OneShotTimeout,
         EventSchedulerEvent_KickHid,
+        EventSchedulerEvent_UsbResend,
         EventSchedulerEvent_Count
     } event_scheduler_event_t;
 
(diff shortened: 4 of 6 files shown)


semnil:
https://github.com/UltimateHackingKeyboard/firmware/pull/1540

kareltucek:
Hi there! 

Sorry for the late reply!

A problem is that this will stuck the keyboard until that report is successfully sent. That is a far too high price, especially with uhk80's multi-host paradigm.

What exactly is the problem on your end

- Are those usb stalls caused by that fast usb output? 
- How long are they? I would expect 128ms to be long enough for the hub to recover in case their cause is independent from UHK. If not, we can seek ways to increase that timeout. 

Would you please take a look at https://github.com/UltimateHackingKeyboard/firmware/pull/1544 ? Does it fix the issue for you, or do we need to find way to increase that timeout?

kareltucek:
I have merged the other PR. If you have any comments or insights, please do share ;-).

Closing for now.

PR #1543, opened 2026-05-08
Fix bit-shift typo in IS31FL3737B configuration register write.

semnil:
`<` was used where `<<` was intended, leaving the PWM frequency at 8.4 kHz (chip default) instead of the intended 26.7 kHz on UHK60v2 / IS31FL3737B.


kareltucek:
Nice catch! Thanks!

PR #1544, opened 2026-05-08
Throttle usb retries.

kareltucek:
Changelog:

- fix: don't clog usb hubs by too fast usb retries.
- fix: in case of failed report sends, always send up-to-date state once the connection is alive again.
- fix: change resending logic to allow up to 1024ms long usb hub stalls without loosing reports.

### DIFF
--- a/right/src/event_scheduler.c
+++ b/right/src/event_scheduler.c
@@ -246,6 +246,9 @@ static void processEvt(event_scheduler_event_t evt)
             BtConn_KickHid();
 #endif
             break;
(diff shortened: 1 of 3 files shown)


kareltucek:
Also, when giving up, we should still make the keyboard send some state once the connection is alive. This way, we loose some keystrokes pressed along the way, but keys should not get stuck due to dropped report.

kareltucek:
Actually added logic to allow up to 1024ms stalls without freezing up the keyboard.

Issue #1114, opened 2025-02-03
Jittery mouse keys movement when using dongle


pcooke9:
After upgrading the firmware to Release v12.3.3, the movement via mouse keys is jittery when using the dongle. It feels similar to the BLE mouse jitter, but not quite as bad. Previous firmware versions were buttery smooth.

The touchpad movement feels fine though.

kareltucek:
~~Would you please give a try to this firmware?~~

~~[uhk-firmware-12.3.3-uart-fixes-v6.5.tar.gz](https://github.com/user-attachments/files/18644573/uhk-firmware-12.3.3-uart-fixes-v6.5.tar.gz)~~

Scratch it. Already got the trouble with it.

kareltucek:
Do modules feel jittery over usb?

pcooke9:
No, not over USB. Pretty smooth there.

Touchpad movement over dongle feels like it has a similar amount of jitter as dongle's mouse keys though. (Touchpad is the only module I have, btw.)

kareltucek:
Two findings:

- there was a scheduling issue, which forced 11ms ble intervals instead of 7.5ms, because full 7.5ms were reserved for data transmission. Solved by reducing the reservation to 2.5ms. This means slower transmission of big packets (as they will likely be fragmented)

- running advertisement is disrupting ble streaming and makes us miss transport windows every ~35ms.

pcooke9:
I haven't tried FW release v17.1.1 yet, but I still experience this with FW 17.1.1-a3ec363 staging branch.

I'm not sure if it's just an issue on my system or not. Can anyone else confirm if this is still present with dongle mouse keys on Win 11?

kareltucek:
Let's reopen and debug it. I am afraid we will need access to dongle zephyr shell for that, which means that we are blocked by https://github.com/UltimateHackingKeyboard/agent/pull/2956

kareltucek:
Well, if you use https://github.com/UltimateHackingKeyboard/agent/actions/runs/27213615425 , it will give you access to dongle's zephyr shell (https://uhk.io/knowledgebase/how-to-report-firmware-issues). If you type in `uhk jitterTest 1`, the terminal will dump a log of first 50 mouse moves every 1 second. First column is time delta, second is distance delta in the horizontal direction. Ideally test with constant speed, something like 1000px/s. 

During testing, please don't enable logging on right side - the oled log rendering may throw off timing...

My testing says:

[...]

PR #1559, opened 2026-05-21
Use exponential backoff for USB report retry throttle

kareltucek:
The previous fixed 10 ms minimum retry delay caused mouse-key stutter when USB reported "busy" rather than a hard error: every busy reply pushed the next attempt out by a full 10 ms.

Refactor ShouldResendReport into a simple retry counter plus a new GetResendThrottleDelay() that returns the per-attempt delay. The schedule is 1, 2, 4, 8, 16, 32 ms and then capped at 32 ms; MAX_RETRIES is sized so the cumulative wait stays at the original ~1024 ms budget.

Also switches the synchronous dongle resend loop in messenger.c (which already called ShouldResendReport) to the new counter semantics implicitly.

Changelog:
- make usb retries back off exponentially.

### DIFF
--- a/device/src/messenger.c
+++ b/device/src/messenger.c
@@ -242,6 +242,8 @@ static void processSyncablePropertyDongle(device_id_t src, const uint8_t* data, 
 #if DEVICE_IS_UHK_DONGLE
     uint8_t retryCounter = 0;
     while (ShouldResendReport(ret == 0, &retryCounter)) {
+        uint16_t delay = GetResendThrottleDelay(retryCounter);
+        k_sleep(K_MSEC(delay));
         k_sem_take(&dongleUsbSem, K_MSEC(128));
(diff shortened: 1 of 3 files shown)


pcooke9:
Thanks, I can confirm this fixes the mouse keys jitter via USB. I'm not sure if improved movement via dongle is also expected or not, but it's still pretty jittery for me.

kareltucek:
When did the dongle jitter start? Was it present in 17.0.0.?

pcooke9:
Well, a really long time ago 😅. It started way back in https://github.com/UltimateHackingKeyboard/firmware/issues/1114

Sorry, I don't think it's worse recently or anything though. I just don't understand the "dev speak" for whatever you said here 😅:

> Also switches the synchronous dongle resend loop in messenger.c (which already called ShouldResendReport) to the new counter semantics implicitly.

PR #1560, opened 2026-05-23
fix: enable I2C glitch filter (Width=15) to reject EMI above 2 MHz

semnil:
Set the K22 glitch filter to its maximum value on the main I2C bus only,
which runs to the left half over an external cable and is exposed to
external EMI sources. At the 60 MHz bus clock, FLT=15 absorbs glitches
up to 15 / 60 MHz = 250 ns, i.e. continuous EMI at frequencies of
1 / (2 * 250 ns) = 2 MHz or higher.

The EEPROM bus (I2C1) keeps the filter disabled. It runs at 1 MHz
(Fast-mode Plus) on short on-PCB traces only, where there is no EMI
exposure and 250 ns of timing margin cannot be spared from t_HIGH.

### DIFF
--- a/right/src/init_peripherals.c
+++ b/right/src/init_peripherals.c
@@ -123,6 +123,11 @@ static void initI2cBus(i2c_bus_t *i2cBus)
     i2c_master_config_t masterConfig;
     I2C_MasterGetDefaultConfig(&masterConfig);
     masterConfig.baudRate_Bps = i2cBus == &i2cMainBus ? I2cMainBusRequestedBaudRateBps : I2C_EEPROM_BUS_BAUD_RATE;
+    // Glitch filter is only enabled on the main bus, where the inter-half cable
+    // is exposed to external EMI. The EEPROM bus runs at 1 MHz on short on-PCB
+    // traces, so the filter is left disabled there to keep the SCL high time
+    // within Fast-mode Plus margins.
+    masterConfig.glitchFilterWidth = i2cBus == &i2cMainBus ? 15U : 0U;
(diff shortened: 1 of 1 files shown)


mondalaci:
I'm very intrigued about your PR, and it's the first time I've heard about the glitch filter. Are you a Kinetis expert or embedded developer?

Does your code manipulate I2Cx_FLT[^1] on the low level? How does the math work out to achieve above-4MHz rejection?

[^1]: https://github.com/UltimateHackingKeyboard/datasheets/blob/master/mcu/K22P121M120SF7RM.pdf page 1174

semnil:
@mondalaci

Thank you for pointing that out.
I have revised the code after clarifying the calculation formula and realizing that it also affects the EEPROM bus (I2C1_FLT).

I am a former embedded software developer.
This glitch filter setting proved effective when noise from the power supply was riding on the SCL/SDA lines of the I2C main bus.
Without a grunge filter, frequent interrupts were consuming most of the processing time, resulting in the situation described in [#1090](https://github.com/UltimateHackingKeyboard/firmware/issues/1090).

mondalaci:
Thank you for clarifying!

I've been testing this for several hours, and it seems as stable as the master to me. I can't reproduce the I2C issue, but I trust you, and the fix makes sense.

@kareltucek Any objections to merging this PR?

kareltucek:
No objections. Let's merge it.

mondalaci:
@semnil Thank you for your contribution! I'm very excited about this one. Our customers can generally work around I2C-related UHK 60 issues with our help if they occur, but sometimes they cannot, which can be painful. I hope that this fix solves the issue at its core, and I'm looking forward to the long-term results.

PR #1564, opened 2026-05-29
Jitter

### LINKED ISSUE #1114: Jittery mouse keys movement when using dongle

kareltucek:
Todo: 
- cleanup
- changelog

Closes #1114 .

Changelog:

- fix: decrease the dongle interval to actual 7.5ms, instead of scheduling-forced 11ms
- fix: increase advertising interval when not actively connecting, to reduce dongle mouse jitter

### DIFF
--- a/device/prj.conf.overlays/nrf_shared.conf
+++ b/device/prj.conf.overlays/nrf_shared.conf
@@ -29,6 +29,13 @@ CONFIG_BT_SMP=y
 
 CONFIG_BT_FILTER_ACCEPT_LIST=y
 
+# Experiment: shrink the SoftDevice controller's per-event reservation and
+# central ACL event spacing so a 7.5ms connection interval has scheduling
+# headroom (default 7500us leaves none). Trades NUS bulk throughput for lower
+# mouse latency/jitter.
+CONFIG_BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT=2500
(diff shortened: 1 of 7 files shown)


kareltucek:
@mondalaci I am increasing the advertising interval to 1 second when uhk is not trying to actively switch the host. Are we ok with this?

mondalaci:
What's the rationale for this change, and what was the previous behavior?

kareltucek:
Advertisement occupies outgoing ble transports, making the right<->dongle connection miss every ~5th transport window (with fast advertising). This makes cursor movement over dongle jittery - as Phil reports in #1114 .

mondalaci:
Sounds reasonable.

Just for the record, what was the previous advertising interval?

kareltucek:
```
/** Recommended minimum advertising interval for fast advertising (mode 1). 30 ms. */
#define BT_GAP_ADV_FAST_INT_MIN_1               0x0030
/** Recommended maximum advertising interval for fast advertising (mode 1). 60 ms. */
#define BT_GAP_ADV_FAST_INT_MAX_1               0x0060
```

Effectively 35ms according to logs.

PR #1565, opened 2026-06-01
uhk60: fix LED report filtering when report IDs aren't used

### LINKED ISSUE #1562: c2usb: led states not reported on uhk60
@benedekkupper on uhk60, it seems that `void keyboard_app::set_report(hid::report::type type, const std::span<const uint8_t> &data)` isn't getting called at all. 

Would you take a look please? 

(As a result, caps lock led isn't getting updated. It seems we have accidentally removed the led update code - fixed in 839586f .)

benedekkupper:
Closes #1562

Changelog:
- Fix: caps lock led on uhk60


kareltucek:
Perfect, thanks!

PR #1578, opened 2026-06-18
Show message when giving up connection attempt.

kareltucek:
Changelog:

- Show '<device> unavailable' when giving up connection attempt to a host.


kareltucek:
Closes #1578

PR #1590, opened 2026-07-02
UHK threat analysis

benedekkupper:
This is a PR to discuss about the device security through a threat analysis. This document shall be thorough, all conceivable threats should be listed. A lot of threats require physical access to the device, which aren't that relevant, once an attacker has physical access, they could as well replace the keyboard with a replica, or add a badUSB cable to it, etc. Let's perform the discussion as inline review of the document.

In my opinion, the change that would deliver the most added security is a password requirement (entered and processed on the keyboard, not sent to host) for FW update and for configuration R/W.

### DIFF
--- a/doc-dev/other/threat_analysis.md
+++ b/doc-dev/other/threat_analysis.md
@@ -1,0 +1,321 @@
+# Cybersecurity Threat Analysis: Ultimate Hacking Keyboard Firmware
+
+**Scope:** UHK60 (v1/v2) and UHK80 (left/right/dongle). Attack surfaces analyzed: USB HID protocol, macro engine, BLE stack, I2C module bus, firmware update pathway, and physical access.
+
+**Threat model assumption:** The keyboard is treated as a high-value target because it sits in the HID chain of every keystroke — including passwords, MFA codes, and shell commands — and because it has the ability to synthesize keystrokes autonomously.
+
+---
+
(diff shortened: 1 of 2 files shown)


mondalaci:
From a usability standpoint, I'd keep firmware updates and configuration reads/writes as they are by default.

However, we could add toggleable secure modes for firmware upgrades and dangerous USB commands, including configuration writes. I think separate security modes for various categories would make sense, but they feel overkill for configuration reads. A devMode-like mechanism could be used in $onInit to toggle these secure modes, and they should temporarily reset to their defaults when the UHK receives a hardware reset for recovery purposes, if a custom password is used instead of a firmware-hardcoded input.

The behavior of the secure modes is non-trivial, as not a single but usually multiple commands need to be executed, and subsequent commands should not be permitted. We'll need a mechanism to allow for this scenario.

The UHK is developer-friendly, so non-signed firmwares must be allowed.

kareltucek:
Security wise, it seems reasonable to me to require a secure default setting. A simple action that however requires user's deliberate acknowledgement. I think along the lines of the 4 corner keys as was discussed in some of the tickets.

Then there can be a macro command to disable security.

mondalaci:
That may work. Please elaborate on the specifics when you have time. This is not a priority issue. The end of your last comment seems cut off.

PR #1592, opened 2026-07-03
Usb diagnose

### LINKED ISSUE #1548: Benedek wish: on-demand c2usb state dump.
Story: Trying to get full streaming c2usb logs of issues is complicated. Partly because we pull those logs over usb, which means that with too low logging level, any usb log causes more usb logs, resulting in a flood which may or may not interfere with UHK operation. However, we often get into situation where we have a partly-operational that is sufficiently operational to run a macro or a zephyr command on-demand

Wish: I would like to have a function that I can call, which prints c2usb state, including any relevant debug information. 

@benedekkupper what do you think? Would it be useful to hunt those recent usb stack issues?

kareltucek:
Closes #1548


kareltucek:
[uhk-firmware-17.2.0-c1873a8.tar.gz](https://github.com/user-attachments/files/29747761/uhk-firmware-17.2.0-c1873a8.tar.gz)

PR #1606, opened 2026-07-26
Key Chords

firngrod:
Added key chords as a macro accessible feature.

I have left a self-review with some additional context which may not be readily available from the code, along with some questions about implementation strategy preferences.

There are also a lot of openers for discussions in the code comments, so don't skip those, by the way.


right/src/macros/commands.c
```diff
@@ -1302,8 +1305,12 @@ uint8_t Macros_TryConsumeKeyId(parser_context_t* ctx)
     uint8_t keyId = MacroKeyIdParser_TryConsumeKeyId(ctx);
 
     if (keyId == 255 && isNUM(ctx)) {
+        bool ErrorBefore = Macros_ParserError;
+        Macros_ParserError = false;
         uint8_t num = Macros_ConsumeInt(ctx);
-        if (Macros_ParserError) {
+        bool ErrorAfter = Macros_ParserError;
+        Macros_ParserError |= ErrorBefore;
+        if (ErrorAfter) {
```
firngrod, line 1278: Pretty sure I did this because I was getting misleading error messages.

right/src/macros/core.c
```diff
@@ -420,10 +420,22 @@ macro_result_t Macros_ExecMacro(uint8_t macroIndex)
...
+uint8_t startMacro(
+    uint8_t index,
+    key_state_t *keyState,
+    uint16_t argumentOffset,
+    uint8_t keyActivationId,
+    uint8_t parentMacroSlot,
+    bool runFirstAction,
+    const char *inlineText,
+    const macro_state_t *parentMacroState
+);
```
firngrod, line 430: The same refactoring from https://github.com/UltimateHackingKeyboard/firmware/pull/1507 because I need more initial state carry-over.

right/src/macros/core.c
```diff
@@ -461,11 +474,22 @@ uint8_t initMacro(
...
+        S->ms.currentMacroStartTime = S->ms.currentMacroStartTime;
+        S->ms.secondaryRoleState = S->ms.secondaryRoleState;
+        S->ms.isDoubletap = S->ms.isDoubletap;
+        S->ms.chordActivationId = S->ms.chordActivationId;
+    }
+    else {
+        S->ms.currentMacroStartTime = CurrentPostponedTime;
+        if (keyState != NULL) {
+            S->ms.isDoubletap = KeyHistory_WasLastDoubletap();
+            S->ms.chordActivationId = KeyHistory_GetChordActivationIdOfLastAction(); // returns invalid id if it was not a chord.
```
firngrod, line 487: I'm not a big fan of pulling the fact of chord activation out of the ether like this, transferring it indirectly through history.  The same thing is done for doubletap though.
Avoiding it would probably require the previously discussed key event type in `usb_report_updater.c`.

right/src/macros/set_command.c
```diff
@@ -844,6 +851,62 @@ static macro_variable_t keymapAction(parser_context_t* ctx, set_command_action_t
...
+    }
+
+    key_action_t chordAction = parseKeyAction(ctx);
+
+    if (Macros_ParserError || Macros_DryRun) {
+        return noneVar();
+    }
+
+    if (!Chords_TryAddChord(layerId, keys, keyCount, &chordAction)) {
+        Macros_ReportErrorPos(ctx, "Could not add chords, no more free chord slots.");
```
firngrod, line 905: Presumptions of error cause from a `bool`.  So far it's the only way it can fail, but maybe I should just do enum error codes to be proper?

right/src/config_manager.c
```diff
@@ -271,6 +271,9 @@ const config_t DefaultCfg = (config_t){
         .Macros_OneShotTimeout = 500,
         .AutoShiftDelay = 0,
         .ChordingDelay = 0,
+        .Chords_Timeout = 75,
```
firngrod, line 274: I use 30ms, @mhantsch mentioned using 100ms.  Put it somewhere in the middle.

right/src/key_history.c
```diff
@@ -1,55 +1,134 @@
...
+    return lastPress.multiTapCount % 2 == 0;
 }
 
 bool KeyHistory_WasLastMultitap()
 {
-    return lastPress.doubletapState >= DoubletapState_Multitap;
+    return lastPress.multiTapCount > 1;
+}
+
+uint8_t KeyHistory_GetChordActivationIdOfLastAction()
```
firngrod, line 124: Kind of presumes that the activation Id hasn't changed.  Should be changed if we ever support actual history.

right/src/postponer.c
```diff
@@ -483,10 +494,10 @@ uint32_t PostponerExtended_LastPressTime()
     return lastPressTime;
 }
 
-void PostponerExtended_ConsumePendingKeypresses(int count, bool suppress)
+void PostponerExtended_ConsumePendingKeypresses(int count)
 {
     for (int i = 0; i < count; i++) {
-        consumeOneKeypress(suppress);
+        consumeOneKeypress();
```
firngrod, line 500: It was not even in the function definition anymore.

right/src/usb_report_sender.c
```diff
@@ -136,12 +136,7 @@ static void clearMouseMovement(void) {
     }
 }
 
-static bool mouseButtonsChanged(void) {
```
firngrod, line 139: Warnings getting treated as errors.  As far as I can tell, these value actually were unused.

right/src/usb_report_updater.c
```diff
@@ -365,13 +366,13 @@ static void applyKeystrokePrimary(key_state_t *keyState, key_action_cached_t *ca
...
 
-static void applyKeystrokeSecondary(key_state_t *keyState, key_action_t *action, key_action_t *actionBase, usb_keyboard_reports_t* reports)
+static void applyKeystrokeSecondary(key_state_t *keyState, key_action_cached_t *cachedAction, usb_keyboard_reports_t* reports)
 {
-    secondary_role_t secondaryRole = action->keystroke.secondaryRole;
+    secondary_role_t secondaryRole = cachedAction->action.keystroke.secondaryRole;
     if ( IS_SECONDARY_ROLE_LAYER_SWITCHER(secondaryRole) ) {
         // If the cached action is the current base role, then hold, otherwise keymap was changed. In that case do nothing just
         // as a well behaved hold action should.
-        if(KeyState_Active(keyState) && action->type == actionBase->type && action->keystroke.secondaryRole == actionBase->keystroke.secondaryRole) {
```
firngrod, line 376: The uncached action was only used for this purpose:  To disable layer hold for secondary keys on keymap switch if the key was not mapped to the same layer on the new map.  I cannot see any other reasonable case for this code.  There is the unreasonable case of remapping the key from a macro while it's held.
I decided to remove that to get rid of passing the base action here because the base action has very little relation to the cached action when a chord is pressed.  Also, pure layer holds, (not secondary keys) are not subject to the same disabling on map change.

right/src/usb_report_updater.c
```diff
@@ -641,6 +642,7 @@ static void commitKeyState(key_state_t *keyState, bool active, uint8_t pressTime
     } else {
         KEY_TIMING(KeyTiming_RecordKeystroke(keyState, active, Timer_GetCurrentTime(), Timer_GetCurrentTime()));
         keyState->current = active;
+        EventVector_Set(EventVector_NativeActionsPostponing);
```
firngrod, line 648: As discussed, keys can now only get activated one per cycle.  Any more and the rest are postponed to future cycles.

right/src/chords.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "key_action.h"
+
+#define MAX_CHORD_KEYS 5
+#define CHORDS_ACTIVATIONID_SIZE 6
+#define CHORDS_INVALID_ACTIVATION_ID 0
+
+typedef enum {
+    ChordApplicationType_LeadingKey,
+    ChordApplicationType_AllKeys,
```
firngrod, line 12: Technically, the leading keys version has no benefit that I can see over the other except for being cheaper to perform.  Is that worth the complexity of the choice?  Maybe there should only be the all keys version?

firngrod (commented):
A couple of comments to start off the discussion.

right/src/macros/set_command.c
mhantsch: Either return `bool` and an additional `&failReason` parameter; or an `enum` as you suggest (0=OK, others are error codes).

Couldn't it also fail if the same chord (same keys) has already been registered?

right/src/config_manager.c
mhantsch: Sounds reasonable as a default. Is it configurable somewhere?

right/src/chords.h
firngrod: On the other hand, the macro support for release detection for all-keys lifetime was pretty complex, basically the entirety of [this commit](https://github.com/UltimateHackingKeyboard/firmware/pull/1606/changes/7c3c6c0ca0eb2ec8d81ed3690ab128b16141feea), probably a fifth of the entire PR, so maybe that should be the one to go.  Trouble is that I can actually conceive of real-world usecases for it, especially on the 60 where thumb keys are less abundant than on the 80.  With chords, the mod button can be overloaded to be various layers depending on which key(s) it was chorded with on press, and the other key can be released while modkey is held to maintain the layer.  Not really useful for me with my "No mapped combination keys" philosophy, but probably for someone else.

right/src/macros/set_command.c
firngrod: Same chord will overwrite previous chord with new action, or delete it if the new action is `none`

right/src/config_manager.c
firngrod: Yeah, it's in the reference manual.

right/src/macros/set_command.c
mhantsch: So it's `TryAddOrReplaceChord()`? How about `TrySetChord()` or just `SetChord()` (anything is a try if it can fail).

firngrod:
@kareltucek I know you are probably just busy with more important stuff right now, but pinging you in case it just slipped under your radar.  If you are busy, no pressure.

firngrod:
Just updated the branch with latest master.  I have not yet tested these changes.  The last time I tried to test with latest master, my keyboard was unusable until I rolled it back again, which wasn't easy.  I just could not get 18.0.0 to work on my UHK60v1.

I will try again sometime soon with this latest master merge.

firngrod:
That being said, I have run the chords changes off of 17.2.0 for months, and I'm happy with how it works.  I cannot go back to the chordless world.

mhantsch:
I merged 18.0.1 into my own fork (with `macroArg`) and it works both on UHK80 and UHK60v2. I haven't tried it on UHK60v1.

mondalaci:
Great, thanks!

firngrod:
I think there may have been a mistake here?  I don't believe this PR has had any real code review, unless you did that last night, @mondalaci ?  I just merged the latest master and specifically made a note that I have not tested the merge yet.

mondalaci:
@firngrod I'm sorry, and thanks for the heads-up! I meant to merge a different PR. Please revert this one.

kareltucek:
@firngrod sorry for complications, and sorry for still not getting to review this for so long :-(. 

I have force pushed the old master. I am not sure what to do about the PR though, as it can't be flipped into an open state again... 

>  I meant to merge a different PR. Please revert this one.

@mondalaci  Which one?

mondalaci:
@kareltucek An Agent PR. Not relevant.

firngrod:
@kareltucek No worries.  I just assumed you were on vacation or something.
Regarding the PR, I have two monitors and a fancy keyboard, I'll make a new one and copy the comments over.

kareltucek:
We can also continue the discussion here, and transition to the new one when convenient 🤷.

firngrod:
I have not looked at this code for over a month.  Making a new PR will also give me a chance to look over things again, so it's all good.

pcooke9:
> I just could not get 18.0.0 to work on my UHK60v1

I just dug my UHK60v1 out of storage for the first time in almost a year. Recent master v18.0.2 is working fine for me.

firngrod:
I have also flashed my branch with 18.0.2 merged, and it works fine.  I suspect it has to do with the patch note "Fix: make uhk60 not freeze on reboot."  ;)

PR #1609, opened 2026-07-30
Switchover fixes

kareltucek:
Fixes to the switchover refactor.

### DIFF
--- a/device/src/bt_conn.c
+++ b/device/src/bt_conn.c
@@ -1155,7 +1155,6 @@ void BtConn_ReserveConnections() {
         BtConn_DisconnectAllUnidentified();
 
         uint8_t unusedConnectionCount = BtConn_UnusedPeripheralConnectionCount();
-        bool selectedConnectionIsBleHid = Connections_Type(CurrentHostConnectionId) == ConnectionType_BtHid;
 
         if (unusedConnectionCount == 0) {
             disconnectOldestHost();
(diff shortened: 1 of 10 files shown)


kareltucek:
[uhk-firmware-17.2.0-staging_260730.tar.gz](https://github.com/user-attachments/files/30587061/uhk-firmware-17.2.0-staging_260730.tar.gz)

pcooke9:
> [uhk-firmware-17.2.0-staging_260730.tar.gz](https://github.com/user-attachments/files/30587061/uhk-firmware-17.2.0-staging_260730.tar.gz)

This version doesn't wake PC over USB. If you have a dongle connected to the same PC, you can switch to it and then wake the PC with the dongle.

In fact, if the PC sleeps or even reboots etc., USB won't reconnect at all until using the `reboot` macro command. Physically reconnecting USB cable doesn't resolve it either. You can still `switchHost` to dongle or whatever, but USB doesn't reconnect.

kareltucek:
Oops! Fixed here: https://github.com/user-attachments/files/30597660/uhk-firmware-17.2.0-staging_260731_v2.tar.gz

pcooke9:
> Oops! Fixed here: https://github.com/user-attachments/files/30597660/uhk-firmware-17.2.0-staging_260731_v2.tar.gz

The USB reconnection issue is resolved now.

___

I'm not sure if the following issues were present in the previous build or not, but...

This build has trouble connecting to my phone for some reason. The phone shows it's connected, but UHK still says it's trying to connect. Every now and then it will appear connected on both UHK and phone, but phone doesn't receive UHK output. Very rarely, the phone will work properly, but it feels like pure luck when it does. If it does manage to work, it will almost definitely have trouble again upon restarting the phone BT radio, or rebooting the UHK, etc..

Connecting to my PC's BT6-USB adapter BLE works almost every time.🤷🏻‍♂️

Also some duplicate device issues. I had one of my dongles plugged into my TV. Then I reset my config to default, and that dongle was recognized as a BLE device. Then I moved it from the TV to the PC and Agent offered to re-pair it. It was then properly recognized as a dongle, but the other BLE entry remained in Agent's Connections list.

Sorry, I don't have logs ATM, but I can try to gather some later if you need em.

kareltucek:
> This build has trouble connecting to my phone for some reason. 

I have pushed another version with some bt adjustments.

> Sorry, I don't have logs ATM, but I can try to gather some later if you need em.

I don't think they would help ¯\\_(ツ)_/¯

PR #1620, opened 2026-09-04
c2usb update to fix USB disconnected after rollover reconfigure

benedekkupper:
This PR fixes the specific scenario of #1542 when it is triggered by `uhk rollover` shell command.


kareltucek:
Thank you!

PR #1624, opened 2026-09-13
Firngrod/key chords

firngrod:
Added key chords as a macro accessible feature.

I have left a self-review with some additional context which may not be readily available from the code, along with some questions about implementation strategy preferences.

There are also a lot of openers for discussions in the code comments, so don't skip those, by the way.


doc-dev/reference-manual.md
```diff
@@ -722,6 +726,14 @@ Key actions can be parametrized with macro arguments. These arguments can be exp
...
 
+- Chords
+  Chords is a concept where a number of keys can be linked together such that if they are pressed together all at once, like a chord on a piano, a specific action is taken, different from what each of the keys do on their own.  Anywhere between 2 and 5 keys can be chorded together and associated with an `ACTION`
+
+  - `set chord[.LAYERID_BASIC] ACTION` Configures a chord with an action.  If a layer is specified, the chord will be available when that layer is active, if not, it will be available on any layer.  The same chord can be defined for different actions for different layers, including the `any` layer.  If action is `none`, the chord is removed and the keys will work as individual keys if the chord is pressed, not as an empty action.
```
firngrod, line 732: Should the chord count cap be mentioned?

doc-dev/reference-manual.md
```diff
@@ -722,6 +726,14 @@ Key actions can be parametrized with macro arguments. These arguments can be exp
...
+  - `set chordTimeout <ms, 0-255 (INT)>` sets the interval within which all keys of a chord must be pressed in order for the chord to be activated.
+  - `set chordPriorIdleTime <ms, 0-65535 (INT)>` sets a limit to how soon a chord can be activated after a key has been pressed.  This helps prevent accidental activations while typing rapidly.
```
firngrod, line 734: Mention that chords do not block chord activations, only other key actions?

doc-dev/reference-manual.md
```diff
@@ -722,6 +726,14 @@ Key actions can be parametrized with macro arguments. These arguments can be exp
...
+  - `set chordLifetime { leadingKey | allKeys }` sets how long a chord activation lasts.  With `leadingKey`, the chord is considered released when earliest key pressed in the activation is released, with `allKeys`, the chord is considered released when all keys in the chord is released.
```
firngrod, line 735: The main differentiator is that with `allKeys`, the release is deterministically when all keys are released, not while with `leadingKey`, it can realistically be any of the keys.  This can allow for, for example, overloading of a mod key.

right/src/macros/commands.c
```diff
@@ -1267,8 +1270,12 @@ uint8_t Macros_TryConsumeKeyId(parser_context_t* ctx)
     uint8_t keyId = MacroKeyIdParser_TryConsumeKeyId(ctx);
 
     if (keyId == 255 && isNUM(ctx)) {
+        bool ErrorBefore = Macros_ParserError;
+        Macros_ParserError = false;
         uint8_t num = Macros_ConsumeInt(ctx);
-        if (Macros_ParserError) {
+        bool ErrorAfter = Macros_ParserError;
+        Macros_ParserError |= ErrorBefore;
+        if (ErrorAfter) {
```
firngrod, line 1278: Pretty sure I did this because I was getting misleading error messages.

right/src/macros/core.c
```diff
@@ -63,9 +63,6 @@ macro_scope_state_t MacroScopeState[MACRO_SCOPE_STATE_POOL_SIZE];
 macro_state_t MacroState[MACRO_STATE_POOL_SIZE];
 macro_state_t *S = NULL;
 
-macro_history_t MacroHistory[MACRO_HISTORY_POOL_SIZE];
-uint8_t MacroHistoryPosition = 0;
-
```
firngrod, line 68: Dead code since doubletap was moved

right/src/macros/core.c
```diff
@@ -423,10 +417,22 @@ macro_result_t Macros_ExecMacro(uint8_t macroIndex)
...
+    uint8_t index,
+    key_state_t *keyState,
+    uint16_t argumentOffset,
+    uint8_t keyActivationId,
+    uint8_t parentMacroSlot,
+    bool runFirstAction,
+    const char *inlineText,
+    const macro_state_t *parentMacroState
+);
+
```
firngrod, line 431: The same refactoring from #1507 because I need more initial state carry-over.

right/src/macros/core.c
```diff
@@ -464,11 +471,22 @@ uint8_t initMacro(
...
+        S->ms.currentMacroStartTime = S->ms.currentMacroStartTime;
+        S->ms.secondaryRoleState = S->ms.secondaryRoleState;
+        S->ms.isDoubletap = S->ms.isDoubletap;
+        S->ms.chordActivationId = S->ms.chordActivationId;
+    }
+    else {
+        S->ms.currentMacroStartTime = CurrentPostponedTime;
+        if (keyState != NULL) {
+            S->ms.isDoubletap = KeyHistory_WasLastDoubletap();
+            S->ms.chordActivationId = KeyHistory_GetChordActivationIdOfLastAction(); // returns invalid id if it was not a chord.
```
firngrod, line 487: I'm not a big fan of pulling the fact of chord activation out of the ether like this, transferring it indirectly through history. The same thing is done for doubletap.
Avoiding it would probably require the previously discussed key event type in usb_report_updater.c, which would in some ways be more proper, and yet in other ways not.

right/src/macros/set_command.c
```diff
@@ -800,6 +801,17 @@ static macro_variable_t navigationModeAction(parser_context_t* ctx, set_command_
     return noneVar();
 }
 
+static inline key_coordinates_t validateKeyId(uint8_t keyId, const char *at) {
```
firngrod, line 804: Naming is hard.  Could be named for the other main purpose of splitting into coordinates.

right/src/macros/set_command.c
```diff
@@ -845,6 +852,62 @@ static macro_variable_t keymapAction(parser_context_t* ctx, set_command_action_t
...
+        return noneVar();
+    }
+
+    key_action_t chordAction = parseKeyAction(ctx);
+
+    if (Macros_ParserError || Macros_DryRun) {
+        return noneVar();
+    }
+
+    if (!Chords_TryAddChord(layerId, keys, keyCount, &chordAction)) {
```
firngrod, line 904: Condensed discussion from previous PR:

The error message presumes failure (false) means out of chord slots.  This is currently true, existing chords are replaced, but that could change.  Generally, returning enummed error codes is more proper, but makes code much more verbose.  Should I do that instead?

Also, as @mhantsch mentions, probably `TryAddOrReplaceChord` is a better name.

right/src/chords.c
firngrod, line 1: A lot of the comments in this file could be PR comments.

right/src/chords.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "key_action.h"
+
+#define MAX_CHORD_KEYS 5
+#define CHORDS_ACTIVATIONID_SIZE 6
+#define CHORDS_INVALID_ACTIVATION_ID 0
+
+typedef enum {
+    ChordApplicationType_LeadingKey,
+    ChordApplicationType_AllKeys,
```
firngrod, line 12: A discussion about the value of choice between these two lifetime strategies would not be unwarranted.

From a user perspective, `LeadingKey` (might as well be `TrailingKey`, by the way) has no benefit, but codewise, it is much simpler to do.

However, when the code for `AllKeys` exists, using `LeadingKey` is not going to be noticably faster than `AllKeys`, the cost of `AllKeys` is the code complexito to have it at all.  It adds so much complexity to have to remember key states and cross-reference multiple keys when releasing.

So there is a real question of whether there should be a choice at all, and if not, whether to use one or the other.

right/src/config_manager.c
```diff
@@ -271,6 +271,9 @@ const config_t DefaultCfg = (config_t){
         .Macros_OneShotTimeout = 500,
         .AutoShiftDelay = 0,
         .ChordingDelay = 0,
+        .Chords_Timeout = 75,
```
firngrod, line 274: I use 40ms now, @mhantsch mentioned using 100ms.  I put the default arbitrarily somewhere in the middle.

right/src/key_history.c
```diff
@@ -1,55 +1,134 @@
...
+    return lastPress.multiTapCount % 2 == 0;
 }
 
 bool KeyHistory_WasLastMultitap()
 {
-    return lastPress.doubletapState >= DoubletapState_Multitap;
+    return lastPress.multiTapCount > 1;
+}
+
+uint8_t KeyHistory_GetChordActivationIdOfLastAction()
```
firngrod, line 124: The implementation here fetches activationId from the chord rather than having a cached value.  This would of course have to change if we ever want to have an actual histrory.

right/src/postponer.c
```diff
@@ -483,10 +494,10 @@ uint32_t PostponerExtended_LastPressTime()
     return lastPressTime;
 }
 
-void PostponerExtended_ConsumePendingKeypresses(int count, bool suppress)
+void PostponerExtended_ConsumePendingKeypresses(int count)
 {
     for (int i = 0; i < count; i++) {
-        consumeOneKeypress(suppress);
+        consumeOneKeypress();
```
firngrod, line 500: It was not even in the function definition any more.

right/src/usb_report_updater.c
```diff
@@ -367,13 +368,13 @@ static void applyKeystrokePrimary(key_state_t *keyState, key_action_cached_t *ca
...
-static void applyKeystrokeSecondary(key_state_t *keyState, key_action_t *action, key_action_t *actionBase, usb_keyboard_reports_t* reports)
+static void applyKeystrokeSecondary(key_state_t *keyState, key_action_cached_t *cachedAction, usb_keyboard_reports_t* reports)
 {
-    secondary_role_t secondaryRole = action->keystroke.secondaryRole;
+    secondary_role_t secondaryRole = cachedAction->action.keystroke.secondaryRole;
     if ( IS_SECONDARY_ROLE_LAYER_SWITCHER(secondaryRole) ) {
         // If the cached action is the current base role, then hold, otherwise keymap was changed. In that case do nothing just
         // as a well behaved hold action should.
-        if(KeyState_Active(keyState) && action->type == actionBase->type && action->keystroke.secondaryRole == actionBase->keystroke.secondaryRole) {
+        if(KeyState_Active(keyState)) {
```
firngrod, line 377: The uncached action was only used for this purpose: To disable layer hold for secondary keys on keymap switch if the key was not mapped to the same layer on the new map. I cannot see any other reasonable case for this code. There is the unreasonable case of remapping the key from a macro while it's held.
I decided to remove that to get rid of passing the base action here. Deciding on a base layer action was tricky in the cases where the held key was a chord. Also, pure layer holds, (not secondary keys) are not subject to the same disabling on map change.

right/src/usb_report_updater.c
```diff
@@ -644,6 +645,7 @@ static void commitKeyState(key_state_t *keyState, bool active, uint8_t pressTime
         DEBUG_KEY_LIFE(applied);
         KEY_TIMING(KeyTiming_RecordKeystroke(keyState, active, Timer_GetCurrentTime(), Timer_GetCurrentTime()));
         keyState->current = active;
+        EventVector_Set(EventVector_NativeActionsPostponing);
```
firngrod, line 648: As we discussed in the issue thread, only one key can now get activated per cycle, with the rest being pushed on the queue before actual key evaluation, in order to eliminate cases where one key evaluating for chords or secondary roles would not see when another was pressed within the same cycle.

right/src/usb_report_updater.c
```diff
@@ -790,17 +815,24 @@ static void updateActionStates() {
...
 
                 //apply active-layer action
-                ApplyKeyAction(keyState, cachedAction, actionBase, &NativeKeyboardReports);
+                ApplyKeyAction(keyState, cachedAction, &NativeKeyboardReports);
 
                 if (KeyState_DeactivatedNow(keyState)) {
+                    // I kind of want to only trigger chord releases for chord-holding keys.
+                    // It's safe as it is, but it is not exactly free if there are chords.
+                    // Maybe grab a bit in the keyState for isHoldingChord?
+                    // I could also get rid of the need to do this by storing activationIds for the holding keys, costing memory per chord.
```
firngrod, line 834: Consider this a PR comment.  It should be removed before merge.

firngrod (commented):
Conversation starters

doc-dev/reference-manual.md
```diff
@@ -183,6 +183,10 @@ COMMAND = set oneShotTimeout <time in ms (INT)>
 COMMAND = set macroEngine.batchSize <number of commands to execute per one update cycle INT>
 COMMAND = set navigationModeAction.NAVIGATION_MODE_CUSTOM.DIRECTION ACTION
 COMMAND = set keymapAction.LAYERID.KEYID ACTION
+COMMAND = set chord[.LAYERID_BASIC] ACTION
```
kareltucek, line 186: I suspect this should read `set chord[.LAYERID_BASIC] KEYID KEYID [KEYID]* ACTION`

In any case please add some example somewhere. Feel free to add an example-oriented section into the user-guide.md

doc-dev/reference-manual.md
kareltucek: Weak preference for yes.

doc-dev/reference-manual.md
kareltucek, line 172: (Someone should probably mention somewhere that this is an independent concept.)

right/src/chords.c
```diff
@@ -0,0 +1,412 @@
...
+    ChordSearch_Partial = 0b01, // There was no exact match, but there were chords eligible with more keys
+    ChordSearch_Exact = 0b10, // There was an exact match and no other potential matches
+} chord_search_result_t;
+
+chord_def_t Chords[MAX_CHORDS_COUNT];
+uint8_t ChordCount = 0;
+uint8_t NextActivationId = CHORDS_INVALID_ACTIVATION_ID + 1;
+
+// Not sure how nice or not this is.  I like named parameters rather than just true/false in calls.
+// I also don't like the all-caps of macros, this C doesn't have consexpr, and I don't know if global consts will get optimized.
```
kareltucek, line 24: Interesting. I bit cryptic. I kind of like it as long as it doesn't pollute global namespace.

right/src/chords.c
```diff
@@ -0,0 +1,412 @@
...
+#include "key_history.h"
+#include "keymap.h"
+#include "layer.h"
+#include "macros/status_buffer.h"
+#include "postponer.h"
+#include "utils.h"
+
+#define MAX_CHORDS_COUNT 64
+
+// Not sure if this should be an enum or just macros.
```
kareltucek, line 13: Enum is fine.

right/src/usb_report_updater.c
```diff
@@ -790,17 +815,24 @@ static void updateActionStates() {
...
                 Trace_Printc("w4");
 
                 //apply active-layer action
-                ApplyKeyAction(keyState, cachedAction, actionBase, &NativeKeyboardReports);
+                ApplyKeyAction(keyState, cachedAction, &NativeKeyboardReports);
 
                 if (KeyState_DeactivatedNow(keyState)) {
+                    // I kind of want to only trigger chord releases for chord-holding keys.
+                    // It's safe as it is, but it is not exactly free if there are chords.
+                    // Maybe grab a bit in the keyState for isHoldingChord?
```
kareltucek, line 833: I am in favor for grabbing a bit in the keyState for isHoldingChord.

right/src/macros/set_command.c
kareltucek: This is fine.

I would probably formulate the error so that it is not a strict statement - like ending with a question mark instead of a period, to at least give myself a hint to check the path in case it behaves weird.

right/src/usb_report_updater.c
kareltucek, line 834: I don't understand the `storing activationIds` idea, but I suspect it would take time for lookups at other places.

right/src/usb_report_updater.c
```diff
@@ -742,14 +744,24 @@ static void updateActionStates() {
     for (uint8_t slotId=0; slotId<SLOT_COUNT; slotId++) {
```
kareltucek, line 744: From what I remember, this loop is quite performance critical.

For that reason, I had been keeping it one loop without separating the preprocessing into a separate one.

right/src/usb_report_updater.c
```diff
@@ -790,17 +815,24 @@ static void updateActionStates() {
...
 
                 cachedAction = &actionCache[slotId][keyId];
-                actionBase = &CurrentKeymap[LayerId_Base][slotId][keyId].action;
 
                 Trace_Printc("w3");
-                //apply base-layer holds
-                applyLayerHolds(keyState, actionBase);
+                // apply base-layer holds, but only if the key is not currenty applying another action
+                // Also, don't apply layers from keys currently undergoing chord evaluation
+                if (chordRes != ChordResolution_Wait && cachedAction->action.type == KeyActionType_None) {
```
kareltucek, line 822: This changes layer hold semantics. See `right/src/layer_switcher.c` scenario comments.

This is getting a bit convoluted, as I had probably introduced some artifacts too over time.

I should probably back off, write some tests and see where all the discrepancies that I am seeing come from.

kareltucek (commented):
Just a couple comments from a first, very shallow pass.

It will take a few more days till I get back for a deeper pass.

doc-dev/reference-manual.md
firngrod: Yeah, woopsie

right/src/chords.c
firngrod: I just like that it's extremely obvious what means pressed and what means released.

right/src/usb_report_updater.c
firngrod: Well, we need a solution for when 2 keys are pressed within the same cycle:
`a` and `b` are a chord and pressed in one cycle.  `a` has lowest id and runs first, starts chord resolution but does not see `b` as it hasn't been dealt with yet, starts postponing, waiting for more keys.  `b` is then postponed, but nothing causes `a` to be handled again until timeout.
Another option could be to always schedule a run the next cycle if postponing has been enabled and a key postponed because of it within the same cycle, though that could get tricky with where secondary keys set postponing.  Alternatively, always trigger a cycle if a key is put in postponer?

right/src/usb_report_updater.c
firngrod: Only for keys which have an action associated to them on layers in which they are pressed.
This code change targets the case where key `a` was pressed on layer `mod` with action `keypress leftArrow`, but `a` has a layer switch to `fn` on base layer, so when mod is released, `a` now holds `leftArrow` *and* toggles `fn`.  It' an edge case.
I get the need for empty actions to work like that, so key release/press order is not strictly enforced on layer dancing, but I feel that if the key is bound on the first layer, the user has already forefeited clean layer dancing.
I will change it to only skip the layer hold if chord is awaiting resolution if you want a more conservative approach.

right/src/usb_report_updater.c
firngrod: Will do that then.  That will make release key handling a bit cheaper if chords are enabled with maintain until all keys released.

right/src/usb_report_updater.c
firngrod: If active chords record the key activation id for each key which "holds it", release checks for the chord can just compare key activation ids, as per normal key release procedure.  That would remove the need for tracking releases like this, but that would cost 5 bytes per chord slot.  I like isHoldingChord, unless I find out that it's bad somehow.

right/src/usb_report_updater.c
firngrod: Just always scheduling another cycle whenever a key is put on postponer might be the best/cleanest idea?  It may cost a bit of processing, but only while there's downtime anyway.

doc-dev/reference-manual.md
firngrod: Chords have not been released.  It's not too late to name them something else, though chord is the most natural name.  Unfortunate that key cluster kind of rules out that name for chords.

right/src/usb_report_updater.c
firngrod: What I'm not too fond of is the way layer switches are handled so many places.  Would it be possible to recheck pressed keys when switching layer to base and fake an activation for the first pressed with a cached none-action and base layer layer switch action, if any?  Then this case of layer handling could be removed, and the one in `ApplyKeyAction` could be the main one?  It's also complicated, in a different way.  Also, secondary role layer switches are still different.

right/src/chords.h
firngrod: I could also implement `userDefined` where the first key in the macro definition becomes the "carrying" key regardless of how the keys were pressed.  That allows for determinism in what key to hold, while being simpler than `AllKeys` to implement.  It's slightly less versatile than `AllKeys`, but I think it may be worth it for code simplicity.  Then Macros can forget about chords again and release check becomes a simple key release check again.
EDIT:  I have personal reservations against doing the `userDefined` option, which I may write up a small essay on. Basically, it makes an idea I have for a very specific mix of chords and secondary roles impossible.

doc-dev/reference-manual.md
firngrod: Maybe a combo?

right/src/chords.c
```diff
@@ -0,0 +1,412 @@
...
+            PostponerExtended_ConsumePendingKeypresses(keyCount - 1);
+        }
+        else if (Cfg.Chords_ApplicationType == ChordApplicationType_AllKeys) {
+            if (matchedChord->action.type == KeyActionType_PlayMacro) {
+                // Do not activate on the first key, but rather on the last.
+                // This is to prevent the following issues:
+                //  - activateKeyPostponed with prepend, or consumePending modifying the roll-out
+                //  - the rest of the chord waiting on the queue causing problems with ifSecondary in the macro we run
+                // We are still hypothetically vulnerable to other macros using those commands, but that's highly hypothetical as they would
+                // likely be used while postponeKeys is active, meaning we would not be resolving anyway.
```
firngrod, line 366: I could probably treat the keystrokes the same as the macros and activate them on the last key, but I wasn't entirely sure if it would cause issues to have the first keys hold the action without an `ActivatedNow` round and then having the last key do an `AcitvatedNow`.  Probably it's fine, didn't take the chance.  If it is fine, this logic can be simplified.

firngrod:
I'm still fretting over the complexity of `allKeys`.  I don't like the price for what it provides.
I just thought of another possibility:  Use the current `allKeys` functionality for simple keystrokes, but for macros, make the first key provided in the definition the holding key, always.  Then the macro code does not have to know anything about chords, release checks in macros become a single key again, all of the release tracking for chords goes away.
The downside:  Behavior is different for keystrokes and macros.  A chord mapped to `keystroke mouseBtnLeft` does not behave exactly like a chord mapped to `macro leftClick` where `leftClick` is just `holdKey mouseBtnLeft`.  This could be mitigated through documentation, and macros can stil do `ifNotActive a ifNotActive b exit` if they really want to maintain for all keys, though that brings back the issues solved through `key_state_t.activationId`

PR #1625, opened 2026-09-15
UHK60 LED control through Windows

benedekkupper:
Plug in the UHK60 to a Windows PC, open Settings / Personalization / Dynamic Lighting, and play around with the settings. My favorite is the Wave effect.

At the moment this is only made to work correctly on UHK60.

For accurate operation, the UHK must tell the PC how long it takes to update all LEDs before a new set of values can be received, this should be set based on real measurements (I just eyeballed it to 35ms).

Furthermore, each LEDs position has to be reported to the PC with um accuracy in 3D space - again, I only eyeballed the currently used values, but they look not half bad IMHO.


kareltucek:
Interesting!

> Furthermore, after a few minutes of animation the keyboard usually gets into an all LEDs off state, which it's not able to recover from, until a button is pressed.

Uhk has independent handling of sleep modes and brightness, which converges here:

https://github.com/UltimateHackingKeyboard/firmware/blob/8c8c1d7823876257ce88069d2eb7140c7b0d04a5/right/src/led_manager.c#L36-L39

In practice, just add your mode here:

https://github.com/UltimateHackingKeyboard/firmware/blob/8c8c1d7823876257ce88069d2eb7140c7b0d04a5/right/src/ledmap.c#L759

benedekkupper:
A video on how it looks: https://drive.google.com/file/d/11gFxeSFJDL8kNkSkNB8e0HFM87VdiLmC/view?usp=sharing

I have updated the code to explicitly disable it on UHK80. That system is more complicated, a lot of design decisions need to be made. On UHK60 it is rather straightforward, I just hijacked the temporary backlight mode, which is only used on UHK80 for BT pairing.

Windows allows disabling this feature globally on all devices, or individually per device. I have split the UHK into two virtual devices, as they are more likely used in split mode, but that means that disabling one side and enabling the other doesn't work well.

Another Windows setting is whether applications are allowed to control the LEDs on each device individually. This is managed by the OS itself.

One thing to consider, if there should be a configuration value on the UHK side to enable this feature. At the moment the assumption is that the host side is sending messages that respect the timing requirements specified by the device. This might not always be the case, e.g. if on Linux a userspace app is misusing this HID interface, the LED buffer can get corrupted.

mondalaci:
Is this LED control feature Windows-specific, or is there an OS agnostic standard behind it?

benedekkupper:
> Is this LED control feature Windows-specific, or is there an OS agnostic standard behind it?

It is standard HID, any OS or userspace application can decide to implement support for it. Example applications:
https://openrgb.org/
https://github.com/xz-dev/hid-rgb-ctl
In particular OpenRGB seems to have a significant community around it.

